PR #6, opened 2024-03-04
feat(ForceFeedback): add force feedback support

ShadowApex:


src/input/target/gamepad.rs
```diff
@@ -19,7 +25,10 @@ use crate::input::{
 
 use super::TargetCommand;
 
+/// Size of the [TargetCommand] buffer for receiving input events
 const BUFFER_SIZE: usize = 2048;
+/// How long to sleep before polling for events.
+const POLL_RATE: Duration = Duration::from_micros(1666);
```
pastaq, line 31: I think this should be defined by the source device.

ShadowApex: We can look into this in a future PR.

PR #19, opened 2024-03-14
fix(Intercept): always send guide events over dbus when intercept mode > 0

ShadowApex:


Ericky14 (approved):

src/input/composite_device/mod.rs
```diff
@@ -522,17 +521,32 @@ impl CompositeDevice {
...
         if matches!(self.intercept_mode, InterceptMode::Always) {
+            let event = TargetCommand::WriteEvent(event);
+            #[allow(clippy::for_kv_map)]
             for (_, target) in &self.target_dbus_devices {
                 target.send(event.clone()).await?;
             }
             return Ok(());
         }
 
+        // When intercept mode is enabled, send ALL Guide button events over DBus
```
pastaq, line 535: Won't this interfere with Steam chords in overlay mode?

PR #36, opened 2024-04-03
feat(Gyro): add gyro support for BMI IMU

ShadowApex:


src/input/source/iio.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /// Run the source IIO device
+    pub async fn run(&self) -> Result<(), Box<dyn Error>> {
+        // Run the appropriate IIO driver
+        let Some(name) = self.info.name.clone() else {
+            return Err("Unable to determine IIO driver because no name was found".into());
+        };
+
+        // BMI Driver
```
pastaq, line 93: This is an assumption that isn't always true. The bmi323 has a different pattern and the gyro in the orange pi is "display_gyro" despite being a bmi260

ShadowApex: I can add these as patterns. I just need to know what the possible values are.

PR #50, opened 2024-04-14
feat(SteamDeck): add rumble support to Steam Deck controller source device

ShadowApex:


Makefile
```diff
@@ -174,6 +162,22 @@ dist/$(NAME).raw: dist/$(NAME).tar.gz
 	mv $(CACHE_DIR)/$(NAME).raw $@
 	cd dist && sha256sum $(NAME).raw > $(NAME).raw.sha256.txt
 
+$(CACHE_DIR)/libiio:
```
pastaq, line 165: More secret undocumented makefile changes I see...

ShadowApex: This just caches libiio for the systemd-exension so it is not downloaded every time you build the systemd-extension.

Issue #60, opened 2024-04-25
ROG Ally IMU causes InputPlumber to silently crash


pastaq:
When the BMI323 IMU it properly initialized (#59 ) , InputPlumber works for 3-6 minutes before suddenly hard locking. When this happens all input is lost and whatever the last key or keys were being pressed are stuck in a held state. Attempting to restart the service or pkil -9 inputplumber causes the client program to hard freeze. Rebooting the system fails due to a hung process and the system requires a hard reset.

This happens even in games that don't read the DualSense IMU device/support gyro.

All testing performed with the DualSense Edge target device.

PR #61, opened 2024-04-26
fix(CompositeDevice): Properly clear BMI source device command queue.

pastaq:
- Adds buffer reading to BMI IIO deivce to ensure the command queue doesn't overflow and hand the driver.
- Uses try_send instead of send().await to ensure an improperly implemented driver cannot block the main thread.

### DIFF
--- a/src/input/source/iio/bmi_imu.rs
+++ b/src/input/source/iio/bmi_imu.rs
@@ -1,7 +1,10 @@
 use core::time;
 use std::{error::Error, f64::consts::PI, thread};
 
-use tokio::sync::broadcast;
+use tokio::sync::{
+    broadcast,
+    mpsc::{self, error::TryRecvError},
+};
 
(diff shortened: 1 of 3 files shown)


pastaq:
Solves #60

src/input/source/iio/bmi_imu.rs
```diff
@@ -84,13 +94,14 @@ impl IMU {
                     }
 
                     // Sleep between each poll iteration
-                    let duration = time::Duration::from_micros(250);
+                    let duration = time::Duration::from_millis(40);
```
ShadowApex, line 97: Do we want to change this back to 250µs? In the future, we can get this from the driver interface based on polling frequency.

ShadowApex (changes requested):

ShadowApex (approved):

PR #63, opened 2024-04-28
fix(CompositeDevice): Ensure non-button events such as gyro do not break intercept mode checking.

pastaq:
- Only checks intercept mode events that are of the 'Button' sub-type for keyboard, gamepad, and mouse. Axis events are now ignored.
- Move logging for IIO finding into the proper location.

### DIFF
--- a/src/input/composite_device/mod.rs
+++ b/src/input/composite_device/mod.rs
@@ -1152,7 +1152,7 @@ impl CompositeDevice {
             }
         }
 
-        for mut event in events {
+        for event in events {
             let cap = event.as_capability();
 
             // Track what is currently active so we can ignore extra events.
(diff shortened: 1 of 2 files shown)


src/input/composite_device/mod.rs
```diff
@@ -1152,105 +1152,57 @@ impl CompositeDevice {
...
                     }
-                    log::debug!("We have enough events to evaluate intercept mode.");
-                    // Clean up the active inputs list.
-                    for c in self.intercept_activation_caps.clone() {
-                        if self.active_inputs.contains(&c) {
-                            log::trace!("Removed inactive capability: {c:?}");
-                            let index = self.active_inputs.iter().position(|r| r == &c).unwrap();
-                            self.active_inputs.remove(index);
+                }
+                Capability::Gamepad(t) => match t {
```
ShadowApex, line 1175: Use `Capability::Gamepad(ref t)` to avoid needing to do `cap.clone()`

src/input/composite_device/mod.rs
```diff
@@ -1152,105 +1152,57 @@ impl CompositeDevice {
...
-                    if self.active_inputs.contains(&c) {
-                        log::trace!("Removed inactive capability: {c:?}");
-                        let index = self.active_inputs.iter().position(|r| r == &c).unwrap();
-                        self.active_inputs.remove(index);
+                    Gamepad::Axis(_)
+                    | Gamepad::Trigger(_)
+                    | Gamepad::Accelerometer
+                    | Gamepad::Gyro => {}
+                },
+                Capability::Mouse(t) => match t {
```
ShadowApex, line 1192: Use `Capability::Mouse(ref t)` to avoid needing to `cap.clone()`

src/input/composite_device/mod.rs
```diff
@@ -1152,105 +1152,57 @@ impl CompositeDevice {
             }
         }
 
-        for mut event in events {
+        for event in events {
             let cap = event.as_capability();
 
             // Track what is currently active so we can ignore extra events.
-            match cap {
+            match cap.clone() {
```
ShadowApex, line 1159:
```suggestion
            match cap {
```

src/input/composite_device/mod.rs
```diff
@@ -1152,105 +1152,57 @@ impl CompositeDevice {
...
-                    if self.intercept_active_inputs.contains(&cap) {
-                        log::debug!("The event is already in the list. Skipping.");
-                        continue;
-                    };
-                    self.intercept_active_inputs.push(cap.clone());
-
-                    if self.intercept_active_inputs.len() != self.intercept_activation_caps.len() {
-                        log::debug!("More events needed to activate intercept mode.");
+                    if !self
+                        .check_intercept_keys(event.clone(), is_pressed, intercept)
```
ShadowApex, line 1169: We should borrow `event` here instead of cloning it to avoid cloning all button events.

ShadowApex (changes requested):

ShadowApex (approved):

PR #75, opened 2024-04-30
WIP: feat(Touchpad): Add Touchpad event type.

pastaq:


src/input/capability.rs
```diff
@@ -1048,3 +1050,105 @@ impl FromStr for Keyboard {
...
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum TouchpadAxis {
+    LeftPad,
+    RightPad,
+    CenterPad,
+    MultiTouch,
```
pastaq, line 1128: These are placeholders until we figure out the best way to handle multitouch.

src/input/event/value.rs
```diff
@@ -32,6 +32,12 @@ pub enum InputValue {
         y: Option<f64>,
         z: Option<f64>,
     },
+    TouchValue {
+        index: u64,
+        pressed: bool,
```
ShadowApex, line 37: I think we should rename this to `is_touching` instead of `pressed`, so we don't confuse it with a touchpad "press" (which would be like clicking the touchpad on the Steam Deck)

src/input/event/value.rs
```diff
@@ -101,6 +113,11 @@ impl InputValue {
                             },
                             // Gamepad Button -> Keyboard
                             Capability::Keyboard(_) => Ok(self.clone()),
+                            Capability::Touchpad(touch) => match touch {
+                                Touchpad::LeftPad(_) => todo!(),
+                                Touchpad::RightPad(_) => todo!(),
+                                Touchpad::CenterPad(_) => todo!(),
```
ShadowApex, line 119: We should make sure to replace these `todo!()` with `Err(TranslationError::NotImplemented)` so the thread doesn't crash if someone tries to do this translation in an input profile.

src/input/source/hidraw/lego.rs
```diff
@@ -114,7 +115,10 @@ fn normalize_axis_value(event: event::AxisEvent) -> InputValue {
             let y = normalize_signed_value(value.y as f64, min, max);
             let y = Some(-y); // Y-Axis is inverted
 
-            InputValue::Vector2 { x, y }
+            // Pass the index as z value
+            let z = Some(value.index as f64);
+
+            InputValue::Vector3 { x, y, z }
```
ShadowApex, line 121: We want to use `InputValue::TouchValue` for this now, right?

src/input/source/hidraw/steam_deck.rs
```diff
@@ -17,14 +17,18 @@ use tokio::sync::{
 use crate::{
     drivers::{
         dualsense::hid_report::SetStatePackedOutputData,
+        lego::driver,
```
ShadowApex, line 20: Why are we importing the lego driver in the steam deck driver? o.O

pastaq: Automatic import. Good catch

pastaq:
Resolved in #91

PR #81, opened 2024-05-02
feat(CompositeDevice): add interface to set target devices at runtime

ShadowApex:
- Add create target device
- Add attach target device
- Add set target device
- Add profile config for setting target input devices

### DIFF
--- a/rootfs/usr/share/inputplumber/profiles/default.yaml
+++ b/rootfs/usr/share/inputplumber/profiles/default.yaml
@@ -8,6 +8,12 @@ kind: DeviceProfile
 # Name of the device profile
 name: Default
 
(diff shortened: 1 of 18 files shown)


pastaq:
We should consider emitting a dbus signal when this is done so client programs can know if they need to do anything after the device switches. Updated glyphs for example.

ShadowApex:
> We should consider emitting a dbus signal when this is done so client programs can know if they need to do anything after the device switches. Updated glyphs for example.

We can do that, but switching the target device(s) is already going to be initiated by the client either by calling `SetTargetDevices` or through loading an input profile. We should also not be updating glyphs based on the target device(s), and instead we should be updating them based on the source device(s). E.g. the UI will show the glyphs for the real controller instead of the virtual controller.

pastaq (approved):

PR #85, opened 2024-05-06
feat(CompositeDevice): add get device profile name and target capabilities

ShadowApex:
Add the ability to fetch the Composite Device profile name and all target capabilities.


src/input/composite_device/mod.rs
```diff
@@ -102,7 +104,16 @@ impl DBusInterface {
     /// Name of the composite device
     #[dbus_interface(property)]
     async fn name(&self) -> fdo::Result<String> {
-        Ok("CompositeDevice".into())
```
pastaq, line 105: This is how we were doing this? lol

pastaq (approved):
lgtm

src/input/composite_device/mod.rs
ShadowApex: Haha, yep!

PR #90, opened 2024-05-07
fix(DBus): seperate DBus interfaces into their own module

ShadowApex:
This change moves all of the DBus interface code into their own separate files (and updates it to zbus 4.x format). This can let us re-use various DBus interfaces multiple times. Such as using a single interface for `org.shadowblip.Input.Gamepad` across all target devices.


ShadowApex:
Note that this change currently just adds these interfaces, but are not used yet. Once we have integrated #75, we can use these.

pastaq (approved):

PR #94, opened 2024-05-09
fix(CompositeDevice): use recv_many to batch process the queue

ShadowApex:


pastaq:
Do we want to increase the scope of this PR to have more mitigations? This doesn't solve the core issue entirely.

pastaq (approved):
We can merge this, but should look into what is blocking before we merge #101 or the problem will get even worse

Issue #76, opened 2024-05-01
Issue: ROG Ally sometimes captured before the controller is in full Gamepad mode


pastaq:
The ROG Ally has a unique interface with multiple operating modes. The cold boot default mode is a dekstop mode with multiple shortcuts programmed into various combinations using the back paddles as the trigger under the same keycode for both. hid-Asus manages this device in Linux and during startup sets the device into gamepad mode, which ensures the back paddles are independent buttons, and that they don't cancel existing inputs. If the device is grabbed to early (presumably, this could be a driver issue) the hid-asus driver will fail to set the mode, the buttons will not be independent, and they will cancel any held input. As an example, if you were to set the back paddles to be the gear shift up/down in a game and the right trigger as your accelerator, during normal mode this would work as you would expect. Back paddles would permit both up and down shifting, and the accelerator input would not be affected by the buttons. In the controllers desktop mode however you would only have down shift available and the trigger would be canceled as if fully released and gameplay is interrupted.  This is a problem even if you don't map the back paddles as they would still interrupt any held axis or button.

pastaq:
Tagging @flukejones as he is the driver maintainer and may have some insight.

flukejones:
I would suggest polling the sysfs item for mode change as internally the driver polls the device for change and setup completing

pastaq:
```
[    4.086787] asus_wmi: ASUS WMI generic driver loaded
[    4.310820] asus 0003:0B05:1ABE.0001: Setting up ROG USB interface
[    4.310822] asus 0003:0B05:1ABE.0001: Setting up ROG Ally interface
[    4.315389] asus 0003:0B05:1ABE.0001: ROG Ally not ready, retry 0
[    4.321398] asus 0003:0B05:1ABE.0001: ROG Ally not ready, retry 1
[    4.327388] asus 0003:0B05:1ABE.0001: ROG Ally not ready, retry 2
[    4.327390] asus 0003:0B05:1ABE.0001: ROG Ally never responded with a ready
```
This repeats multiple times until it times out. We'll need to figure out a device specific quirk to poll for the mode change ans recommended.

pastaq:
This only happens on a fresh boot. On restart or suspend/resume the device behaves as expected.

flukejones:
[...]

PR #96, opened 2024-05-09
Update ally_type1.yaml

kzkzkzkzkzk:
By setting long press of AC/CC button as back button 2, you can assign it to each game. It will be recognized as R4/L4.

### DIFF
--- a/rootfs/usr/share/inputplumber/capability_maps/ally_type1.yaml
+++ b/rootfs/usr/share/inputplumber/capability_maps/ally_type1.yaml
@@ -26,7 +26,7 @@ mapping:
       - keyboard: KeyDelete
     target_event:
       gamepad:
-        button: Keyboard
+        button: LeftPaddle2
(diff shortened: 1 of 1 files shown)


kzkzkzkzkzk:
In rogally, there is a bug where the back buttons are unified to the right side when starting up after shutting down. This long press is mapped as keyboard and quick access 2 by default, but by setting it to back button 2, all buttons can be assigned.

pastaq:
We won't be doing this. This is a workaround of the problem rather than a true fix that will break our ability to open the quick bar in steam plus as this device doesn't have a dedicated guide button that can be held. 

Instead we simply need to either wait for the driver so set up the device properly or try to set it up ourselves.

This does indicate that we need to swap which paddles are activated by paddle1/2 as 1 should be l/r4 and 2 should be l/r 5

pastaq:
you can follow the development of the proper resolution to this issue in #76

PR #107, opened 2024-05-27
Fix Onexplayer AMD being recognized as intel model

alkazar:
My personal Onexplayer AMD was being matched as an intel model and had incorrect mappings. This fixes it.


pastaq:
If this was positively matching on Intel we need to ensure that only Intel devices match on Intel as well.

alkazar:
Unfortunately, I don't have any OXP Intel devices, or any other OXP devices for that matter.

ShadowApex (approved):

PR #120, opened 2024-06-08
fix(Device): Add OrangePi NEO mount matrix.

pastaq:


pastaq:
Addressed in #121

philmmanjaro:
Would be nice to be merged before #121

pastaq:
> Would be nice to be merged before #121

We could do that, but I think it would be better to amend #121 slightly to comment out grabbing the touchpads for now. I don't want to have to rebase later if we end up touching some parts of the same code. That will have the same effect but keep the pending PR's sitting in bitrot down.

Issue #119, opened 2024-06-08
Feature: Add touchpad support to the OranePi NEO


pastaq:
The OrangePi Neo has two touchpads, one on either side. They have different names 

Touchpad 1:
```
[gamer@opineo ~]$ udevadm info --attribute-walk --path=$(udevadm info --query=path --name=/dev/hidraw1)

Udevadm info starts with the device specified by the devpath and then
walks up the chain of parent devices. It prints for every device
found, all possible attributes in the udev rules key format.
A rule to match, can be composed by the attributes of the device
and the attributes from one single parent device.

  looking at device '/devices/platform/AMDI0010:00/i2c-0/i2c-OPI0001:00/0018:0911:5288.0002/hidraw/hidraw1':
    KERNEL=="hidraw1"
    SUBSYSTEM=="hidraw"
    DRIVER==""
    ATTR{power/control}=="auto"
    ATTR{power/runtime_active_time}=="0"
    ATTR{power/runtime_status}=="unsupported"
    ATTR{power/runtime_suspended_time}=="0"

  looking at parent device '/devices/platform/AMDI0010:00/i2c-0/i2c-OPI0001:00/0018:0911:5288.0002':
    KERNELS=="0018:0911:5288.0002"
    SUBSYSTEMS=="hid"
    DRIVERS=="hid-multitouch"
    ATTRS{country}=="00"
    ATTRS{power/control}=="auto"
    ATTRS{power/runtime_active_time}=="0"
    ATTRS{power/runtime_status}=="unsupported"
    ATTRS{power/runtime_suspended_time}=="0"
    ATTRS{quirks}=="334864"

  looking at parent device '/devices/platform/AMDI0010:00/i2c-0/i2c-OPI0001:00':
    KERNELS=="i2c-OPI0001:00"
    SUBSYSTEMS=="i2c"
    DRIVERS=="i2c_hid_acpi"
    ATTRS{name}=="OPI0001:00"
    ATTRS{power/control}=="auto"
    ATTRS{power/runtime_active_time}=="0"
    ATTRS{power/runtime_status}=="unsupported"
    ATTRS{power/runtime_suspended_time}=="0"
    ATTRS{power/wakeup}=="enabled"
    ATTRS{power/wakeup_abort_count}=="0"
    ATTRS{power/wakeup_active}=="0"
    ATTRS{power/wakeup_active_count}=="1605"
    ATTRS{power/wakeup_count}=="0"
[...]

PR #121, opened 2024-06-08
feat(Device): Add support for left/right touchpad events for the OrangePi Neo.

pastaq:


pastaq:
Resolves #119

philmmanjaro:
I would rather fix the mount_matrix on a separate PR as intended before and wait with this PR until we have a solution to use the touchpads as Mouses when the Gamepad profile doesn't support touch in its controller. This would be Xbox Controller as an example. Also when switching on desktop mode the regular mouse is gone.

We might want to find a solution also to only activate them when running a steam session or gamescope. Has some aspects we can discuss ...

pastaq:
Don't merge until we can translate Touchpad events to all other relevant events (Mouse, Joystick, add a virtual touchpad device)

pastaq:
I modified the device profile for the OrangePi NEO so it wont capture the touchpads. These lines can be uncommented once we add touchpads to the schema and/or add the steam deck controller as the default target. This removes the need to wait for this PR and makes #120 irrelevant.

ShadowApex (approved):

philmmanjaro:
OK, is a good compromise for now.

PR #122, opened 2024-06-12
fix(devices): Ignore devices that aren't specified.

pastaq:
- We should ignore anything we don't have a specific configuration for until we have more implementations for unusual devices. This will prevent loads of bug reports of controllers not working.

### DIFF
--- a/rootfs/usr/share/inputplumber/devices/60-xbox_360_gamepad.yaml
+++ b/rootfs/usr/share/inputplumber/devices/60-xbox_360_gamepad.yaml
@@ -1,0 +1,182 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ShadowBlip/InputPlumber/main/rootfs/usr/share/inputplumber/schema/composite_device_v1.json
+# Schema version number
+version: 1
+
+# The type of configuration schema
+kind: CompositeDevice
+
+# Name of the composite device mapping
+name: Xbox 360 Gamepad
+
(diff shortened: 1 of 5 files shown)


ShadowApex:
We should add configs for gamepads we know work, then we can include this change (or remove `70-generic-gamepads.yaml`)

rootfs/usr/share/inputplumber/devices/60-xbox_360_gamepad.yaml
```diff
@@ -0,0 +1,272 @@
...
+
+  - group: gamepad
+    evdev:
+      vendor_id: "0x0079"
+      product_id: "0x18d4"
+
+  - group: gamepad
+    evdev:
+      vendor_id: "0x03eb"
+      product_id: { "0xff01", "0xff02" }
```
ShadowApex, line 31: This does not conform to the YAML schema. If you were trying to glob match on the id, it would be like:

```yaml
product_id: "{0xff01,0xff02}"
```

From the Legion Go profile, it also looks like there's no `0x` prefix the way we're parsing the product id.

ShadowApex (changes requested):

rootfs/usr/share/inputplumber/devices/60-xbox_360_gamepad.yaml
pastaq: -_-

ShadowApex (approved):

PR #124, opened 2024-06-14
fix(LegionGo): Fix input when a single controller is detached.

pastaq:
- Uses the DInput report when one of the two controllers is disconnected as the device driven by the xpad driver is not present in that mode. 
- Fix the Dinput report PackedStruct
- Convert DInput axes into xinput axes ( we could reverse this math to add more fidelity in Dinput mode vice less)
- Capture more devices on the Legion go.
- Remove some types and code we are never going to use, like accel data from the controller HID report.
- Add safe default for IMU polling in the event the sample rate is 0.


pastaq:
While researching this I discovered the IMU is exposed as an HIDRAW report. That might work better than using the IIO interface. The IIO interface doesn't work. It could be because the sample rate is 0.00~ as I am able to watch cat the angvel sysfs just fine.

PR #126, opened 2024-06-17
fix(Capability): Block Sync events

pastaq:
Sync events do not need to be processed. All target devices submit a sync event with every emitted event and processing them adds overhead to emitting an superflous event. Block them instead.

### DIFF
--- a/src/input/composite_device/mod.rs
+++ b/src/input/composite_device/mod.rs
@@ -605,7 +605,10 @@ impl CompositeDevice {
 
         // Only send valid events to the target device(s)
         if cap == Capability::NotImplemented {
-            log::trace!("Refusing to send 'NotImplemented' event to target devices");
+            log::trace!(
+                "Refusing to send '{}' event to target devices.",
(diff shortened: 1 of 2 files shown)


src/input/composite_device/mod.rs
```diff
@@ -604,8 +604,11 @@ impl CompositeDevice {
         //log::trace!("Event capability: {:?}", cap);
 
         // Only send valid events to the target device(s)
-        if cap == Capability::NotImplemented {
-            log::trace!("Refusing to send 'NotImplemented' event to target devices");
+        if cap == Capability::NotImplemented || cap == Capability::Sync {
```
ShadowApex, line 607: Instead of blocking the events at the composite device, why don't we instead block them in the `evdev` source device? That will prevent those events from filling up the channel buffer.

pastaq: Updated

ShadowApex (approved):

PR #129, opened 2024-06-21
feat(Touchscreen): add support for touchscreen devices

ShadowApex:


pastaq (approved):

src/input/event/dbus.rs
```diff
@@ -335,8 +359,8 @@ fn actions_from_capability(capability: Capability) -> Vec<Action> {
             Keyboard::KeyInsert => vec![Action::None],
             Keyboard::KeyDelete => vec![Action::None],
             Keyboard::KeyMute => vec![Action::VolumeMute],
-            Keyboard::KeyVolumeDown => vec![Action::VolumeUp],
-            Keyboard::KeyVolumeUp => vec![Action::VolumeDown],
+            Keyboard::KeyVolumeDown => vec![Action::VolumeDown],
```
pastaq, line 362: Lol

rootfs/usr/share/inputplumber/devices/50-steam_deck.yaml
```diff
@@ -24,14 +24,30 @@ matches:
...
+    evdev:
+      name: AT Translated Set 2 keyboard
+      phys_path: isa0060/serio0/input0
 
 # The target input device(s) that the virtual device profile can use
 target_devices:
   - gamepad
   - mouse
   - keyboard
+  - touchscreen
```
pastaq, line 53: I think we should be more specific as to the model here. When I add the Legion Go and OrangePi Neo touchscreens, as well as evdev only touchscreens, we will need to have more than one option available. This generic name is sub-optimal when it's not universal.

PR #131, opened 2024-06-24
fix(LegionGo): Add accel_gyro_3d implementation for IIO IMU driver.

pastaq:
- Adds accel_gyro_3d implementation of IIO IMU driver. This implementation doesn't support setting scale or sample frequency and has a different scaling factor for the gyro and accelerometer than BMI devices.
- Renames IIO IMU to BMI IMU for disabiguation and improves debug statements.
- Adds mount matric for LegionGo display IMU.

### DIFF
--- a/src/input/source/iio.rs
+++ b/src/input/source/iio.rs
@@ -1,9 +1,10 @@
-pub mod iio_imu;
+pub mod accel_gyro_3d;
+pub mod bmi_imu;
 
 use std::error::Error;
 
 use glob_match::glob_match;
-use tokio::sync::{broadcast, mpsc};
+use tokio::sync::mpsc;
 use zbus::{fdo, Connection};
 use zbus_macros::dbus_interface;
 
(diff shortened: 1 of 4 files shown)


src/input/source/iio.rs
```diff
@@ -119,15 +120,26 @@ impl IIODevice {
...
+                self.info.clone(),
+                self.config.clone(),
+                tx,
+                rx,
+                self.get_id(),
+            );
+            driver.run().await?;
+
+        // AccelGryo3D IMU Driver
+        } else if glob_match("{gyro_3d,accel_3d}", name.as_str()) {
```
ShadowApex, line 138: Maybe we can create an enum and write a small function to get the IMU type like we did for hidraw?

ShadowApex (approved):

src/input/source/iio.rs
pastaq: Done

ShadowApex (approved):

PR #133, opened 2024-06-28
refactor(CompositeDevice): add CompositeDeviceClient

ShadowApex:
This change adds a new `CompositeDeviceClient` to simplify communicating with a `CompositeDevice`. It provides helper methods for sending and receiving data over a channel.

For example, instead of doing:

```rust
tx.send(Command::ProcessEvent(device_id, event).await?;
```

we can now do:

```rust
composite_device.process_event(device_id, event).await?;
```

This is especially helpful when getting data back from a composite device.


pastaq (approved):
Uh, sure.

PR #147, opened 2024-07-11
feat(Touch): add basic touch input translation support

ShadowApex:
This change adds two new config entries for input profiles: `touchpad` and `touchscreen`.

To add these translations, I had to:
* Update the `CapabilityConfig` struct with the new translation config
* Update the device profile schema with the new config fields
* Update `From<CapabilityConfig> for Capability` to convert a translation config into a capability
* Update `InputValue::translate` to translate the input value to the target

### DIFF
--- a/src/input/capability.rs
+++ b/src/input/capability.rs
@@ -63,6 +63,12 @@ impl FromStr for Capability {
             "DBus" => Ok(Capability::DBus(Action::from_str(
                 parts.join(":").as_str(),
             )?)),
+            "Touchpad" => Ok(Capability::Touchpad(Touchpad::from_str(
+                parts.join(":").as_str(),
+            )?)),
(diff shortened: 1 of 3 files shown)


Ericky14 (approved):

src/input/capability.rs
```diff
@@ -63,6 +63,12 @@ impl FromStr for Capability {
             "DBus" => Ok(Capability::DBus(Action::from_str(
                 parts.join(":").as_str(),
             )?)),
+            "Touchpad" => Ok(Capability::Touchpad(Touchpad::from_str(
+                parts.join(":").as_str(),
```
pastaq, line 67: clippy will complain about using a string here, use `'` instead of `"` for char.

pastaq (approved):

src/input/capability.rs
ShadowApex: I get errors if I try to do that:
image

PR #152, opened 2024-07-17
feat(DS5): add support for Sony DS5 source device

ShadowApex:


src/drivers/dualsense/driver.rs
```diff
@@ -1,14 +1,31 @@
...
+    event::{AccelerometerEvent, AccelerometerInput, AxisEvent, AxisInput, Event, TouchAxisInput},
+    hid_report::PackedInputDataReport,
+};
+
 // Source: https://github.com/torvalds/linux/blob/master/drivers/hid/hid-playstation.c
 pub const DS5_EDGE_NAME: &str = "Sony Interactive Entertainment DualSense Edge Wireless Controller";
 pub const DS5_EDGE_VERSION: u16 = 256;
-pub const DS5_EDGE_VID: u16 = 0x054C;
-pub const DS5_EDGE_PID: u16 = 0x0DF2;
+pub const DS5_EDGE_VID: u16 = 0x054c;
```
pastaq, line 19: This can be deleted

src/input/event/evdev.rs
```diff
@@ -617,6 +617,7 @@ fn event_codes_from_capability(capability: Capability) -> Vec<u16> {
                 GamepadButton::RightTop => vec![],
                 GamepadButton::RightTrigger => vec![KeyCode::BTN_TR2.0],
                 GamepadButton::Screenshot => vec![KeyCode::KEY_RECORD.0],
+                GamepadButton::Mute => vec![KeyCode::KEY_MUTE.0],
```
pastaq, line 620: Was this verified as the target event in evtest?

src/input/source/hidraw/steam_deck.rs
pastaq, line 1: Why are we touching this file in a ds5 pr?

src/input/event/evdev.rs
ShadowApex: In evtest, no event is emitted at all. Open to suggestions if we want this to translate to evdev at all.

src/input/source/hidraw/steam_deck.rs
pastaq: Well the udev branch is going to touch these devices FYI, it's already going to be a big thing to manage the merge conflicts as it is going to touch the entire source side of the program. If we could keep things in scope until that is complete I'd appreciate it.

rootfs/usr/share/inputplumber/devices/60-ps5_ds_gamepad.yaml
```diff
@@ -17,29 +17,30 @@ single_source: false
...
+    hidraw:
+      vendor_id: 0x054c
+      product_id: 0x0ce6
+
+  # Block the evdev implementations
+  - group: gamepad
+    blocked: true
     evdev:
-      name: Sony Interactive Entertainment DualSense Wireless Controller
+      name: "{Sony Interactive Entertainment DualSense Wireless Controller,DualSense Wireless Controller}"
```
pastaq, line 29: While it might not matter, this is going to attach any additional evdevs from additional controllers to the first controller. Given that there is *some* processing that occurs for each event, even if blocked, this could negatively impact latency on the first controller compared to the others.

src/input/event/evdev.rs
pastaq: It think it's intended to be used with a headset attachment for the controller. Maybe we could only send this event out of the driver if no attachment is detected in the Input state.

src/input/manager.rs
```diff
@@ -1159,9 +1181,57 @@ impl Manager {
         self.source_device_dbus_paths.insert(id.clone(), path);
 
         // Check to see if the device is virtual
+        // TODO: Check if device exists in bluez
```
pastaq, line 1184: TODONE:

rootfs/usr/share/inputplumber/devices/60-ps5_ds_gamepad.yaml
ShadowApex: Blocked source devices are only hidden and never run. So there will be no event processing for blocked devices.

ShadowApex: Also, these are set to `unique: true`, so only one will be added to each composite device.

pastaq (approved):

PR #160, opened 2024-07-24
feat(SteamDeckTarget): add Steam Deck Controller target device through USB emulation

ShadowApex:
Tested with Sony DualSense source device.

image

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ thiserror = "1.0.61"
 tokio = { version = "*", features = ["full"] }
 udev = { version = "^0.8", features = ["mio"] }
 uhid-virt = "0.0.7"
+virtual-usb = { git = "https://github.com/ShadowBlip/virtual-usb-rs.git", rev = "5a7a96a6aedc54f339d9ebff78bf484e5b17728d" }
 zbus = { version = "4.3.1", default-features = false, features = ["tokio"] }
 zbus_macros = "4.3.1"
 
--- a/rootfs/usr/share/inputplumber/devices/50-legion_go.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-legion_go.yaml
@@ -151,6 +151,6 @@ source_devices:
 
 # The target input device(s) that the virtual device profile can use
 target_devices:
-  - xbox-elite
+  - deck
   - mouse
   - keyboard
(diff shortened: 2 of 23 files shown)


pastaq (approved):
Works as expected on the Legion Go and OrangePi Neo. We should verify stability before merging.

philmmanjaro:
Will test this also on my end and will give feedback. Might also tell the vendor to test on their end at factory level.

philmmanjaro:
@pastaq initial feedback on the Neo 8840u: I can see the SteamDeck controller after adding the pull requests to InputPlumber and OpenGamepadUI. When I go to controller settings and begin testing I see buttons working and triggers too. Visual feedback on the SteamDeck controller picture is not always given. Pressed buttons stay highlighted on the picture. When using the touchpads as trackpads no feedback is given in the UI. However I see a regular mouse movement and the colours switch to desktop mode as if you use a mouse.

We might do debug this more.

philmmanjaro:
After adding [a.patch](https://github.com/user-attachments/files/16371800/a.patch.txt) the touchpads got detected as Steam Controller trackpads. I assume OrangePi didn't adopt their device renaming on the current firmware I test on my 8840u device.

philmmanjaro:
QAM button acts as GUIDE button for some odd reason unless I have to learn on how the steam buttons should work. If I select battery symbol and open it once I can shuffle between left and QAM menu by pressing button 6 with the gamepad symbol on the Neo.

philmmanjaro:
For some reason **Street Fighter 6 Demo** didn't work in SteamDeck profile. When switching to DS5 or X it worked.

pastaq:
> QAM button acts as GUIDE button for some odd reason unless I have to learn on how the steam buttons should work. If I select battery symbol and open it once I can shuffle between left and QAM menu by pressing button 6 with the gamepad symbol on the Neo.

We need to add a different capability map for the deck target. currently it gets converted into the chord for qam, which is guide + a. If we remove that it will work as expected but break qam for other targets. This is a WIP.

pastaq (approved):
Send it

philmmanjaro:
I noticed that sleep is totally broken when enabling the profile. More info in the issue: https://github.com/ShadowBlip/InputPlumber/issues/165

pastaq:
> I noticed that sleep is totally broken when enabling the profile. More info in the issue: #165

You didn't have to post this multiple times.

PR #166, opened 2024-07-30
feat(XBOne BT): add support for XBox One Bluetooth controllers

ShadowApex:


pastaq:
What is the plan for rumble support?

ShadowApex:
We can add rumble support in a later PR. We'll need to look into the format of the output reports.

PR #171, opened 2024-08-02
fix(ForceFeedback): add default implementation for FF that does not error

ShadowApex:
This fixes a bug where the target device would stop if an error occurred when uploading a force feedback event to source devices. Force feedback uploads are non-fatal now.


pastaq (commented):
LGTM

PR #183, opened 2024-08-09
fix(Suspend/Resume): Remove resume service

### LINKED ISSUE #77: Issue: Input is not picked up after resume event
After a suspend/resume cycle, there is no output to the system from the virtual device. trace logs show the events being written to the target device, but evtest confirms no output from the target device. Restarting the service returns full functionality. Confirmed on the Legion Go, ROG Ally, and AYANEO FLIP all using the DualSense or DualSense Edge profile.

pastaq:
- Removes inputplumer-resume.service. Testing on v0.34.2 shows the service is no longer needed for device input to be registered after a resume event. Specific reasons for the behavior change are not known, but I suspect a commbination of using udev and unique ID's for the DS5 has resovled the issue.

Fixes #77


ShadowApex (changes requested):
We also need to remove this service reference from `Cargo.toml` and `pkg/rpm/inputplumber.spec`

ShadowApex (approved):

Issue #172, opened 2024-08-03
Anbernic Win600 Buttons Not Working


Ponjen:
*This issue report originated from ChimeraOS, but was told to report here as well (https://github.com/ChimeraOS/chimeraos/issues/1003)*

For the Anbernic Win600, the Home, Windows Key button (Quick Access), and keyboard button don't work. Even trying to manually map them in the controller configuration doesn't work, as Steam doesn't recognize them. Volume keys and power tap to sleep works.

fewtarius:
The Win600 does not correctly emulate an Xbox controller, minimally the xpad driver needs to be patched.

See this repo for reference: https://github.com/lualiliu/holoiso-win600-driver

murilo-moraes:
Same issue here for my Win600. Were you able to patch the xpad driver? Here the build folder for kernel modules is missing.

pastaq:
> The Win600 does not correctly emulate an Xbox controller, minimally the xpad driver needs to be patched.
> 
> See this repo for reference: https://github.com/lualiliu/holoiso-win600-driver

This is new information to me. I thought the custom anbernic driver was just to grab one of the buttons as home. Does using the udev rule for a new xpad device not work?

uejji:
#186 will hopefully help with this.  Needs more work to be fully supported, though.

pastaq:
What specifically is not working with that patch applied?

Ponjen:
> What specifically is not working with that patch applied?

Doesn't seem possible to install the patch on ChimeraOS, at least not as a typical user. It's due to the read only file system, and I can't find any info of a command to disable it like SteamOS.

uejji:
> What specifically is not working with that patch applied?

Let's consider the following capability map I've worked up for Win600 for testing (not with Lualiliu's patch):

`# yaml-language-server: $schema=https://raw.githubusercontent.com/ShadowBlip/InputPlumber/main/rootfs/usr/share/inputplumber/schema/capability_map_v1.json
# Schema version number
version: 1

# The type of configuration schema
kind: CapabilityMap

# Name for the device event map
name: Anbernic Type 1

# Unique identifier of the capability mapping
id: anb1

# List of mapped events that are activated by a specific set of activation keys.
mapping:
  - name: Home (Short Press)
    source_events:
      - keyboard: KeyLeftMeta
      - keyboard: KeyG
    target_event:
      gamepad:
        button: Guide
  - name: Home (Long Press)
    source_events:
[...]

PR #186, opened 2024-08-13
Update sys_vendor for Win600

uejji:
sys_vendor value is initial caps on both of my units.  Unless there are some units that report differently?


pastaq:
I don't have hardware to verify so you're in a better position to confirm than I am. Thanks.

pastaq (approved):

uejji:
Just also submitted an updated capability map for Win600.  It's not perfect, but it should allow Win600 users most of the functionality provided by InputPlumber.  Windows key short press is still a problem point, however.

pastaq:
Fixes #172

PR #187, opened 2024-08-13
fix(Power Button): add power button support over dbus

ShadowApex:


pastaq:
We need to have a larger conversation about this. Currently ChimeraOS uses https://github.com/ShadowBlip/steam-powerbuttond (and we have been recommending that to other folks as well). Steam's suspend/resume/power menu functionality is done entirely in unprivileged userspace for the current user as it needs to point to the steam binary in ~/.local 

We don't want to add hacky workarounds like HandyGCCS had to access user directory as root. Steam currently doesn't use dbus for this functionality either, and not everyone wants to use opengamepadui with steam. If we're going to do this we should include a user level systemd daemon for receiving the dbus signals that can then route the command as needed (OpenGamepadUI, call Steam, or Playtron). That client should probably default to steam as it won't have any interaction and the client software could change its settings.

PR #189, opened 2024-08-14
fix(GPD): Add mount matrix for GPD Win Mini

justinweiss:
This mount matrix felt right on the Win Mini 2023 playing games through Steam Input with the ds5 selected.

I've heard the 2024 has the same DMI device entries but may have the IMU mounted differently -- not sure what could be done about that.

### DIFF
--- a/rootfs/usr/share/inputplumber/devices/50-gpd_winmini.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-gpd_winmini.yaml
@@ -35,6 +35,10 @@ source_devices:
   - group: imu
     iio:
       name: i2c-BMI0160:00
+      mount_matrix:
+        x: [1, 0, 0]
+        y: [0, -1, 0]
+        z: [0, 0, -1]
 
 # Optional configuration for the composite device
 options:
--- a/rootfs/usr/share/inputplumber/devices/50-gpd_winmini.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-gpd_winmini.yaml
@@ -37,8 +37,8 @@ source_devices:
       name: i2c-BMI0160:00
       mount_matrix:
         x: [1, 0, 0]
-        y: [0, -1, 0]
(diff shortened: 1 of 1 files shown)


pastaq:
I've seen a few different ways of identifying them. I think the BIOS version in DMI might be unique for each.

justinweiss:
Here's the DMI info for my 2023 model. I updated the BIOS on this, I think there are at least two other BIOS versions for the 2023.
[dmi.txt](https://github.com/user-attachments/files/16615231/dmi.txt)

jim-80:
[dmi.txt](https://github.com/user-attachments/files/16615539/dmi.txt)
This is the output for the 2024 Win Mini

pastaq:
Thanks. Doesn't look like there's a predictable DMI string we can use. I'll look into how some other tools are determining this.

pastaq:
HandheldCompanion matches on the processor model.
https://github.com/CasperH2O/HandheldCompanion/blob/a719c46b7ad3c11b90d52323e6c74319b706c2ad/HandheldCompanion/Devices/IDevice.cs#L347-L354

We should definitely do that here as well. Looking at our schema, we don't currently support this in the configuration files.

pastaq:
Can you update the mount matrix to coincide with this PR?
https://github.com/ShadowBlip/InputPlumber/pull/220#issuecomment-2435668917

Essentially we want the "Deck" target to produce correct animations and the Dualsense target to not, as this will provide natural movement in game by default for both targets.

justinweiss:
:+1: , I'll take a look. It'll have to change a bit anyway, since the upstream driver doesn't pull the mount matrix from ACPI yet.

justinweiss:
I updated the mount matrix to coincide with the Deck target on the current ChimeraOS unstable.

pastaq (approved):

pastaq (approved):
We can merge for now and create an issue for separating the models.

PR #194, opened 2024-08-19
fix(XPad Bluetooth): resolve race condition if a device is quickly added, removed, and added again

ShadowApex:
When connecting an XBox Series controller using bluetooth, this behavior was observed:

```
[2024-08-19T19:33:06Z DEBUG inputplumber::watcher] inotify CREATE: Some("hidraw4")
[2024-08-19T19:33:06Z DEBUG inputplumber::watcher] inotify DELETE: Some("hidraw4")
[2024-08-19T19:33:06Z DEBUG inputplumber::watcher] inotify CREATE: Some("hidraw4")
```

This would ultimately result in the `CompositeDevice` stopping when it received the `DELETE` event because no source devices would remain.

Normally this would be fine, and we'd expect the second `CREATE` event to create a new `CompositeDevice`. But because the input `Manager` and `CompositeDevice` run as separate tasks, processing at different speeds, we'd run into a situation where the `CompositeDevice` has not had a chance to actually stop before the `Manager` processes the second `CREATE` event. 

As a consequence, the `Manager` would see that a `CompositeDevice` still exists for this hidraw device, so it would send a command to the existing `CompositeDevice` asking it to manage that source device instead of creating a new `CompositeDevice`. Meanwhile, as soon as the `CompositeDevice` receives the `DELETE` event, the device will stop, even if a `CREATE` event is further in its queue.

This change resolves this issue by only actually stopping the `CompositeDevice` after processing all the commands in its queue.

### DIFF
--- a/src/drivers/xpad_uhid/driver.rs
+++ b/src/drivers/xpad_uhid/driver.rs
@@ -104,7 +104,7 @@ impl Driver {
 
         let events = match report_id {
             DATA => {
-                log::debug!("Got touch data.");
+                log::debug!("Got input data.");
                 if bytes_read != PACKET_SIZE {
                     return Err("Invalid packet size for Keyboard or Touchpad Data.".into());
                 }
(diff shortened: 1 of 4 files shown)


pastaq (approved):

pastaq:
Looks good.

Related note, I'm testing the xpadneo driver right now. It might be something Playtron should look into using. It is only for the BT mode of Xbox style controllers and seems to work OOTB to create an evdev. If we don't have to implement a driver for every BT device that would be ideal IMO. The only downside is that every device appears as an xb360 device usign the standard 0x28e product ID. I might make a PR to the driver to address that so things like the series and elite 2 are able to be discerned. If we use that, the product ID will need to be in the elite 2 config or we risk losing capabilities for those gamepads. It shouldn't conflict though as the xpadneo driver unloads the hid driver, so we'll lose the HIDRAW node.

PR #196, opened 2024-08-26
feat(Evdev): expose source device capabilities over dbus

ShadowApex:
This change exposes source evdev device capabilities over DBus, providing an interface for applications to query what kind of input devices are connected and what they are capable of (i.e. checking if a keyboard is plugged in). The values returned are in the form of Linux event codes that the device supports.

image

This change also exposes udev properties over DBus:
image

### DIFF
--- a/src/dbus/interface/source/evdev.rs
+++ b/src/dbus/interface/source/evdev.rs
@@ -1,20 +1,33 @@
-use std::error::Error;
+use std::{collections::HashMap, error::Error};
 
+use evdev::EventType;
 use zbus::{fdo, Connection};
 use zbus_macros::interface;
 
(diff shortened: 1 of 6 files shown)


ShadowApex:
After some thought, I decided to pull out the shared udev functionality into its own interface that can be used on all device types. We can now define device type-specific properties/methods in their respective interfaces, and use the `SourceUdevDevice` interface for shared properties/methods. 

I've left existing properties in the other interfaces alone for backwards compatibility:
image

pastaq (approved):

pastaq:
Seems weird to have duplicate entries in evdev and udev. I assume those will be deprecated at some point?

PR #213, opened 2024-10-08
Change some Ayaneo devices to xbox-elite

uejji:
InputPlumber maps LC and RC to LeftPaddle1 and RightPaddle1, respectively.  However, Ayaneo devices are set to xb360 and cannot take advantage.

This changes tested Ayaneo devices to xbox-elite.  Air Plus Mendo and Kun are untested and are not included.


uejji:
Air Plus Mendocino tested working as well.

pastaq:
I'll update Kun separately, thanks.

PR #216, opened 2024-10-16
fix(target): Adjust QAM chord timing

uejji:
Occasionally, pushing the QAM button may have SteamOS process the emitted Guide and A button outputs in the incorrect order (eg pressing A and then opening the Steam menu).  This seems to occur more often on devices that do not support holding the QAM output (such as ROG Ally and MSI Claw).

This changes the Guide button to be held an extra 80ms before emitting A.  In my testing I failed to see a single aberrant event once this change was applied.  Testing was performed on MSI Claw, Win600, ROG Ally and Ayaneo Air Plus.


ShadowApex (approved):
I have also sometimes seen passthrough, so I think this is a good change.

pastaq (approved):

PR #225, opened 2024-11-07
fix(CompositeDevice): allow multiple mappings to be applied for an event

imLinguin:
This only affects some devices that poll both axis at the same time. This triggered only left and right events completely ignoring y axis.


ShadowApex (approved):
Looks good! Thanks for the fix!

PR #227, opened 2024-11-19
feat(CompositeDevice): add new GAMEPAD_ONLY intercept mode and fix Steam Deck trackpads

ShadowApex:
This change introduces a new intercept mode called `GAMEPAD_ONLY` (enum `3`) mode. In this mode, only gamepad events will be intercepted; allowing mouse, keyboard, and touch inputs to be passed to their respective virtual target devices like normal. The primary motivation for this change is that when devices are in intercept `ALL` mode, and they have touchscreen/touchpad/mouse/keyboard source devices, all of those events will be sent to the DBus target device. This results in those events going unhandled because we do not currently have DBus event equivalent translations for all touchscreen/touchpad/mouse/keyboard events.

We should be able to not intercept these events in this mode because mouse/keyboard events will be routed based on Gamescope's input focus. The alternative to this would be for us to create DBus event equivalents for all keyboard/mouse/touchpad/touchscreen events.

In addition to this new mode, this change also includes fixes for the Steam Deck driver trackpads. It will now disable the built-in mouse emulation for the trackpads and correctly normalize the touchpad values.


pastaq:
How does this interact with the capability map? I.e. if a keyboard device has two events translate into a guide button, will they be passed through or will the translation happen first?

ShadowApex:
> How does this interact with the capability map? I.e. if a keyboard device has two events translate into a guide button, will they be passed through or will the translation happen first?

Intercept happens at the last stage of event handling in the `write_event()` method, which occurs after capability map and input profile translations.

Ericky14 (approved):

PR #231, opened 2024-11-20
fix(ds5): dpad getting "stuck"

### LINKED ISSUE #230: DualSense 5 source device dpad input gets "stuck" in certain cases
@imLinguin reported that when using the DualSense 5 controller, dpad input can get stuck when performing the following input:

- Press dpad right
- Press dpad down
- Release dpad right

In this case, the "dpad right" will continue to be held. This appears to be due to how we're handling dpad input in the DS5 driver implementation:
https://github.com/ShadowBlip/InputPlumber/blob/main/src/drivers/dualsense/driver.rs#L248-L354

imLinguin:
I hate this solution. 
If anyone has better idea on how to solve this, please please bring it up.  
Closes #230

This essentially moves logic for releasing buttons after the logic that maps button pressed events.
The logic especially handles transitions between in-between states like `NorthEast -> East` - this will release a North button.

### DIFF
--- a/src/drivers/dualsense/driver.rs
+++ b/src/drivers/dualsense/driver.rs
@@ -246,125 +246,50 @@ impl Driver {
             })));
         }
         if state.dpad != old_state.dpad {
-            match state.dpad {
-                Direction::North => events.push(Event::Button(ButtonEvent::DPadUp(BinaryInput {
+            let new_dpad = state.dpad.as_bitflag();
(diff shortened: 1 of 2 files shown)


imLinguin:
Managed to find a somewhat better approach, by first mapping enum back to integer, while splitting fields like NorthEast to separate bits. This allows us to get rid of this huge match statement for finding differences between states

ShadowApex (approved):

PR #233, opened 2024-11-22
fix(TargetDevice): add method for target devices to query composite device config

ShadowApex:
This change adds new methods to allow target devices to query the composite device it's attached to for its configuration and source devices. This can allow us to modify the behavior of target devices based on the composite device config and/or source devices.


pastaq (commented):
I guess this is approved...

PR #239, opened 2024-11-27
feat(Lua): add event scripting via Lua

### LINKED ISSUE #238: [Feature] Add Lua scripting support
Libraries like [mlua](https://github.com/mlua-rs/mlua) could allow us to embed Lua to enable users to do advanced event processing through scripting.

ShadowApex:
This change adds Lua scripting support for custom input routing. Users can now write Lua scripts to add custom input routing and translation logic that can be executed on every input event in the input pipeline.

Whenever a new composite device is initialized, each composite device instantiates its own Lua runtime instance and will load any lua scripts from the following directories:

* `/etc/inputplumber/scripts.d`
* `/usr/share/inputplumber/scripts`

Each script should return a lua table containing function(s) that will get executed by the composite device at certain stages of the input processing pipeline. This can allow users to easily implement their own custom input routing logic.

It can also be used as a pretty nice debugging tool. For example, this script will print whenever certain input events are emitted:

```lua
-- preprocess_event is called on every input event -before- capability mapping
-- and input profile translation.
local preprocess_event = function(event)
  if event.capability == "Gamepad:Button:Guide" then
    print("Guide button event:", event.value)
  end
  
  return true
end

return {
  preprocess_event = preprocess_event,
}
```

Fixes  #238

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -272,7 +272,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "regex",
- "rustc-hash",
+ "rustc-hash 1.1.0",
  "shlex",
  "syn 2.0.66",
 ]
@@ -288,6 +288,9 @@ name = "bitflags"
 version = "2.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf4b9d6a944f767f8e5e0db018570623c85f3d925ac718db4e06d0187adb21c1"
+dependencies = [
+ "serde",
+]
 
 [[package]]
 name = "bitvec"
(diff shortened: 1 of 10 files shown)


pastaq:
Neat

R1kaB3rN:
Question: Why did you decide to make the event scripting feature required and not optional instead?

As an enhancement, if the feature is intended to be included by default, I think it would be a plus to allow it be disabled when compiling Input Plumber.

ShadowApex:
> Question: Why did you decide to make the event scripting feature required and not optional instead?
> 
> As an enhancement, if the feature is intended to be included by default, I think it would be a plus to allow it be disabled when compiling Input Plumber.

We can make this feature optional via a compilation flag. It does not need to be required.

pastaq:
Aside from a rebase, what is this waiting on?

ShadowApex:
> Aside from a rebase, what is this waiting on?

This change needs better error handling to include. I'm still not completely sure that this feature will be useful enough to maintain. I think this was an interesting experiment, but I'm not sure if the extra complexity is worth it.

PR #246, opened 2024-12-31
Implement LED source output device

NeroReflex:
Implement LED source output device contextually a code cleanup:
  - Remove duplicated code
  - Avoid spawning useless tasks
  - Keep references to some spwned tasks and join then in the main() function
  - Avoid calling a CLI software to fetch devices data and use the (already in use) udev crate
  -  Add VSCode launch.json file to be able to remote debug the software


src/udev/mod.rs
pastaq, line 1: This is an interface that predates our use of udev-rs and is being depreciated over time. I don't know if there is value in updating it vice removing it and using the newer udev methods.

pastaq (changes requested):
I left a few comments but I'll add some more general notes here. I haven't given everything in the PR a completely thorough look due to some of the reasons posted below.

1.) I'm not a fan of sweeping refactoring that also includes new features as it is difficult to both review and test. I would prefer one PR to add LED RGB and one PR to refactor manager/devnode usage. Order isn't super important but I think it would be easier to refactor then add new features after.

2.) Please explain some of the reasons behind many of the changes. With so many very large changes it is difficult to grasp the context without having discussions beforehand or adding lots of reasoning in the pull request itself.

3.) Please ensure all commit titles are using the semantics format so our CI will pick everything up properly.

NeroReflex:
> I left a few comments but I'll add some more general notes here. I haven't given everything in the PR a completely thorough look due to some of the reasons posted below.
> 
> 1.) I'm not a fan of sweeping refactoring that also includes new features as it is difficult to both review and test. I would prefer one PR to add LED RGB and one PR to refactor manager/devnode usage. Order isn't super important but I think it would be easier to refactor then add new features after.
> 
> 2.) Please explain some of the reasons behind many of the changes. With so many very large changes it is difficult to grasp the context without having discussions beforehand or adding lots of reasoning in the pull request itself.
> 
> 3.) Please ensure all commit titles are using the semantics format so our CI will pick everything up properly.

1) the PR is divided in two stages: early commits are refactors that are needed to avoid losing referemces to spawned threads, then there is the commit that adds leds code and remove useless specialization of source devices creation: as a result of that removal leds are also being taken into consideration. Previous code was very long spanning multiple monitora is length and adding another specialization was both unpractical and basically more tech debt.

2) well see 1, I can write that to commit messages if you like that more

3) what? Where do I read about that?

src/input/manager.rs
```diff
@@ -194,15 +197,7 @@ impl Manager {
         }
     }
```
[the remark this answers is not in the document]
NeroReflex: See general answer. Basically reduce reliance of devnode() as leds don't have one, while also removing duplicated code with very minimal specialization

src/input/manager.rs
```diff
@@ -399,6 +402,56 @@ impl Manager {
...
+        }
+
+        log::debug!("Starting input manager task...");
+
+        let _ = tokio::join!(
+            // Discover all (initial) devices
+            Self::discover_all_devices(&cmd_tx_all_devices),
+            // Watch for IIO device events
+            Self::watch_iio_devices(cmd_tx_iio),
+            // Watch for iio and other devices
```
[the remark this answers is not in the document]
NeroReflex: When I refactored this method I kept the logic intact. This question is one I cannot answer.

src/input/manager.rs
```diff
@@ -399,6 +402,56 @@ impl Manager {
         Ok(())
     }
 
+    /// Starts listening for [Command] messages to be sent from clients and
+    /// dispatch those events.
+    pub async fn run(&mut self) -> Result<(), Box<dyn Error + Send + Sync>> {
+        let tx_for_listen_on_dbus = self.tx.clone();
+        let dbus_for_listen_on_dbus = self.dbus.clone();
+
+        let this: &Self = self;
```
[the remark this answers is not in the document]
NeroReflex: Self is mutable. Is used to make the borrow checker happy...

src/input/source/led/multicolor_chassis.rs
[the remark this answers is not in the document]
NeroReflex: The way it's implemented is easy to extend to more devices. It scan multi_index and creates a map of component. If a component is recognised (currently only rgb is) when a color arrives it reads the multi_intensity file and changes only relevant and recognised values. There is no way to specialize this when just multi_index contains different count or rgb or rgba.

src/config/mod.rs
```diff
@@ -362,6 +363,24 @@ pub struct IIO {
...
+    pub led_fixed_color: Option<LedFixedColor>,
+}
+
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+#[serde(rename_all = "snake_case")]
+#[allow(clippy::upper_case_acronyms)]
+pub struct LedFixedColor {
+    pub r: u8,
+    pub g: u8,
+    pub b: u8,
```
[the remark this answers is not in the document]
NeroReflex: Yes, I can do that. An Option<u8> for alpha

src/input/manager.rs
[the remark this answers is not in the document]
NeroReflex: I re-watched the code: discover_all_devices does the first initialization, while watch_iio_devices is meant to keep living on.

Before the refactor discover_all_devices was spawned in a task and that tash's reference was lost.

src/input/manager.rs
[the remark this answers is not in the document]
NeroReflex: well.... I solved this.

src/dbus/interface/source/led.rs
```diff
@@ -0,0 +1,44 @@
+use crate::input::source::iio::get_dbus_path;
+use crate::udev::device::UdevDevice;
+use std::error::Error;
+use zbus::{fdo, Connection};
+use zbus_macros::interface;
+/// DBusInterface exposing information about a HIDRaw device
```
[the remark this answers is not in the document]
NeroReflex: well I copied it from iio, so it is wrong in IIO too. lol.

src/config/mod.rs
```diff
@@ -620,17 +647,39 @@ impl CompositeDeviceConfig {
...
+            log::trace!("Checking name: {name} against {dname}");
+            if !glob_match(name.as_str(), dname.as_str()) {
+                return false;
+            }
+        }
+
+        true
+    }
+
+    /// Returns true if a given iio device is within a list of iio configs.
```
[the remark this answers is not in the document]
NeroReflex: fixed

src/dbus/interface/source/led.rs
```diff
@@ -0,0 +1,44 @@
...
+pub struct SourceLedInterface {
+    device: UdevDevice,
+}
+impl SourceLedInterface {
+    pub fn new(device: UdevDevice) -> SourceLedInterface {
+        SourceLedInterface { device }
+    }
+    /// Creates a new instance of the source hidraw interface on DBus. Returns
```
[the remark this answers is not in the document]
NeroReflex: same as before: iio is also wrong.

src/udev/mod.rs
NeroReflex: > This is an interface that predates our use of udev-rs and is being depreciated over time. I don't know if there is value in updating it vice removing it and using the newer udev methods.

I think there is value in that: if it is kept as small as possible it will reflect the bare minimum needed for the software to run and allows for future ways of discovering devices.

src/drivers/opineo/driver.rs
```diff
@@ -51,12 +52,24 @@ pub struct Driver {
 impl Driver {
     pub fn new(udevice: UdevDevice) -> Result<Self, Box<dyn Error + Send + Sync>> {
         let path = udevice.devnode();
-        let cs_path = CString::new(path.clone())?;
+        if !Path::new(path.as_str()).exists() {
+            return Err(format!(
+                "Device '{}' does not have a devnode: a valid OrangePi NEO Controller has one",
```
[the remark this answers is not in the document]
NeroReflex: I want the message to include the fact it's coming from an opi device... Can you give me a phrase that includes that?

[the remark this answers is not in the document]
pastaq: "Failed to find device node for {}, aborting initialization as OrangePi NEO Controller."

src/input/manager.rs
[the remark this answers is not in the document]
pastaq: The code highlight was unfortunate and led you to check something I didn't intend. The comment  for line 422 and the comment for line 444 `Self::watch_hidraw_devices(cmd_tx_hidraw, &mut watcher_rx),` are strange. will watch_hidraw_devices watch for iio? I know that both are in /dev so it's possible?

src/input/composite_device/mod.rs
```diff
@@ -308,66 +312,66 @@ impl CompositeDevice {
                 match cmd {
                     CompositeCommand::ProcessEvent(device_id, event) => {
                         if let Err(e) = self.process_event(device_id, event).await {
-                            log::error!("Failed to process event: {:?}", e);
+                            log::error!("Failed to process event: {}", e.to_string());
```
flukejones, line 327: All of this is unnecessary and adds noise to the commit series. These types of things should be done as a completely separate "cleanup" series, probably using clippy also.

src/udev/device.rs
```diff
@@ -653,10 +653,21 @@ pub struct Device {
 }
 
 impl Device {
+    /// Returns a path to the sysfs device link, such as
+    /// "/sys/devices/pci0000:00/0000:00:08.1/0000:09:00.3/usb1/1-3/1-3:1.2/0003:0B05:1ABE.0003/hidraw/hidraw2/device"
+    fn sysnode_device_link(&self) -> String {
+        let s = self.path.as_str();
+        match s.ends_with("/device") {
```
flukejones, line 663: Is this actually the correct thing to do? In changes below you change from this:
```rust
        let path = format!("/sys{}/device/id/vendor", self.path.clone());
```
to this:
```rust
        let path = format!("{}/id/vendor", self.sysnode_device_link());
```
The path now has `device` appended to it since it didn't end with it. Only `get_parent_device_name()` expected this.

I would suggest adding some unit tests. Or drop this change completely since the commit apparently addresses only the `u64 change and not this.

In fact please drop all the changes that aren't `u64`. You've changed the behaviour unnecessarily

src/udev/mod.rs
flukejones: You really want to not use `Command::new("udevadm")` in these things unless there is no other way. I would suggest full removal

src/udev/mod.rs
```diff
@@ -129,89 +130,54 @@ async fn reload_all() -> Result<(), Box<dyn Error>> {
...
-        if line.starts_with("V: ") {
-            let line = line.replace("V: ", "");
-            device.driver = line;
-            continue;
+    let output = (match path.starts_with("/dev/") {
+        true => {
+            let metadata = fs::metadata(path).map_err(|e| Box::new(e))?;
+            let devtype = match metadata.st_mode() & nix::libc::S_IFMT {
+                nix::libc::S_IFCHR => Some(udev::DeviceType::Character),
+                nix::libc::S_IFBLK => Some(udev::DeviceType::Block),
```
flukejones, line 139: It's likely to be preferable to use the `rustix` crate instead of nix in most cases now.

src/config/mod.rs
```diff
@@ -650,41 +699,39 @@ impl CompositeDeviceConfig {
             device
         );
 
-        let evdev_config = evdev_config.clone();
-
-        if let Some(name) = evdev_config.name {
+        if let Some(name) = evdev_config.name.as_ref() {
```
flukejones, line 703: can also be done as
```rust
        if let Some(name) = &evdev_config.name {
```
which is more concise. Keep in mind that this change isn't required at all, since `String` is clone and not copy, the string is moved to here and ownership taken, then the `.as_str()` is called below to get a str ref.

Personally I would drop these changes from the commit to reduce noise.

src/dbus/interface/source/led.rs
```diff
@@ -0,0 +1,44 @@
...
+    pub async fn listen_on_dbus(
+        conn: Connection,
+        device: UdevDevice,
+    ) -> Result<(), Box<dyn Error>> {
+        let iface = SourceLedInterface::new(device);
+        let Ok(id) = iface.id() else {
+            return Ok(());
+        };
+        let path = get_dbus_path(id);
+        tokio::task::spawn(async move {
```
flukejones, line 26: This is done in an async function, is it *really* required just to start the object server? The `zbus` crate starts async tasks itself, see https://github.com/dbus2/zbus/blob/71834a41a008214acda27b31dc5948c8ca08a594/zbus/src/connection/mod.rs#L950

src/input/composite_device/mod.rs
```diff
@@ -257,22 +267,28 @@ impl CompositeDevice {
...
         // Keep track of all target devices
         for (path, target) in targets.iter() {
             if let Err(e) = target.set_composite_device(self.client()).await {
-                return Err(
-                    format!("Failed to set composite device for target device: {:?}", e).into(),
-                );
+                return Err(format!(
+                    "Failed to set composite device for target device: {}",
+                    e.to_string()
+                )
```
ShadowApex, line 281: If the error already implements the `Display` trait, then you can just include the variable in the format string to do the equivalent:
```suggestion
                return Err(format!(
                    "Failed to set composite device for target device: {e}"
                )
```

src/input/composite_device/mod.rs
```diff
@@ -1404,38 +1404,45 @@ impl CompositeDevice {
             }
         }
 
-        let subsystem = device.subsystem();
-
-        let source_device = match subsystem.as_str() {
+        let dev_name = (&device).name();
+        let dev_subsystem = (&device).subsystem();
```
ShadowApex, line 1408: this expression borrows a value the compiler would automatically borrow
```suggestion
        let dev_name = device.name();
        let dev_subsystem = device.subsystem();
```

src/input/composite_device/mod.rs
```diff
@@ -1404,38 +1404,45 @@ impl CompositeDevice {
...
                 // Get any defined config for the IIO device
-                let config = if let Some(device_config) = self.config.get_matching_device(&device) {
-                    device_config.iio
-                } else {
-                    None
-                };
+                let config = self
+                    .config
+                    .get_matching_device(&device)
+                    .map_or(None, |cfg| cfg.iio);
```
ShadowApex, line 1428:
```suggestion
                let config = self
                    .config
                    .get_matching_device(&device)
                    .and_then(|cfg| cfg.iio);
```

src/input/composite_device/mod.rs
```diff
@@ -1404,38 +1404,45 @@ impl CompositeDevice {
...
+
+                log::debug!("Adding {dev_subsystem} source device: {dev_name}");
+                SourceDevice::Iio(IioDevice::new(device, self.client(), config)?)
+            }
+            "leds" => {
+                // Get any defined config for the IIO device
+                let config = self
+                    .config
+                    .get_matching_device(&device)
+                    .map_or(None, |cfg| cfg.led);
```
ShadowApex, line 1438:
```suggestion
                let config = self
                    .config
                    .get_matching_device(&device)
                    .and_then(|cfg| cfg.led);
```

src/input/manager.rs
```diff
@@ -692,22 +745,33 @@ impl Manager {
...
+                log::trace!(
+                    "Skipping composite device {composite_device}: is a single source device"
+                );
                 continue;
             }
             if config.maximum_sources.unwrap_or(0) == 1 {
-                log::trace!("{:?} is a single source device. Skipping.", config.name);
+                log::trace!(
+                    "Skipping composite device {composite_device}: is a single source device"
+                );
```
ShadowApex, line 774: With the removal of the second loop, how will new composite devices be created in these circumstances? The expected behavior should be:
- If a device matches an existing composite device, and it hasn't reached its maximum number of source devices, it should be added to the existing composite device.
- If a device matches an existing composite device, but it HAS reached its maximum, a new composite device should be created.
- Lastly, if no existing composite device exists, a new one should be created.

With these changes, I don't see how a new composite device could be ever be created if an existing one reaches its max.

ShadowApex (changes requested):

src/input/manager.rs
ShadowApex: For example, you can follow these steps to see this problem manifested:

- Enable management of all supported input devices: `busctl set-property org.shadowblip.InputPlumber /org/shadowblip/InputPlumber/Manager org.shadowblip.InputManager ManageAllDevices b 1`
- Plug in an XBox 360 controller (or other supported external controller). A Composite Device should be created for that controller.
- Plug in a second controller of the same type. A second Composite Device should be created for that second controller.

src/udev/device.rs
pastaq: Agree, only the bugfix is needed here.

src/input/manager.rs
```diff
@@ -1294,13 +1294,16 @@ impl Manager {
         let dev_path = device.devpath();
         let sys_name = device.sysname();
         if sys_name.is_empty() {
+            log::warn!(
```
pastaq, line 960: This is going to be very noisy. Use trace, or at most debug.

src/config/mod.rs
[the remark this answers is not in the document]
pastaq: I think some provide a duty cycle to change the overall brightness.

pastaq (changes requested):
I would drop these commits:

cd3dbdd9bc9b822ae6b2d7c6cfe9255655797618 
53edb302463419bd3d7fad7bac2d4428d13c66f6
bbee9b3ca79cb54e28be046300e26c7cbd3de69d

I would cherry pick these commits into a separate cleanup PR and drop them from this one:
005be0023880ae3e04cd33d4c9f95419780bbe6c
7135d069010cb501b4904cd108cdd33c7fb1727d
27f33ecc438faec0961396a4a9319ff35872ca96
54db7ae2c70af10e1fe4bdeead40c611075107a8
45d1cdf5a9e21548bc452926e5840aff8f18644b
45eb7117f0e2f8a9bc09ed960daebe1b92baaa6b
c65effe9d7e15e65e683b9d6c9d6adb34fc2c271
0b79670db7007fd9e99d2b53236a05792d25f0a9

src/dbus/interface/source/led.rs
NeroReflex: > This is done in an async function, is it _really_ required just to start the object server? The `zbus` crate starts async tasks itself, see https://github.com/dbus2/zbus/blob/71834a41a008214acda27b31dc5948c8ca08a594/zbus/src/connection/mod.rs#L950

I copied this code from iio_imu and I mantained like that for consistency. The main problem is that right now listen_on_dbus() the function that contains the spawn is called by manager.rs in a function that must not terminate an the spawned function has a ```rust
conn.object_server().at(path.clone(), iface).await
```

so either more manager.rs logic is changed or that will have to stay like that, unless the await immediatly returns (I assume it won't, but I can try)...

File: src/udev/mod.rs
Comment (NeroReflex) [reply src/udev/mod.rs:1]: > You really want to not use `Command::new("udevadm")` in these things unless there is no other way. I would suggest full removal

See your answer here: https://github.com/ShadowBlip/InputPlumber/pull/246#discussion_r1900584039

Other changes are to cope well with the removal of calling udevadm so you three have to decide if that class has to be:
- fixed and I have to find some other way to implement leds (there might be one that will be more complex, will that be fine?): might be major work for me
- left bugged (it's rare, who cares lol) and call udevadm and not touch that code at all: remove noise from the PR
- pick all changes and futher rework that struct to make it relevant again: maybe needless work as UdevDevice is used way more anyway
- drop that udev::Device structure completely and only use UdevDevice: probably better, but still adds noise so I will do in a separate PR

File: .vscode/launch.json [hunk 20]
```diff
@@ -0,0 +1,79 @@
...
+                "filter": {
+                    "name": "inputplumber",
+                    "kind": "bin"
+                }
+            },
+            "args": [],
+            "cwd": "${workspaceFolder}"
+        }
+    ]
+}
```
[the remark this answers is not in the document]
NeroReflex: > We shouldn't commit editor-specific configurations

This is not editor configuration as in "dev favourite color palette" this is what is needed to remote debug it. The PATH env var is important too as it specify where to find udevadm.

src/input/manager.rs
```diff
@@ -1626,125 +1380,102 @@ impl Manager {
...
-                                "/dev" => {
-                                    if !name.starts_with("hidraw") {
-                                        None
-                                    } else {
-                                        Some("hidraw")
-                                    }
+    async fn watch_hidraw_devices(
+        cmd_tx: mpsc::Sender<ManagerCommand>,
+        watcher_rx: &mut mpsc::Receiver<WatchEvent>,
+    ) -> () {
```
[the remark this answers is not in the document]
NeroReflex: > 

pastaq said otherwise in another comment....

src/input/manager.rs
```diff
@@ -1760,6 +1491,9 @@ impl Manager {
         let iio_devices = udev::discover_devices("iio")?;
         let iio_devices = iio_devices.into_iter().map(|dev| dev.into()).collect();
         Manager::discover_devices(cmd_tx, iio_devices).await?;
+        let led_devices = udev::discover_devices("leds")?;
+        let led_devices = led_devices.into_iter().map(|dev| dev.into()).collect();
+        Manager::discover_devices(&cmd_tx, led_devices).await?;
```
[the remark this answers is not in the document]
NeroReflex: A leftover from a previous cleanup. I will solve this in the new PR

NeroReflex:
Superceeded by #268

PR #247, opened 2025-01-02
Debug

NeroReflex:
Add a debug file for VSCode IDE to remote debug the software


ShadowApex (changes requested):
I wouldn't mind documentation on how to set up remote debugging, but I would prefer to not have editor-specific configuration in the repository.

pastaq:
I agree with @ShadowApex

PR #248, opened 2025-01-02
Fix(udev::Device): number is defined as an u64 in udev

NeroReflex:
The device number is defined as an u64 in udev: fix a possible bug by using the correct data type


ShadowApex (approved):

NeroReflex:
Do not merge this until you have decided what to do about this struct

pastaq:
I think once we've merged all the current pending PR's, including the RGB stuff, I'll work on fully deprecating using this interface for enumeration of devices. UdevDevice seems to be able to do everything for that. This module will just be for device hiding.

PR #249, opened 2025-01-03
Refactor code making it easier to add new device types

NeroReflex:
Since LEDs devices do not have a devnode the hard requirement for source devices of having one is to be removed and the presence check must be postponed to type of devices that actually have one and requires it to work properly.

Refactor manager.rs to remove device specialization where no particular benefits are provided by that specialization.

This PR reduces the overall number of running threads, without changing how the application behaves.


src/input/manager.rs
```diff
@@ -738,280 +782,77 @@ impl Manager {
...
-                }
-                _ => (),
-            };
-            log::trace!(
-                "Device {id} does not match existing device: {:?}",
-                config.name
-            );
+            }
         }
+
```
ShadowApex, line 855: Reduce nesting:

```rust
            let Some(source_device) = config.get_matching_device(&device) else {
                log::trace!(
                    "Device {id} does not match existing device: {:?}",
                    config.name
                );
                continue;
            };

            // Check if the device has already been used in this config or not,
            // stop here if the device must be unique.
            if let Some(sources) = self.composite_device_sources.get(composite_device) {
                for source in sources {
                    if *source != source_device {
                        continue;
                    }
                    if let Some(ignored) = source_device.ignore {
                        if ignored {
                            log::debug!(
                                        "Ignoring device {:?}, not adding to composite device: {composite_device}",
                                        source_device
                                    );
                            break 'start;
                        }
                    }
                    if let Some(unique) = source_device.clone().unique {
                        if unique {
                            log::trace!(
                                        "Found unique device {:?}, not adding to composite device {composite_device}",
                                        source_device
                                    );
                            break 'start;
                        }
                    // Default to being unique
                    } else {
                        log::trace!(
                                    "Found unique device {:?}, not adding to composite device {composite_device}",
                                    source_device
                                );
                        break 'start;
                    }
                }
            }

            log::info!("Found missing {} device, adding source device {id} to existing composite device: {composite_device:?}", device.subsystem());
            let Some(client) = self.composite_devices.get(composite_device.as_str()) else {
                log::error!("No existing composite device found for key {composite_device:?}");
                continue;
            };
            self.add_device_to_composite_device(device, client).await?;
            self.source_devices_used
                .insert(id.clone(), composite_device.clone());
            let composite_id = composite_device.clone();
            if !self.composite_device_sources.contains_key(&composite_id) {
                self.composite_device_sources
                    .insert(composite_id.clone(), Vec::new());
            }
            let sources = self
                .composite_device_sources
                .get_mut(&composite_id)
                .unwrap();
            sources.push(source_device.clone());
            self.source_devices.insert(id, source_device.clone());

            return Ok(());
        }

```

src/input/manager.rs
```diff
@@ -1043,155 +884,40 @@ impl Manager {
...
+                    .await?;
+
+                    return Ok(());
                 }
-                _ => (),
+                None => {}
             }
+
             log::trace!("Device does not match config: {:?}", config.name);
         }
```
ShadowApex, line 922: Reduce nesting

```rust
            // Check if this device matches any source configs
            let Some(source_device) = config.get_matching_device(&device) else {
                log::trace!("Device does not match config: {:?}", config.name);
                continue;
            };
            if let Some(ignored) = source_device.ignore {
                if ignored {
                    log::trace!("Event device configured to ignore: {:?}", device);
                    return Ok(());
                }
            }
            log::info!(
                "Found a matching {} device {id}, creating CompositeDevice",
                device.subsystem()
            );
            let dev = self
                .create_composite_device_from_config(&config, device)
                .await?;

            // Get the target input devices from the config
            let target_devices_config = config.target_devices.clone();

            // Create the composite deivce
            self.start_composite_device(
                dev,
                config.clone(),
                target_devices_config,
                source_device.clone(),
            )
            .await?;

            return Ok(());
        }

```

ShadowApex (changes requested):

src/input/manager.rs
NeroReflex: This code has to behave like the previous one, and the previous one is like that, furthermore I don't understand what you are asking me: nesting or cyclomatic complexity? Nesting of what in particular?

ShadowApex: We should try to reduce the level of nested branches where possible to improve readability.

For example:

```rust
match foo {
    Some(thing) => {
        log::info!("Got some {thing}!");
    },
    None => {
        log::error!("Got none!");
    },
}
```

could be re-written as:

```rust
let Some(thing) = foo else {
    log::error!("Got none!");
    return;
};
log::info!("Got some {thing}!");
```

The code in my comment was an example of how you could write that section so it behaves in the same way, but with less levels of nested branches.

src/input/manager.rs
```diff
@@ -1205,7 +917,8 @@ impl Manager {
         device: UdevDevice,
         id: String,
     ) -> Result<(), Box<dyn Error>> {
-        log::debug!("Source device removed: {}", device.devnode());
+        let dev_name = device.name();
+        log::debug!("Source device removed: {dev_name}");
```
pastaq, line 921: If we aren't reusing dev_name then we don't need to break it out.

src/input/manager.rs
```diff
@@ -1478,9 +1202,10 @@ impl Manager {
     }
 
     async fn on_device_removed(&mut self, device: UdevDevice) -> Result<(), Box<dyn Error>> {
-        log::debug!("Device removed: {:?}", device.devnode());
+        let dev_name = device.name();
```
pastaq, line 1205: Same here.

src/input/manager.rs
```diff
@@ -1573,18 +1298,9 @@ impl Manager {
...
-        let cmd_tx = self.tx.clone();
-        task::spawn(async move {
-            if let Err(e) = Manager::discover_all_devices(&cmd_tx).await {
-                log::error!("Failed to perform initial device discovery: {e:?}");
-            }
-        });
-
-        // Watch for IIO device events.
-        let cmd_tx = self.tx.clone();
+    fn watch_iio_devices(
```
pastaq, line 1301: Please add a /// comment

src/input/manager.rs
```diff
@@ -1601,150 +1317,129 @@ impl Manager {
...
-                    WatchEvent::Create { name, base_path } => {
-                        let subsystem = {
-                            match base_path.as_str() {
-                                "/dev" => {
-                                    if !name.starts_with("hidraw") {
-                                        None
-                                    } else {
-                                        Some("hidraw")
-                                    }
+    async fn watch_hidraw_devices(
```
pastaq, line 1350: Please add /// comment.

src/input/manager.rs
```diff
@@ -1601,150 +1317,129 @@ impl Manager {
...
-                        // Wait until the device has initialized with udev
-                        const MAX_TRIES: u8 = 80;
-                        let mut attempt: u8 = 0;
-                        loop {
-                            // Break after max attempts reached
-                            if attempt > MAX_TRIES {
-                                log::warn!("Unable to create initialized UdevDevice for {base_path}/{name} after {MAX_TRIES} attempts.");
-                                continue 'outer;
                             }
+                            "/dev/input" => Some("input"),
```
pastaq, line 1366: This function watches both hidraw(/dev/hidraw#) and evdev (/dev/input/event#), so the function name is misleading. Maybe call it watch_devnode_devices, watch_inofity_devices, something similar?
@ShadowApex opinion?

src/main.rs
```diff
@@ -57,19 +45,52 @@ async fn main() -> Result<(), Box<dyn Error>> {
     // Create an InputManager instance
     let mut input_manager = Manager::new(connection.clone());
 
-    // Start the input manager and listen on DBus
-    tokio::spawn(async move {
-        log::debug!("Starting input manager thread");
-        if let Err(e) = input_manager.run().await {
-            log::error!("Error running input manager: {:?}", e);
+    let (ctrl_c_result, input_man_result, request_name_result) = tokio::join!(
```
pastaq, line 48: I haven't ever experienced an issue with ctrl+c. When does this happen?

pastaq (changes requested):
Some minor nits and a couple questions. Overall I like the changes, good work. Especially the simplified functions, that has been on my TODO list for a while.

src/main.rs
NeroReflex: The reference to the spawned thread is always lost in the original code as the tokio::spawn result was not even captured. I don't really like losing track of spawned tasks inside functions otherwise the need to debug with tokio console will intensify at the point of being a must to understand how things works together (even in the sense of sequential vs parallel)

src/input/manager.rs
```diff
@@ -296,13 +289,16 @@ impl Manager {
                     self.target_devices.remove(&path);
                 }
                 ManagerCommand::DeviceAdded { device } => {
+                    let dev_name = device.name();
+                    let dev_sysname = device.sysname();
+
                     if let Err(e) = self.on_device_added(device).await {
-                        log::error!("Error adding device: {e:?}");
+                        log::error!("Error adding device '{dev_name} ({dev_sysname})': {e}");
```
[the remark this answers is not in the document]
NeroReflex: > Why are we breaking these out if they are only used in the error? If you want a more "readable" error then at least capture them inside the Err path so we aren't using extra memory. Even then, I think `log::error!("Error adding device '{} ({})': {e}", device.name(). device.sysname());` would be better

cargo format will break that in many lines adding up noise for no real benefit.

The reason those are not inside is that self.on_device_added wants a Device, and not a &Device so I would need to clone device and then clone the rest too as .name() containst a call to .clone()

src/input/manager.rs
NeroReflex: > If we aren't reusing dev_name then we don't need to break it out.

Same as before: cargo fmt will transform those two lines in 4 lines where one will be );

src/input/manager.rs
NeroReflex: > Same here.

yes, same answer too

src/input/manager.rs
```diff
@@ -194,15 +195,7 @@ impl Manager {
...
-    /// Starts listening for [Command] messages to be sent from clients and
-    /// dispatch those events.
-    pub async fn run(&mut self) -> Result<(), Box<dyn Error>> {
-        // Start tasks for discovering new input devices
-        self.watch_input_devices().await?;
-
-        // Create a DBus interface
-        self.listen_on_dbus().await?;
-
+    async fn events_loop(&mut self) -> Result<(), Box<dyn Error>> {
```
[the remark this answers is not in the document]
NeroReflex: That is going to mess the commit even more, you will see large number of linea removed, same number of file added with no way of checking if they have been re-added as they were. I would advice against in this case.

src/input/manager.rs
NeroReflex: Done

src/input/manager.rs
NeroReflex: Done

src/input/manager.rs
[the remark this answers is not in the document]
pastaq: I think that is fine, the scope of this PR is much smaller and easier to follow.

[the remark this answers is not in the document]
NeroReflex: > I think that is fine, the scope of this PR is much smaller and easier to follow.

So? What do I do?

src/input/manager.rs
NeroReflex: I have named it "watch_devnodes" since it watches /dev and associate new devnodes with the corresponding udev device.

src/input/manager.rs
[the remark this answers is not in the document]
ShadowApex: I think what pastaq means is why did you not write it like this:

```rust
                ManagerCommand::DeviceAdded { device } => {
                    if let Err(e) = self.on_device_added(device).await {
                        let dev_name = device.name();
                        let dev_sysname = device.sysname();
                        log::error!("Error adding device '{dev_name} ({dev_sysname})': {e}");
                    }
                }
```

instead of how you wrote it:

```rust
                ManagerCommand::DeviceAdded { device } => {
                    let dev_name = device.name();
                    let dev_sysname = device.sysname();

                    if let Err(e) = self.on_device_added(device).await {
                        log::error!("Error adding device '{dev_name} ({dev_sysname})': {e}");
                    }
                }
```

@pastaq , the reason he cannot do this is that `device` is moved into the `self.on_device_added` function, so he cannot reference `device` again. He would need to refactor all downstream functions to use `&UdevDevice` instead of `UdevDevice` in order to accomplish what you're asking. Since this path is only used when devices are plugged in, I don't think this will be a problem to leave it how it's currently implemented.

src/input/manager.rs
```diff
@@ -1601,150 +1314,130 @@ impl Manager {
...
-                                    if !name.starts_with("hidraw") {
-                                        None
-                                    } else {
-                                        Some("hidraw")
-                                    }
+    /// Watch for appearance and disappearence of devices is /dev and associate the corresponding udev device
+    async fn watch_devnodes(
+        cmd_tx: mpsc::Sender<ManagerCommand>,
+        watcher_rx: &mut mpsc::Receiver<WatchEvent>,
+    ) -> () {
```
ShadowApex, line 1351:
```suggestion
    ) {
```

NeroReflex: > -> ()

Done

src/input/manager.rs
NeroReflex: Well... Done.

src/input/manager.rs
NeroReflex: Done (yestarday)

src/input/manager.rs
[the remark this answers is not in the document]
pastaq: > I think that is fine, the scope of this PR is much smaller and easier to follow.

This got added to the wrong comment. I blame mobile firefox/

src/input/manager.rs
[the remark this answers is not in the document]
pastaq: I think that is fine, the scope of this PR is much smaller and easier to follow.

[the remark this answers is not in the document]
NeroReflex: I moved run immediately after new() so I guess it's done.

src/input/manager.rs
```diff
@@ -196,13 +197,49 @@ impl Manager {
...
+                watcher::watch(INPUT_PATH.into(), tx)
+            });
+        }
+
+        log::debug!("Starting input manager task...");
+
+        let _ = tokio::join!(
+            Self::discover_all_devices(&cmd_tx_all_devices),
+            Self::watch_iio_devices(cmd_tx_iio),
+            Self::watch_devnodes(cmd_tx_hidraw, &mut watcher_rx),
```
pastaq, line 233: Should probably change the name of the cloned tx here since the func name changed.

src/input/manager.rs
```diff
@@ -196,13 +197,49 @@ impl Manager {
 
     /// Starts listening for [Command] messages to be sent from clients and
     /// dispatch those events.
-    pub async fn run(&mut self) -> Result<(), Box<dyn Error>> {
-        // Start tasks for discovering new input devices
-        self.watch_input_devices().await?;
+    pub async fn run(&mut self) -> Result<(), Box<dyn Error + Send + Sync>> {
+        let tx_for_listen_on_dbus = self.tx.clone();
```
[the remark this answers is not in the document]
NeroReflex: > Why are you breaking out all these variables when you could just pass self.tx.clone() to the function?

Self::discover_all_devices accepts a reference and the compiler is not willing to accept a &self.tx.clone(). Others were done that way to keep consistency with the first. I changed them just now (but kept cmd_tx_all_devices).

src/input/manager.rs
NeroReflex: Done by replacing with self.tx.clone()

pastaq (approved):

src/input/manager.rs
```diff
@@ -1601,150 +1311,130 @@ impl Manager {
...
+                        {
+                            let Ok(device) = ::udev::Device::from_subsystem_sysname(
+                                subsystem.to_string(),
+                                name.clone(),
+                            ) else {
+                                log::debug!(
+                                    "Unable to create UdevDevice from {base_path}/{name} to check initialization"
+                                );
+                                attempt += 1;
+                                tokio::time::sleep(Duration::from_millis(10)).await;
```
pastaq, line 1391: I know this was 10ms before, but my Elite 2 controller doesn't initialize fully before the 800ms have passed. It seems to connect fine if this is 50ms.

pastaq (changes requested):
When ManageAllDevices is set to true, Manager fails to create a CompositeDevice for an external controller. We get a false positive for a virtual controller when it should detect as a bluetooth device. 

Tested on Xbox Elite Series 2 Controller.

[2025-01-09T00:46:30Z DEBUG inputplumber::input::manager] Xbox Wireless Controller (event21) is virtual, skipping consideration for /devices/virtual/misc/uhid/0005:045E:0B22.002C/input/input125/event21

src/input/manager.rs
```diff
@@ -1296,13 +1008,13 @@ impl Manager {
...
                     // Look up the connected device using udev
-                    let device_info = udev::get_device(dev_node.clone()).await?;
+                    let device_info = udev::get_device(dev_path.clone()).await?;
 
                     // Check if the virtual device is using the bluetooth bus
                     // TODO: Can we get properties from UdevDevice::get_attribute_from_tree?
                     let id_bus = device_info.properties.get("ID_BUS");
 
-                    log::debug!("Bus ID for {dev_node}: {id_bus:?}");
+                    log::debug!("Bus ID for {dev_path}: {id_bus:?}");
```
pastaq, line 1017: devnode is required for virtual device checks, devpath will not match the bus ID .

NeroReflex: This is just a debug, am I missing something here?

pastaq: Look at  the two lines above. The device_info grabbed for devnode is not the same as devpath. devpath wont return the same udev device instance, so the info will be different.

NeroReflex: Can this conversation be marked as "resolved"? I have looked more into it but I can't see how the code might behave differently other than printing different debugs...

ShadowApex:
Testing this change with the Sony DualSense controller, I get two `CompositeDevice` objects:

```
└─ /org
  └─ /org/shadowblip
    └─ /org/shadowblip/InputPlumber
      ├─ /org/shadowblip/InputPlumber/CompositeDevice0
      ├─ /org/shadowblip/InputPlumber/CompositeDevice1
```

One device captures the hidraw interface:

```
busctl get-property org.shadowblip.InputPlumber /org/shadowblip/InputPlumber/CompositeDevice0 org.shadowblip.Input.CompositeDevice SourceDevicePaths
as 1 "/dev/hidraw8"
```

And the other captures the evdev nodes:

```
busctl get-property org.shadowblip.InputPlumber /org/shadowblip/InputPlumber/CompositeDevice1 org.shadowblip.Input.CompositeDevice SourceDevicePaths
as 3 "/dev/input/event24" "/dev/input/event25" "/dev/input/event26"
```

NeroReflex:
> Testing this change with the Sony DualSense controller, I get two `CompositeDevice` objects:
> 
> ```
> └─ /org
>   └─ /org/shadowblip
>     └─ /org/shadowblip/InputPlumber
>       ├─ /org/shadowblip/InputPlumber/CompositeDevice0
>       ├─ /org/shadowblip/InputPlumber/CompositeDevice1
> ```
> 
> One device captures the hidraw interface:
> 
> ```
> busctl get-property org.shadowblip.InputPlumber /org/shadowblip/InputPlumber/CompositeDevice0 org.shadowblip.Input.CompositeDevice SourceDevicePaths
> as 1 "/dev/hidraw8"
> ```
> 
> And the other captures the evdev nodes:
> 
> ```
> busctl get-property org.shadowblip.InputPlumber /org/shadowblip/InputPlumber/CompositeDevice1 org.shadowblip.Input.CompositeDevice SourceDevicePaths
> as 3 "/dev/input/event24" "/dev/input/event25" "/dev/input/event26"
> ```

I answer to this mainly for personal record:

After doing this to manage all devices:
```
busctl set-property org.shadowblip.InputPlumber /org/shadowblip/InputPlumber/Manager org.shadowblip.InputManager ManageAllDevices b 1
```

I only got two composite devices:

one for ds5 and one for the ally
```
[gamer@chimeraos ~]$ busctl get-property org.shadowblip.InputPlumber /org/shadowblip/InputPlumber/CompositeDevice0 org.shadowblip.Input.CompositeDevice SourceDevicePaths
as 6 "/dev/hidraw3" "/dev/input/event2" "/dev/input/event6" "/dev/input/event8" "/dev/iio:device0" "/dev/input/event14"
[gamer@chimeraos ~]$ busctl get-property org.shadowblip.InputPlumber /org/shadowblip/InputPlumber/CompositeDevice1 org.shadowblip.Input.CompositeDevice SourceDevicePaths
as 4 "/dev/hidraw1" "/dev/input/event5" "/dev/input/event7" "/dev/input/event9"
[gamer@chimeraos ~]$ busctl get-property org.shadowblip.InputPlumber /org/shadowblip/InputPlumber/CompositeDevice2 org.shadowblip.Input.CompositeDevice SourceDevicePaths
Failed to get property SourceDevicePaths on interface org.shadowblip.Input.CompositeDevice: Unknown object '/org/shadowblip/InputPlumber/CompositeDevice2'
```

ShadowApex:
@NeroReflex It seems like I can no longer replicate the duplicate CompositeDevice issue I was seeing before, so ignore my comment :sweat_smile:

src/input/manager.rs
```diff
@@ -1022,13 +847,11 @@ impl Manager {
...
-                if let Some(options) = config.options.as_ref() {
-                    options.auto_manage.unwrap_or(false)
-                } else {
-                    false
-                }
-            };
+            let auto_manage = config
+                .options
+                .as_ref()
+                .and_then(|options| Some(options.auto_manage.unwrap_or(false)))
```
ShadowApex, line 853: We should fix this clippy warning:

```suggestion
                .map(|options| options.auto_manage.unwrap_or(false))
```

NeroReflex: It's done. Cool suggestion.

ShadowApex (approved):

pastaq (approved):

PR #264, opened 2025-01-23
fix(SourceDevice): code cleanup

NeroReflex:
This commit does not modify how the software behaves, while refactoring some code:
  - remove a method and a big match that was used only once just to avoid calling .clone()
  - solve the SourceDevice scope creep that was in need (for no reason) to know specific internal details about sub-types


pastaq:
I'd prefer it if we can merge #265 prior to this. It will be easier to rebase this after a new device is added vice rebasing adding a new device.

NeroReflex:
> I'd prefer it if we can merge #265 prior to this. It will be easier to rebase this after a new device is added vice rebasing adding a new device.

Okay, when that other PR is merged you just tell me and I will rebase this PR.

pastaq (approved):

ShadowApex (approved):

PR #265, opened 2025-01-23
feat(Legion Go S): Add Legion Go S Support

pastaq:
- Adds Initial support for the Legion Go S

TODO: 
- The touchpad uses REL values instead of ABS values. Everything is plumbed for this to work, but commented out for now.
- The buffer for the go s driver gets filled with the same garbage data when the gyro/accel hidraw interface is grabbed. Since the XInputDataReport for the gamepad doesn't have a report_id, this data can't be filtered out normally. Using hid-recorder to track all hid reports from both interfaces shows that this data doesn't come from the hidraw interface directly. For now I have a workaround to look for the bad data packet (which is consistent and may be an output report?).


src/drivers/legos/driver.rs
```diff
@@ -0,0 +1,404 @@
...
+
+    /// Translate the state into individual events
+    fn translate_xinput(&self, old_state: Option<XInputDataReport>) -> Vec<Event> {
+        let mut events = Vec::new();
+        let Some(state) = self.xinput_state else {
+            return events;
+        };
+
+        // Translate state changes into events if they have changed
+        if let Some(old_state) = old_state {
```
ShadowApex, line 152: Nit: we can reduce nesting here using this pattern instead of `if let Some(old_state)`:

```rust
let Some(old_state) = old_state else {
    return events;
};

// Binary Events
if state.a != old_state.a {
    events.push(Event::GamepadButton(GamepadButtonEvent::A(BinaryInput {
        pressed: state.a,
    })));
}
...
```

src/drivers/legos/driver.rs
```diff
@@ -0,0 +1,402 @@
...
+        //log::debug!("--- Input report ---");
+        //log::debug!("{input_report}");
+        //log::debug!(" ---- End Report ----");
+
+        let report_type = match input_report.report_id {
+            1 => ReportType::AccelData,
+            2 => ReportType::GyroData,
+            _ => {
+                let _report_id = input_report.report_id;
+                return Err("Unknown report type: {_report_id}".into());
```
ShadowApex, line 317:
```suggestion
                let report_id = input_report.report_id;
                return Err(format!("Unknown report type: {report_id}").into());
```

src/drivers/legos/driver.rs
```diff
@@ -0,0 +1,402 @@
...
+                return Ok(events);
+            }
+            ReportType::GyroData => {
+                // Update the state
+                let old_state = self.update_gyro_state(input_report);
+                // Translate the state into a stream of input events
+                let events = self.translate_gyro_data(old_state);
+                return Ok(events);
+            }
+        };
```
ShadowApex, line 336: To fix clippy warnings:

```suggestion
        match report_type {
            ReportType::AccelData => {
                // Update the state
                let old_state = self.update_accel_state(input_report);
                // Translate the state into a stream of input events
                let events = self.translate_accel_data(old_state);
                Ok(events)
            }
            ReportType::GyroData => {
                // Update the state
                let old_state = self.update_gyro_state(input_report);
                // Translate the state into a stream of input events
                let events = self.translate_gyro_data(old_state);
                Ok(events)
            }
        }
```

src/drivers/legos/hid_report.rs
```diff
@@ -0,0 +1,190 @@
...
+            && self.touch_x == data.touch_x
+            && self.touch_y == data.touch_y
+            && self.a_trigger_l == data.a_trigger_l
+            && self.a_trigger_r == data.a_trigger_r
+        {
+            true
+        } else {
+            false
+        }
+    }
```
ShadowApex, line 136: Because this struct implements `PartialEq`, we should be able to compare data directly:

```suggestion
    /// Determines if the current report matches the bad data report.
    pub fn is_bad_data(&self) -> bool {
        *self == bad_data
    }
```

src/drivers/legos/driver.rs
```diff
@@ -0,0 +1,402 @@
...
+        // garbage. Since it doesn't have a report_id we cant reject it. This only seems to hapen
+        // one time, so we can save a lot of checks in the future if we clear a bool once it
+        // happens.
+        if !self.bad_data_passed {
+            if input_report.is_bad_data() {
+                log::debug!("Got bad XInputDataReport, regecting it.");
+                self.bad_data_passed = true;
+                return Ok(vec![]);
+            }
+        }
```
ShadowApex, line 121: If the device doesn't support report ids, this might be the best thing we can do, unless the vendor can provide more information.

To fix the clippy warning, though:

```suggestion
        // Hacky workaround. When the gyro device is grabbed the XInputDataReport is full of
        // garbage. Since it doesn't have a report_id we cant reject it. This only seems to hapen
        // one time, so we can save a lot of checks in the future if we clear a bool once it
        // happens.
        if !self.bad_data_passed && input_report.is_bad_data() {
            log::debug!("Got bad XInputDataReport, regecting it.");
            self.bad_data_passed = true;
            return Ok(vec![]);
        }
```

ShadowApex (changes requested):
A few small things to fix and some nitpicks, but other than that, this looks good! :+1:

src/drivers/legos/driver.rs
pastaq: Clippy complains about this for some reason

ShadowApex: It complains because `format!()` isn't being used to include the `report_id` in the error message.

pastaq:
I'm not sure why my clippy isn't showing the same suggestions. Please review that I didn't miss anything since I'm getting no warnings on my end except the Touchpad events not used..

ShadowApex (approved):

PR #266, opened 2025-01-24
fix(dbus) make dbus path parth of a CompositeDevice at construction

NeroReflex:
This PR improves the handling of the dbus for CompositeDevice: makes it so a dbus is required to have a CompositeDevice and in no situation the service can crash/panic by removing the .unwrap(). It also simplify the code a bit decreasing lines count.

This is also part of the saga "do not spawn tasks without returning the handle and try not to lose tasks references" albeit in this case the reference is still lost, but it is lost in manager.rs and a future code improvement can pick it up.

### DIFF
--- a/src/input/composite_device/mod.rs
+++ b/src/input/composite_device/mod.rs
@@ -8,7 +8,11 @@ use std::{
 };
 
 use evdev::InputEvent;
-use tokio::{sync::mpsc, task::JoinSet, time::Duration};
+use tokio::{
+    sync::mpsc,
+    task::{JoinHandle, JoinSet},
+    time::Duration,
+};
 use zbus::Connection;
 
 use crate::{
@@ -94,7 +98,7 @@ pub struct CompositeDevice {
     /// release events
     emitted_mappings: HashMap<String, CapabilityMapping>,
     /// The DBus path this [CompositeDevice] is listening on
-    dbus_path: Option<String>,
+    dbus_path: String,
     /// Mode defining how inputs should be routed
     intercept_mode: InterceptMode,
     /// Transmit channel for sending commands to this composite device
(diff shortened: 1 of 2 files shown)


pastaq (commented):
LGTM

src/input/composite_device/mod.rs
```diff
@@ -231,22 +236,25 @@ impl CompositeDevice {
         Ok(device)
     }
 
+    /// Return the DBus path of the composite device
+    pub fn dbus_path(&self) -> String {
+        self.dbus_path.clone()
+    }
```
ShadowApex, line 242: Maybe we should return the `&str` and let the caller decide if they want to `clone()`.

```suggestion
    pub fn dbus_path(&self) -> &str {
        self.dbus_path.as_str()
    }
```

src/input/manager.rs
```diff
@@ -1275,20 +1276,15 @@ impl Manager {
...
     /// Returns the next available composite device dbus path
-    fn next_composite_dbus_path(&self) -> String {
-        let max = 2048;
-        let mut i = 0;
-        loop {
-            if i > max {
-                return "Devices exceeded".to_string();
-            }
+    fn next_composite_dbus_path(&self) -> Result<String, Box<dyn Error>> {
+        for i in 0u64.. {
```
ShadowApex, line 1280: That's a lot of devices! :P

ShadowApex (approved):
Minor suggestion, but otherwise looks good! Thanks for this!

src/input/manager.rs
NeroReflex: > That's a lot of devices! :P

Indeed. This was done because udev has the device_num that is an u64, so I wanted to match that. It's more logical than a real problem to be fair, but it won't hurt.

src/input/composite_device/mod.rs
NeroReflex: Done

src/input/manager.rs
NeroReflex: The old UdevDevice (I think it's called like this) wrongly uses u32 still and it parses that number from the output of a CLI command.

In general I think that old class has to be completely removed in favor of the new one that more than half software already uses.

This modification will make things easier when such modification is done (or the correct type is used and that struct is kept)

pastaq (approved):

PR #268, opened 2025-01-26
Implement LEDs devices

NeroReflex:
This is an initial implementation of LEDs devices: as of now only a static color (defined in the configuration) can be set when the device is acquired.

### DIFF
--- a/rootfs/usr/share/inputplumber/devices/50-rog_ally.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-rog_ally.yaml
@@ -53,6 +53,9 @@ source_devices:
         x: [1, 0, 0]
         y: [0, -1, 0]
         z: [0, 0, -1]
+  - group: led
+    led:
+      name: ASUS ROG Ally Config
 
 # Optional configuration for the composite device
 options:
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -307,6 +307,7 @@ pub struct SourceDevice {
     pub evdev: Option<Evdev>,
     pub hidraw: Option<Hidraw>,
     pub iio: Option<IIO>,
+    pub led: Option<Led>,
     pub udev: Option<Udev>,
     pub unique: Option<bool>,
     pub blocked: Option<bool>,
@@ -360,6 +361,24 @@ pub struct IIO {
     pub id: Option<String>,
     pub name: Option<String>,
     pub mount_matrix: Option<MountMatrix>,
+}
+
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+#[serde(rename_all = "snake_case")]
+#[allow(clippy::upper_case_acronyms)]
+pub struct Led {
+    pub id: Option<String>,
+    pub name: Option<String>,
+    pub led_fixed_color: Option<LedFixedColor>,
+}
+
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+#[serde(rename_all = "snake_case")]
+#[allow(clippy::upper_case_acronyms)]
+pub struct LedFixedColor {
+    pub r: u8,
+    pub g: u8,
+    pub b: u8,
 }
 
 #[derive(Debug, Deserialize, Clone, PartialEq)]
@@ -455,6 +474,15 @@ impl CompositeDeviceConfig {
                 for config in self.source_devices.iter() {
                     if let Some(iio_config) = config.iio.as_ref() {
                         if self.has_matching_iio(udevice, iio_config) {
+                            return Some(config.clone());
+                        }
+                    }
+                }
+            }
+            "leds" => {
+                for config in self.source_devices.iter() {
+                    if let Some(led_config) = config.led.as_ref() {
+                        if self.has_matching_led(udevice, led_config) {
                             return Some(config.clone());
                         }
                     }
@@ -578,7 +606,6 @@ impl CompositeDeviceConfig {
     /// Returns true if a given hidraw device is within a list of hidraw configs.
     pub fn has_matching_hidraw(&self, device: &UdevDevice, hidraw_config: &Hidraw) -> bool {
         log::trace!("Checking hidraw config '{:?}'", hidraw_config,);
-        let hidraw_config = hidraw_config.clone();
 
         // TODO: Switch either evdev of hidraw configs to use the same type. Legacy version had i16
         // for hidraw and string for evdev.
@@ -606,7 +633,7 @@ impl CompositeDeviceConfig {
             }
         }
 
-        if let Some(name) = hidraw_config.name {
+        if let Some(name) = hidraw_config.name.as_ref() {
             let dname = device.name();
             log::trace!("Checking name: {name} against {dname}");
             if !glob_match(name.as_str(), dname.as_str()) {
@@ -620,9 +647,8 @@ impl CompositeDeviceConfig {
     /// Returns true if a given iio device is within a list of iio configs.
     pub fn has_matching_iio(&self, device: &UdevDevice, iio_config: &IIO) -> bool {
         log::trace!("Checking iio config: {:?} against {:?}", iio_config, device);
-        let iio_config = iio_config.clone();
 
-        if let Some(id) = iio_config.id {
+        if let Some(id) = iio_config.id.as_ref() {
             let dsyspath = device.syspath();
             log::trace!("Checking id: {id} against {dsyspath}");
             if !glob_match(id.as_str(), dsyspath.as_str()) {
@@ -630,7 +656,30 @@ impl CompositeDeviceConfig {
             }
         }
 
-        if let Some(name) = iio_config.name {
+        if let Some(name) = iio_config.name.as_ref() {
+            let dname = device.name();
+            log::trace!("Checking name: {name} against {dname}");
+            if !glob_match(name.as_str(), dname.as_str()) {
+                return false;
+            }
+        }
+
+        true
+    }
+
+    /// Returns true if a given iio device is within a list of iio configs.
+    pub fn has_matching_led(&self, device: &UdevDevice, led_config: &Led) -> bool {
+        log::trace!("Checking led config: {:?} against {:?}", led_config, device);
+
+        if let Some(id) = led_config.id.as_ref() {
+            let dsyspath = device.syspath();
+            log::trace!("Checking id: {id} against {dsyspath}");
+            if !glob_match(id.as_str(), dsyspath.as_str()) {
+                return false;
+            }
+        }
+
+        if let Some(name) = led_config.name.as_ref() {
             let dname = device.name();
             log::trace!("Checking name: {name} against {dname}");
             if !glob_match(name.as_str(), dname.as_str()) {
@@ -650,9 +699,7 @@ impl CompositeDeviceConfig {
             device
         );
 
-        let evdev_config = evdev_config.clone();
-
-        if let Some(name) = evdev_config.name {
+        if let Some(name) = evdev_config.name.as_ref() {
             let dname = device.name();
             log::trace!("Checking name: {name} against {dname}");
             if !glob_match(name.as_str(), dname.as_str()) {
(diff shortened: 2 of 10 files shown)


pastaq:
Does this replace #268 ?

NeroReflex:
> Does this replace #246 ?

Mostly. That other PR still has an improvement (that has a problem) that will be a future PR after this PR knce I will have fixed it.

rootfs/usr/share/inputplumber/devices/50-rog_ally.yaml
```diff
@@ -53,6 +53,9 @@ source_devices:
         x: [1, 0, 0]
         y: [0, -1, 0]
         z: [0, 0, -1]
+  - group: led
+    led:
+      name: ASUS ROG Ally Config
```
ShadowApex, line 58: We should also add these to the `rootfs/usr/share/inputplumber/schema/composite_device_v1.json` schema file.

Also, is "ASUS ROG Ally Config" the actual name returned by the driver?

src/dbus/interface/source/led.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+#[interface(name = "org.shadowblip.Input.Source.LEDDevice")]
+impl SourceLedInterface {
+    /// Returns the human readable name of the device (e.g. XBox 360 Pad)
+    #[zbus(property)]
+    fn id(&self) -> fdo::Result<String> {
+        Ok(self.device.sysname())
+    }
+}
```
ShadowApex, line 47: We could do this in a follow-up PR, but we should probably make an interface to allow users to change LED colors using DBus.

src/input/composite_device/mod.rs
```diff
@@ -1423,6 +1425,17 @@ impl CompositeDevice {
                 let device = IioDevice::new(device, self.client(), config)?;
                 SourceDevice::Iio(device)
             }
+            "leds" => {
+                // Get any defined config for the IIO device
```
ShadowApex, line 1429:
```suggestion
                // Get any defined config for the LED device
```

src/input/manager.rs
```diff
@@ -1447,6 +1478,9 @@ impl Manager {
         let iio_devices = udev::discover_devices("iio")?;
         let iio_devices = iio_devices.into_iter().map(|dev| dev.into()).collect();
         Manager::discover_devices(cmd_tx, iio_devices).await?;
+        let led_devices = udev::discover_devices("leds")?;
+        let led_devices = led_devices.into_iter().map(|dev| dev.into()).collect();
+        Manager::discover_devices(&cmd_tx, led_devices).await?;
```
ShadowApex, line 1483:
```suggestion
        Manager::discover_devices(cmd_tx, led_devices).await?;
```

src/input/manager.rs
```diff
@@ -1188,6 +1190,35 @@ impl Manager {
                 log::debug!("Finished adding event device {id}");
             }
 
+            "leds" => {
+                log::debug!("LED device added: {} ({})", device.name(), device.sysname());
+
+                // Create a DBus interface for the event device
```
ShadowApex, line 1196:
```suggestion
                // Create a DBus interface for the LED device
```

src/input/manager.rs
```diff
@@ -1188,6 +1190,35 @@ impl Manager {
...
+                if device.is_virtual() {
+                    log::debug!("{} is virtual, skipping consideration.", dev_path);
+                    return Ok(());
+                } else {
+                    log::trace!("Real device: {}", dev_path);
+                }
+                // Signal that a source device was added
+                log::debug!("Spawing task to add source device: {id}");
+                self.on_source_device_added(id.clone(), device).await?;
+                log::debug!("Finished adding event device {id}");
```
ShadowApex, line 1219:
```suggestion
                log::debug!("Finished adding LED device {id}");
```

src/input/source/led.rs
```diff
@@ -0,0 +1,98 @@
+pub mod multicolor_chassis;
+use std::error::Error;
+//use glob_match::glob_match;
```
ShadowApex, line 3:
```suggestion
```

src/input/source/led/multicolor_chassis.rs
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        let multi_index = PathBuf::from(device_info.syspath().as_str()).join("multi_index");
+        let contents = std::fs::read_to_string(multi_index)
+            .map_err(|err| Box::new(MultiColorChassisError::MultiIndexError(err)))?;
+        let multi_index_strings = contents
+            .split_whitespace()
+            .into_iter()
+            .map(|str| String::from(str))
+            .collect::<Vec<String>>();
```
ShadowApex, line 57: To make clippy happy:

```suggestion
        let multi_index_strings = contents
            .split_whitespace()
            .map(String::from)
            .collect::<Vec<String>>();
```

src/input/source/led/multicolor_chassis.rs
```diff
@@ -0,0 +1,167 @@
...
+    #[error("Error updating multi_intensity: {0}")]
+    MultiIntensityUpdateError(std::io::Error),
+
+    #[error("Unsupported index type: {0}")]
+    UnsupportedIndexType(String),
+}
+
+enum IndexType {
+    RGB,
+}
```
ShadowApex, line 26: To make clippy happy:

```suggestion
enum IndexType {
    Rgb,
}
```

src/input/source/led/multicolor_chassis.rs
```diff
@@ -0,0 +1,167 @@
...
+
+/// MultiColorChassis source device implementation
+pub struct MultiColorChassis {
+    multi_intensity_path: PathBuf,
+    multi_index_map: Vec<IndexType>,
+    fixed_color: Option<LedFixedColor>,
```
ShadowApex, line 32: Do we ever use `fixed_color` outside initialization? Maybe we can remove this and just set the color in the constructor?

pastaq:
I would say close that one and make that PR independently. We didn't need to track it as open since we won't be merging it.

src/input/source/led/multicolor_chassis.rs
```diff
@@ -0,0 +1,167 @@
...
+                        .to_string()
+                }
+                _ => String::from(index_value),
+            })
+            .collect::<Vec<String>>())
+        .join(" ");
+        Ok(
+            std::fs::write(self.multi_intensity_path.as_path(), contents)
+                .map_err(|err| Box::new(MultiColorChassisError::MultiIntensityUpdateError(err)))?,
+        )
```
ShadowApex, line 103: To fix clippy warnings:

```suggestion
        let values = self
            .multi_index_map
            .iter()
            .zip(contents.split_whitespace())
            .map(|(index_type, _index_value)| match index_type {
                IndexType::Rgb => {
                    (((r as u32) << 16u32) | ((g as u32) << 8u32) | (b as u32)).to_string()
                }
            })
            .collect::<Vec<String>>();
        let contents = values.join(" ");

        std::fs::write(self.multi_intensity_path.as_path(), contents)
            .map_err(|err| Box::new(MultiColorChassisError::MultiIntensityUpdateError(err)))?;

        Ok(())
```

src/input/source/led/multicolor_chassis.rs
```diff
@@ -0,0 +1,167 @@
...
+
+impl SourceOutputDevice for MultiColorChassis {
+    fn write_event(
+        &mut self,
+        event: crate::input::output_event::OutputEvent,
+    ) -> Result<(), crate::input::source::OutputError> {
+        //log::trace!("Received output event: {event:?}");
+        let _ = event;
+        Ok(())
+    }
```
ShadowApex, line 136: We should probably handle output events so LEDs can be controlled. Or we can handle this in a follow-up PR.

```suggestion
    fn write_event(&mut self, event: OutputEvent) -> Result<(), crate::input::source::OutputError> {
        //log::trace!("Received output event: {event:?}");
        match event {
            OutputEvent::DualSense(report) => {
                if !report.allow_led_color {
                    return Ok(());
                }
                // Handle DualSense LED events...
            }
            _ => {
                return Ok(());
            }
        }

        Ok(())
    }
```

ShadowApex (changes requested):

src/config/mod.rs
```diff
@@ -620,17 +647,39 @@ impl CompositeDeviceConfig {
...
+            log::trace!("Checking name: {name} against {dname}");
+            if !glob_match(name.as_str(), dname.as_str()) {
+                return false;
+            }
+        }
+
+        true
+    }
+
+    /// Returns true if a given iio device is within a list of iio configs.
```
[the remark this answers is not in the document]
NeroReflex: Again? Something has gone wrong since I fixed this already

rootfs/usr/share/inputplumber/devices/50-rog_ally.yaml
NeroReflex: Yes, that's the name luke gave it

src/dbus/interface/source/led.rs
NeroReflex: Indeed. Since I could not figure out how to make colors change via dualsense emulation and dbus I just implemented the basic to have leds change colors once.

src/input/source/led/multicolor_chassis.rs
NeroReflex: Oh no my eyes! Noooo not this.

src/input/source/led/multicolor_chassis.rs
NeroReflex: If we want to set it back after a resume or keep track of it I left it just for good measure, but yeah not very usefult.

src/input/source/led/multicolor_chassis.rs
NeroReflex: I was unable to figure out why so I left it alone for this PR, but I will look into getting it in a follow-up commit in this PR

src/input/source/led/multicolor_chassis.rs
pastaq: We can also do an #allow here to make clippy ignore it since it's an acronym

src/config/mod.rs
```diff
@@ -362,6 +363,24 @@ pub struct IIO {
     pub mount_matrix: Option<MountMatrix>,
 }
 
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+#[serde(rename_all = "snake_case")]
+#[allow(clippy::upper_case_acronyms)]
+pub struct Led {
+    pub id: Option<String>,
+    pub name: Option<String>,
+    pub led_fixed_color: Option<LedFixedColor>,
```
pastaq, line 372: I don't really see a case where we would ever specify a device start up with a specific color as this is entirely user preference. A simple dbus interface should be sufficient to expose and test this feature.

src/dbus/interface/source/led.rs
pastaq: I would also add a dbus bool property to permit or block source devices from adjusting the RGB color in the follow on PR. That will ensure targets like the DS5 don't change the RGB if the user prefers they not be changed.

src/input/source/led.rs
```diff
@@ -0,0 +1,98 @@
...
+    udev::device::UdevDevice,
+};
+/// List of available drivers
+enum DriverType {
+    MultiColorChassis,
+}
+/// [LedDevice] represents an input device using the leds subsystem.
+#[derive(Debug)]
+pub enum LedDevice {
+    MultiColorChassis(SourceDriver<MultiColorChassis>),
```
pastaq, line 17: We might consider changing this class name fro MultiColorChassis to JoystickRing since that's what the new asus driver calls them and I'll be changing ayn/ayaneo-platform drivers to do the same.

src/input/source/led/multicolor_chassis.rs
pastaq: Are we not storing the last setting otherwise? We could call this current_color and report it in dbus.

PR #272, opened 2025-01-28
feat(Touchscreen): add support for generic touchscreen devices

ShadowApex:
This change adds support for generic touchscreen source devices. This change also adds the `config` section of a source device to allow defining arbitrary configuration of that captured source device. i.e. in the case of touchscreens, it can allow the source driver to translate touches based on the orientation of the touchscreen.

### DIFF
--- a/rootfs/usr/share/inputplumber/devices/50-steam_deck.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-steam_deck.yaml
@@ -34,16 +34,17 @@ source_devices:
       product_id: 0x1205
       interface_num: 2
   # Touchscreen
-  - group: mouse
-    hidraw:
-      vendor_id: 0x2808
-      product_id: 0x1015
-  - group: mouse
+  - group: touchscreen
     unique: false
-    blocked: true
-    evdev:
-      name: "FTS3528:00 2808:1015*"
-      handler: event*
+    udev:
+      properties:
+        - name: ID_INPUT_TOUCHSCREEN
+          value: "1"
+      sys_name: "event*"
+      subsystem: input
+    config:
+      touchscreen:
+        orientation: "right"
   # Keyboard
   - group: keyboard
     evdev:
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -308,9 +308,39 @@ pub struct SourceDevice {
     pub hidraw: Option<Hidraw>,
     pub iio: Option<IIO>,
     pub udev: Option<Udev>,
+    pub config: Option<SourceDeviceConfig>,
     pub unique: Option<bool>,
     pub blocked: Option<bool>,
     pub ignore: Option<bool>,
+}
+
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+#[serde(rename_all = "snake_case")]
+pub struct SourceDeviceConfig {
+    pub touchscreen: Option<TouchscreenConfig>,
+    pub imu: Option<ImuConfig>,
+}
+
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+#[serde(rename_all = "snake_case")]
+pub struct TouchscreenConfig {
+    /// Orientation of the touchscreen. Can be one of: ["normal", "left", "right", "upsidedown"]
+    pub orientation: Option<String>,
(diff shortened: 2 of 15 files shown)


pastaq:
I'll test this when I get back.

src/config/mod.rs
```diff
@@ -359,6 +384,7 @@ pub struct UdevAttribute {
 pub struct IIO {
     pub id: Option<String>,
     pub name: Option<String>,
+    // DEPRECATED: use config.imu.mount_matrix instead
```
pastaq, line 387: Can we add a clippy depreciation tag here?

rootfs/usr/share/inputplumber/devices/50-steam_deck.yaml
```diff
@@ -34,16 +34,16 @@ source_devices:
...
+  - group: touchscreen
     unique: false
-    blocked: true
     evdev:
       name: "FTS3528:00 2808:1015*"
       handler: event*
+    config:
+      touchscreen:
+        orientation: "right"
+        width: 1280
```
pastaq, line 45: I think it might be better to use an aspect ratio here and scale it based in the true resolution. We would need to detect that

src/input/source/evdev/touchscreen.rs
```diff
@@ -0,0 +1,383 @@
...
+        device_info: UdevDevice,
+        config: Option<TouchscreenConfig>,
+    ) -> Result<Self, Box<dyn Error + Send + Sync>> {
+        let path = device_info.devnode();
+        log::debug!("Opening device at: {}", path);
+        let mut device = Device::open(path.clone())?;
+        device.grab()?;
+
+        // Set the device to do non-blocking reads
+        // TODO: use epoll to wake up when data is available
```
pastaq, line 121: Maybe we should start implementing epoll with this device?

ShadowApex: Using epoll will require a larger refactor that we should consider in a separate PR.

rootfs/usr/share/inputplumber/devices/50-steam_deck.yaml
ShadowApex: I left `width` and `height` here to allow users to override the width/height of the virtual touchscreen that gets created, but I have updated this PR to now query the source touchscreen for its size if these are not defined.

pastaq:
I'm going to test this on my flip to see if it can fix the bad touchscreen input

src/input/source/evdev/touchscreen.rs
```diff
@@ -0,0 +1,383 @@
...
+            .map(|e| self.translate(e))
+            .filter(|events| !events.is_empty())
+            .flatten()
+            .collect();
+
+        Ok(native_events)
+    }
+
+    /// Returns the possible input events this device is capable of emitting
+    fn get_capabilities(&self) -> Result<Vec<Capability>, InputError> {
```
pastaq, line 295: This will return gamepad capabilities.

src/input/source/evdev/touchscreen.rs
```diff
@@ -0,0 +1,383 @@
...
+        device.grab()?;
+
+        // Set the device to do non-blocking reads
+        // TODO: use epoll to wake up when data is available
+        // https://github.com/emberian/evdev/blob/main/examples/evtest_nonblocking.rs
+        let raw_fd = device.as_raw_fd();
+        nix::fcntl::fcntl(raw_fd, FcntlArg::F_SETFL(OFlag::O_NONBLOCK))?;
+
+        // Query information about the device to get the absolute ranges
+        let mut axes_info = HashMap::new();
```
pastaq, line 127: Add the ability to override the mix/max here, will fix some broken touchscreens.

ShadowApex: I have added a new touchscreen config option called `override_source_size` which will force the source device to use the defined `width` and `height` as its maximums.

E.g.

```yaml
source_devices:
  - group: touchscreen
    udev:
      properties:
        - name: ID_INPUT_TOUCHSCREEN
          value: "1"
      sys_name: "event*"
      subsystem: input
    config:
      touchscreen:
        override_source_size: true
        width: 1920
        height: 1080
```

pastaq (approved):

PR #273, opened 2025-01-31
fix: avoid inputs cancelling each other

### LINKED ISSUE #258: [Bug] Not possible to map same buttons to both D-Pad and analog stick
Currently, mapping a set of keys (say the arrow keys) to both the D-pad and an analog stick produces a broken config: 

- analog controls work
- d-pad only works for up and left but not down and right

From what I understand, this is due to how InputPlumber processes axis mappings. However, having both the d-pad and the left stick mapped to the same directions is very common in platformers, fighting games or shmups so it would be worthwhile to find a workaround for that use case.

imLinguin:
Closes #258

This is a proposition of a solution that addresses said issue. 
By mapping target_capability -> source_capability we can reference "ownership" of an input to given source event within the CompositeDevice. I think this change should be considered potentially breaking.

At the moment this prevents input cancellation for buttons on Gamepad, Keyboard and Mouse capabilities

### DIFF
--- a/src/input/composite_device/mod.rs
+++ b/src/input/composite_device/mod.rs
@@ -3,7 +3,10 @@ pub mod command;
 
 use std::{
     borrow::Borrow,
-    collections::{BTreeSet, HashMap, HashSet},
+    collections::{
+        hash_map::{Entry, HashMap},
+        BTreeSet, HashSet,
+    },
     error::Error,
 };
 
@@ -153,6 +156,9 @@ pub struct CompositeDevice {
     /// List of currently active buttons and keys. Used to block "up" events for
     /// keys that have already been handled.
     active_inputs: Vec<Capability>,
+    /// Mapping of target and source capabilities preventing input release events
+    /// coming from capabilities that didn't start the event
+    exclusive_inputs: HashMap<Capability, Capability>,
 }
 
 impl CompositeDevice {
@@ -203,6 +209,7 @@ impl CompositeDevice {
             intercept_mode_target_cap: Capability::Gamepad(Gamepad::Button(GamepadButton::Guide)),
             intercept_active_inputs: Vec::new(),
             active_inputs: Vec::new(),
+            exclusive_inputs: HashMap::new(),
         };
 
         // Load the capability map if one was defined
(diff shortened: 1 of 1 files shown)


src/input/composite_device/mod.rs
```diff
@@ -831,9 +838,46 @@ impl CompositeDevice {
...
+                                    e.remove();
+                                }
+                                Some(event)
+                            } else {
+                                None
+                            }
+                        }
+                    }
+                })
+                .collect()
```
ShadowApex, line 880: I can foresee us filtering other kinds of events in the future, so I think it would be good to break this out into its own function:

```rust

    /// Filter out input-cancelling events that do not come from same
    /// capability as the initiator
    fn filter_event(&mut self, event: NativeEvent) -> Option<NativeEvent> {
        let Some(src_cap) = event.get_source_capability() else {
            return Some(event);
        };
        let tgt_cap = event.as_capability();

        // Handle only button presses
        if !matches!(
            tgt_cap,
            Capability::Gamepad(Gamepad::Button(_))
                | Capability::Keyboard(_)
                | Capability::Mouse(Mouse::Button(_))
        ) {
            return Some(event);
        }
        let pressed = event.pressed();
        match self.exclusive_inputs.entry(tgt_cap) {
            Entry::Vacant(e) => {
                if pressed {
                    e.insert(src_cap.clone());
                }
                Some(event)
            }
            Entry::Occupied(e) => {
                if e.get() == &src_cap {
                    if !pressed {
                        e.remove();
                    }
                    Some(event)
                } else {
                    None
                }
            }
        }
    }
```

Then that also simplifies the `handle_event` logic:

```rust
        ...
        
        // Translate the event using the device profile.
        let mut events = if self.device_profile.is_some() {
            self.translate_event(&event)
                .await?
                .into_iter()
                .filter_map(|event| self.filter_event(event))
                .collect()
        } else {
            vec![event]
        };
        
        ...
```

ShadowApex (changes requested):
I think this is a good change. Thanks for this! I just had one comment, but otherwise looks good!

ShadowApex (approved):
Looks good! Thank you!

src/input/composite_device/mod.rs
```diff
@@ -153,6 +156,9 @@ pub struct CompositeDevice {
     /// List of currently active buttons and keys. Used to block "up" events for
     /// keys that have already been handled.
     active_inputs: Vec<Capability>,
+    /// Mapping of target and source capabilities preventing input release events
+    /// coming from capabilities that didn't start the event
```
pastaq, line 160: Small nit, can you more clearly define in the comment what the key/value pattern is. As they are both Capability it might become confusing to understand if the target or source capability is the key or the value.

imLinguin: Sure thing. Let me know if this change covers it well enough

pastaq (approved):

PR #275, opened 2025-02-04
feat: expose generic target interface on every target interface

imLinguin:
Finally got the use for `TargetInterface`. This will allow overlays to iterate over target devices to get their type - this is especially useful in cases like controller reordering or anything that relies on target device types

### DIFF
--- a/src/input/target/mod.rs
+++ b/src/input/target/mod.rs
@@ -10,7 +10,7 @@ use horipad_steam::HoripadSteamDevice;
 use thiserror::Error;
 use tokio::sync::mpsc::{self, error::TryRecvError};
 
-use crate::dbus::interface::target::gamepad::TargetGamepadInterface;
+use crate::dbus::interface::target::{gamepad::TargetGamepadInterface, TargetInterface};
 
 use super::{
     capability::Capability,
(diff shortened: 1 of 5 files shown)


pastaq:
> Finally got the use for `TargetInterface`. This will allow overlays to iterate over target devices to get their type - this is especially useful in cases like controller reordering or anything that relies on target device types

Just a note about this. We will need to implement controller reordering directly into InputPlumber. Because some targets need to be stopped during suspend (i.e. deck) they will come back in a different order on resume. I would not do this client side entirely, but instead rely on a dbus interface provided by InputPlumber to do the ordering.

src/input/target/mod.rs
```diff
@@ -244,17 +244,31 @@ impl TryFrom<&str> for TargetDeviceTypeId {
...
+        &mut self,
+        dbus: Connection,
+        path: String,
+        client: TargetDeviceClient,
+        type_id: String,
+    ) {
         log::debug!("Starting dbus interface: {path}");
         log::trace!("Using device client: {client:?}");
         tokio::task::spawn(async move {
             let name = "Gamepad".to_string();
```
ShadowApex, line 257: If we pass the `TargetDeviceTypeId`, we could also use that to populate this `name` variable to something more useful. E.g.

```rust
let name = type_id.name().to_string();
```

ShadowApex (changes requested):

ShadowApex (approved):

PR #278, opened 2025-02-07
feat(SourceDevice): add source device event filtering and configurable device hiding

ShadowApex:
This change adds two new configurable features:

- Source device event filtering - exclude processing certain events from a source device
- Source device passthrough - InputPlumber won't hide or grab exclusive access to a source device

### DIFF
--- a/src/input/source/evdev.rs
+++ b/src/input/source/evdev.rs
@@ -1,10 +1,12 @@
 pub mod blocked;
 pub mod gamepad;
+pub mod keyboard;
 pub mod touchscreen;
 
 use std::{collections::HashMap, error::Error, time::Duration};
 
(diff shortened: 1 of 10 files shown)


src/input/source/evdev.rs
```diff
@@ -133,6 +156,9 @@ impl EventDevice {
         if properties.contains_key("ID_INPUT_JOYSTICK") {
             return DriverType::Gamepad;
         }
+        if properties.contains_key("ID_INPUT_KEYBOARD") {
```
pastaq, line 159: This has the potential to reek havok on devices already using keyboard input for extra buttons. What is the behavior if no config is specified for a legacy evdev atkbd?

pastaq (approved):

PR #284, opened 2025-02-11
feat(Manager): add gamepad reordering

ShadowApex:
This change adds the `GamepadOrder` property to the `Manager` interface to allow users to re-order gamepads.


pastaq (approved):
LGTM

PR #285, opened 2025-02-12
feat(CompositeDevice): Add persist option

blindedone1458:
* Add CompositeDevice config bool option 'persist' 
* If persist is true then continue running CompositeDevice when all SourceDevices have been removed
* Update clear_state for Gamepad targets to default all inputs to stop all input from continuing

This improvement is to handle cases where a SourceDevice is disconnected to maintain the existing TargetDevice

### DIFF
--- a/src/input/target/dualsense.rs
+++ b/src/input/target/dualsense.rs
@@ -979,6 +979,19 @@ impl TargetInputDevice for DualSenseDevice {
         let _ = self.device.destroy();
         Ok(())
     }
+
+    /// Clear any local state on the target device.
+    fn clear_state(&mut self) {
+        let caps = self.get_capabilities().unwrap_or_else(|_| {
+            log::error!("No target device capabilities found while clearing state.");
+            Vec::new()
+        });
+        for cap in caps {
+            let ev = NativeEvent::new(cap, InputValue::Bool(false));
+            self.queued_events
+                .push(ScheduledNativeEvent::new(ev, Duration::from_millis(0)));
+        }
+    }
 }
 
 impl TargetOutputDevice for DualSenseDevice {
--- a/src/input/target/horipad_steam.rs
+++ b/src/input/target/horipad_steam.rs
@@ -1,5 +1,5 @@
 //! Emulates a Horipad Steam Controller as a target input device.
-use std::{cmp::Ordering, error::Error, fmt::Debug, fs::File};
+use std::{cmp::Ordering, error::Error, fmt::Debug, fs::File, time::Duration};
 
 use packed_struct::prelude::*;
 use uhid_virt::{Bus, CreateParams, StreamError, UHIDDevice};
@@ -317,6 +317,19 @@ impl TargetInputDevice for HoripadSteamDevice {
     fn stop(&mut self) -> Result<(), InputError> {
         let _ = self.device.destroy();
         Ok(())
+    }
+
+    /// Clear any local state on the target device.
+    fn clear_state(&mut self) {
+        let caps = self.get_capabilities().unwrap_or_else(|_| {
+            log::error!("No target device capabilities found while clearing state.");
+            Vec::new()
+        });
+        for cap in caps {
+            let ev = NativeEvent::new(cap, InputValue::Bool(false));
+            self.queued_events
+                .push(ScheduledNativeEvent::new(ev, Duration::from_millis(0)));
+        }
     }
 }
 
--- a/src/input/target/steam_deck.rs
+++ b/src/input/target/steam_deck.rs
@@ -831,6 +831,19 @@ impl TargetInputDevice for SteamDeckDevice {
         log::debug!("Finished stopping");
         Ok(())
     }
+
+    /// Clear any local state on the target device.
+    fn clear_state(&mut self) {
+        let caps = self.get_capabilities().unwrap_or_else(|_| {
+            log::error!("No target device capabilities found while clearing state.");
+            Vec::new()
+        });
+        for cap in caps {
+            let ev = NativeEvent::new(cap, InputValue::Bool(false));
+            self.queued_events
+                .push(ScheduledNativeEvent::new(ev, Duration::from_millis(0)));
+        }
+    }
 }
 
 impl TargetOutputDevice for SteamDeckDevice {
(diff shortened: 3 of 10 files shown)


src/input/target/dualsense.rs
```diff
@@ -979,6 +979,19 @@ impl TargetInputDevice for DualSenseDevice {
...
+        let caps = self.get_capabilities().unwrap_or_else(|_| {
+            log::error!("No target device capabilities found while clearing state.");
+            Vec::new()
+        });
+        for cap in caps {
+            let ev = NativeEvent::new(cap, InputValue::Bool(false));
+            self.queued_events
+                .push(ScheduledNativeEvent::new(ev, Duration::from_millis(0)));
+        }
+    }
```
ShadowApex, line 994: Queuing `InputValue::Bool(false)` for all capabilities is a pretty neat way to handle this, but we'll need to make sure this works for certain inputs like analog sticks.

For any hidraw or usb virtual devices, we can simply set the state back to its default instead of queuing:

```suggestion
    /// Clear any local state on the target device.
    fn clear_state(&mut self) {
        self.state = match self.state {
            PackedInputDataReport::Usb(_) => PackedInputDataReport::Usb(Default::default()),
            PackedInputDataReport::Bluetooth(_) => {
                PackedInputDataReport::Bluetooth(Default::default())
            }
        };
    }
```

src/input/target/horipad_steam.rs
```diff
@@ -318,6 +318,19 @@ impl TargetInputDevice for HoripadSteamDevice {
...
+        let caps = self.get_capabilities().unwrap_or_else(|_| {
+            log::error!("No target device capabilities found while clearing state.");
+            Vec::new()
+        });
+        for cap in caps {
+            let ev = NativeEvent::new(cap, InputValue::Bool(false));
+            self.queued_events
+                .push(ScheduledNativeEvent::new(ev, Duration::from_millis(0)));
+        }
+    }
```
ShadowApex, line 333: Like the DualSense, any target devices that are hidraw or usb based can simply reset the gamepad state instead of queuing events:

```suggestion
    /// Clear any local state on the target device.
    fn clear_state(&mut self) {
        self.state = Default::default();
    }
```

src/input/target/steam_deck.rs
```diff
@@ -831,6 +831,19 @@ impl TargetInputDevice for SteamDeckDevice {
...
+        let caps = self.get_capabilities().unwrap_or_else(|_| {
+            log::error!("No target device capabilities found while clearing state.");
+            Vec::new()
+        });
+        for cap in caps {
+            let ev = NativeEvent::new(cap, InputValue::Bool(false));
+            self.queued_events
+                .push(ScheduledNativeEvent::new(ev, Duration::from_millis(0)));
+        }
+    }
```
ShadowApex, line 846:
```suggestion
    /// Clear any local state on the target device.
    fn clear_state(&mut self) {
        self.state = Default::default();
    }
```

src/input/target/xb360.rs
```diff
@@ -218,6 +219,19 @@ impl TargetInputDevice for XBox360Controller {
...
     }
+
+    /// Clear any local state on the target device.
+    fn clear_state(&mut self) {
+        let caps = self.get_capabilities().unwrap_or_else(|_| {
+            log::error!("No target device capabilities found while clearing state.");
+            Vec::new()
+        });
+        for cap in caps {
+            let ev = NativeEvent::new(cap, InputValue::Bool(false));
```
ShadowApex, line 230: This might not work as expected for analog sticks. When any evdev-based target devices translate this `NativeEvent` to an `EvdevEvent`, it will use this logic:
https://github.com/ShadowBlip/InputPlumber/blob/main/src/input/event/evdev.rs#L885-L958

Since the value is set to `InputValue::Bool(false)` and there is axis information available, this will get translated as the minimum value of the axis, which would be like pressing the stick fully in one direction. We may want to consider either adding some logic to handle gamepad axes here, or potentially adding a new `InputValue::Zero` type that could be translated to the appropriate "zero" value for each kind of input.

ShadowApex (changes requested):
Thanks for this change! I think this is a great option to include :)

There are just a few small things in this to address and I think we can include this.

blindedone1458:
Thanks for the feedback @ShadowApex ! I've made the updates for the hidraw and usb targets, I totally missed that their tracked state was being used in that way. I do agree that applying `InputValue::Bool(false)` for all capabilities is a bit inelegant, but it has been working as expected in my testing in that it properly re-centers the joysticks and the triggers are back to zero.

ShadowApex:
Oh, nice! I thought that might cause issues. Let me also do some testing to double check that this will re-center those joysticks. If everything works, we should be able to merge this in. Thanks!

pastaq:
I have a few questions on usability. It seems this is only implemented as a config option. That makes me think there are a few cases we need to consider. Don't consider any of these blocking necessarily, but I want to discuss and have a plan at a minimum before we merge.

- As of #284, suspend will stop all target devices and they will be restarted in player order on resume. What will happen to a composite device during suspend that has no sources or targets?
- If the expectation is to use this in a config, what is the expectation for the default configs in /usr? What are the consequences of we enabled this by default on the gamepad configs? I.e. if a user is playing a game and their controller dies, so they plug it in or swap it. Both cases will detect as a new device (different bus type at a minimum) and create a second controller. Should we have some logic to try and connect any of the same type to the original device? If it's a different typ e(ds4 to Xbox for example) and the original controller is persistent, the new controller will be player 2 and player 1 will be permanently unable to be used unless some client sends a StopComposteDevice over dbus. Is there a graceful way to handle clearing this if we get a new controller?
- IMO this would be most useful as selectively enableable from client software. I'd like to see a dbus & CLI option to turn it on/off on the composite device 
- Doing the above adds another consideration. What happens if a device has this set to on, has no sources, and it is then set to off?

blindedone1458:
Definitely some more things to think about @pastaq and probably not for me to necessarily answer as a new contributor. It did give me a few more ideas for edge cases to test for though like if you plug in the controller when was connected via wireless, which currently would create a second persistent CompositeDevice though I have a fix in mind for that. 
I do think having a CompositeDevice with no sources active shouldn't remain stuck as Player 1 (which with persist: true would be the case as you point out)  so I'll look into having the ordering take that into account as well.

ShadowApex (approved):
Tested and works great! Once the merge conflict has been resolved, we should be able to merge this in!

Thanks!

PR #287, opened 2025-02-13
fix(CompositeDevice): move device hiding logic back to device start instead of device add

ShadowApex:
This change also includes fixes for a deadlock that can occur with `SetTargetDevices` introduced by an earlier version.

### DIFF
--- a/src/udev/mod.rs
+++ b/src/udev/mod.rs
@@ -43,7 +43,8 @@ pub async fn hide_device(path: &str) -> Result<(), Box<dyn Error>> {
 {match_rule}, GOTO="inputplumber_valid"
 GOTO="inputplumber_end"
 LABEL="inputplumber_valid"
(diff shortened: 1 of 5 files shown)


src/udev/mod.rs
```diff
@@ -43,7 +43,7 @@ pub async fn hide_device(path: &str) -> Result<(), Box<dyn Error>> {
 {match_rule}, GOTO="inputplumber_valid"
 GOTO="inputplumber_end"
 LABEL="inputplumber_valid"
-ACTION=="add|change", KERNEL=="hidraw[0-9]*|js[0-9]*|event[0-9]*", SUBSYSTEM=="{subsystem}", MODE:="0000", GROUP:="root", RUN:="{chmod_cmd} 000 {path}", SYMLINK+="inputplumber/by-hidden/%k"
+KERNEL=="hidraw[0-9]*|js[0-9]*|event[0-9]*", SUBSYSTEM=="{subsystem}", MODE="000", GROUP="root", RUN:="{chmod_cmd} 000 {path}", SYMLINK+="inputplumber/by-hidden/%k"
```
pastaq, line 46: Why?

philmmanjaro:
Adding a `ExecStartPre=/bin/sleep 3` to the `inputplumber.service` file hides the internal gaming controller correctly.

PR #288, opened 2025-02-15
fix(CompositeDevice): use early and late rules to fix hiding devices at boot time

ShadowApex:


src/udev/mod.rs
```diff
@@ -48,11 +49,26 @@ KERNEL=="hidraw[0-9]*", SUBSYSTEM=="{subsystem}", MODE:="0000", GROUP:="root", R
 LABEL="inputplumber_end"
 "#
     );
-
-    // Write the udev rule
     fs::create_dir_all(RULES_PREFIX)?;
     let rule_path =
-        format!("{RULES_PREFIX}/{RULE_HIDE_DEVICE_PRIORITY}-inputplumber-hide-{name}.rules");
+        format!("{RULES_PREFIX}/{RULE_HIDE_DEVICE_EARLY_PRIORITY}-inputplumber-hide-{name}.rules");
```
pastaq, line 54: Can we add early/late to the filenames?

ShadowApex: Sure

PR #291, opened 2025-02-17
Temporary support Legion Go 2

XinoAssassin:
83N0 for Z2E (Strix Point)
83N1 for Z2 (Hawk Point)


pastaq (approved):
Why is it labeled temporary?

pastaq (changes requested):
Please rename the commit and force push with semantic versioning.

`fix(Legion Go 2): Add initial support for legion go 2`

XinoAssassin:
> Why is it labeled temporary?

It's still under development and some endpoint may diff from the first gen.

ShadowApex (approved):

PR #294, opened 2025-02-19
feat(Unified Controller): add unified controller driver

ShadowApex:
This change adds the Unified Controller as defined by the [Unified Controller Input Spec](https://github.com/ShadowBlip/UnifiedControllerInputSpec). If used as a target device, it will create a virtual hidraw interface that will send input reports based on the source device capabilities.

### DIFF
--- a/src/drivers/unified_gamepad/capability.rs
+++ b/src/drivers/unified_gamepad/capability.rs
@@ -1,0 +1,707 @@
+use packed_struct::prelude::*;
+
+#[derive(PrimitiveEnum_u16, Clone, Copy, PartialEq, Debug, Default)]
+pub enum InputCapability {
+    #[default]
+    None = 0,
+    KeyboardKeyEsc = 1,
+    KeyboardKey1 = 2,
+    KeyboardKey2 = 3,
+    KeyboardKey3 = 4,
+    KeyboardKey4 = 5,
+    KeyboardKey5 = 6,
+    KeyboardKey6 = 7,
+    KeyboardKey7 = 8,
+    KeyboardKey8 = 9,
+    KeyboardKey9 = 10,
+    KeyboardKey0 = 11,
+    KeyboardKeyMinus = 12,
+    KeyboardKeyEqual = 13,
+    KeyboardKeyBackspace = 14,
+    KeyboardKeyTab = 15,
+    KeyboardKeyQ = 16,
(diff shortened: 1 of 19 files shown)


src/drivers/unified_gamepad/driver.rs
```diff
@@ -0,0 +1,154 @@
...
+                return self.handle_input_report(buf);
+            }
+            ReportType::OutputCapabilityReport => (),
+            ReportType::OutputDataReport => (),
+        }
+
+        Ok(vec![])
+    }
+
+    fn handle_input_report(
```
pastaq, line 113: I like this a lot. I think we can use a similar pattern in other drivers to simplify them.

src/drivers/unified_gamepad/capability.rs
```diff
@@ -0,0 +1,707 @@
...
+    KeyboardKeyKbdLcdMenu4 = 0x2bb,
+    KeyboardKeyKbdLcdMenu5 = 0x2bc,
+
+    MouseButtonLeft = 0x110,
+    MouseButtonRight = 0x111,
+    MouseButtonMiddle = 0x112,
+    MouseButtonSide = 0x113,
+    MouseButtonExtra = 0x114,
+
+    /// Base button, usually on the bottom right, Steam Quick Access Button (...)
```
pastaq, line 603: We should avoid calling this Base Button, those buttons are reserved for the physical base of a joystick. Only the openSD driver uses BTN_BASE for this, which is wrong.

src/drivers/unified_gamepad/capability.rs
```diff
@@ -0,0 +1,707 @@
...
+    GamepadButtonRightPaddle2 = 0x30b,
+    /// Right "side" paddle button, Legion Go M2
+    GamepadButtonRightPaddle3 = 0x30c,
+
+    /// Touch binary sensor for left stick
+    GamepadButtonLeftStickTouch = 0x30d,
+    /// Touch binary sensor for right stick
+    GamepadButtonRightStickTouch = 0x30e,
+
+    /// Dedicated button to open an on-screen keyboard
```
pastaq, line 660: the following 3 events are out of numerical order

pastaq:
Took a quick look, nothing major stands out. I'll run some tests and let you know how it goes. This is great.

pastaq (approved):

PR #296, opened 2025-02-21
fix(evdev) Add in missing mapping for evdev keycodes to capabilities

blindedone1458:
`event_codes_from_capability` already has capability -> keycodes, just adding the missing keycodes -> capability

Additionally fix the capability map schema to match the Config code and the existing capability maps.

### DIFF
--- a/rootfs/usr/share/inputplumber/schema/capability_map_v1.json
+++ b/rootfs/usr/share/inputplumber/schema/capability_map_v1.json
@@ -47,10 +47,10 @@
         "name": {
           "type": "string"
         },
-        "source_event": {
+        "target_event": {
           "$ref": "#/definitions/Event"
         },
-        "target_events": {
+        "source_events": {
           "type": "array",
           "items": {
             "$ref": "#/definitions/Event"
@@ -59,8 +59,8 @@
       },
       "required": [
         "name",
-        "source_event",
-        "target_events"
+        "source_events",
+        "target_event"
       ]
     },
     "Event": {
--- a/src/input/event/evdev.rs
+++ b/src/input/event/evdev.rs
@@ -160,6 +160,12 @@ impl EvdevEvent {
                 KeyCode::BTN_START => Capability::Gamepad(Gamepad::Button(GamepadButton::Start)),
                 KeyCode::BTN_SELECT => Capability::Gamepad(Gamepad::Button(GamepadButton::Select)),
                 KeyCode::BTN_MODE => Capability::Gamepad(Gamepad::Button(GamepadButton::Guide)),
+                KeyCode::BTN_TRIGGER_HAPPY5 => Capability::Gamepad(Gamepad::Button(GamepadButton::RightPaddle1)),
+                KeyCode::BTN_TRIGGER_HAPPY6 => Capability::Gamepad(Gamepad::Button(GamepadButton::RightPaddle2)),
+                KeyCode::BTN_TRIGGER_HAPPY7 => Capability::Gamepad(Gamepad::Button(GamepadButton::LeftPaddle1)),
+                KeyCode::BTN_TRIGGER_HAPPY8 => Capability::Gamepad(Gamepad::Button(GamepadButton::LeftPaddle2)),
+                KeyCode::BTN_TRIGGER_HAPPY9 => Capability::Gamepad(Gamepad::Button(GamepadButton::RightPaddle3)),
+                KeyCode::BTN_TRIGGER_HAPPY10 => Capability::Gamepad(Gamepad::Button(GamepadButton::LeftPaddle3)),
                 KeyCode::BTN_THUMBL => {
                     Capability::Gamepad(Gamepad::Button(GamepadButton::LeftStick))
                 }


pastaq:
The problem with defining BUTTON_TRIGGER_HAPPY_X as a specific capability is that they are used differently by different drivers and manufacturers. Without an established consistency the correct way to handle these keys is to create a capability map for the device that uses them.

blindedone1458:
I agree that defining a specific capability is fraught as there is no standard, but also from how I understand it without defining a capability here these buttons are just ignored as they return `Capability::NotImplemented` and thus cannot be used in the capability map for mapping them. 
This of course could be overcome if the capability map has the option of defining a mapping of raw Evdev keycodes to capabilities. 
So with my change here at least it is maintaining consistency of already defined translations.

pastaq:
Ah, I that we we're doing the reverse already. That's unfortunate... No fault of yours of course, but we should really be translating to new Capabilities rather than shoehorning in a specific implementation. We're also missing 1-4, which are used sometimes as a dpad, sometimes as paddles, and 11-40 which I haven't really seen used yet.

ShadowApex:
I think this is a good change for the time being. As mentioned, we do have to be careful about mapping `BTN_TRIGGER_HAPPY_X` to a specific capability, as they've been used pretty arbitrarily in different drivers. I would be ok with merging this in.

In general, I strongly believe that a `Capability` in InputPlumber should be actual real capabilities of an input device. So in other words, I don't don't think we should ever have a `Capability::TriggerHappyX`. We should probably come to a consensus and codify this somewhere, but I think that we should try to follow these rules when adding new capabilities to InputPlumber:

1. The capability should describe a real function of the device (i.e. no `Button1`, `TriggerHappy1`, etc.)
2. The capability should describe either:
  a. The input type and its location (e.g. `Axis::LeftStick`, `Axis::RightStick`, `Touchpad::Center`, etc.), or
  b. The input type and its function (e.g. `Button::Screenshot`, `Button::Guide`, etc.)

I think that following these rules will make sure that there's always a clear answer of what an input should be mapped to. Once we start adding abstract capabilities like `Axis4`, `Button1`, `TriggerHappy5` etc., it becomes unclear how inputs should be mapped to these capabilities, and we end up in the same place as we are now with the kernel and event codes being arbitrarily mapped to `BTN_TRIGGER_HAPPY_X`, etc.

Longer term, this does bring us to what we should actually do about these `BTN_TRIGGER_HAPPY` event codes that can vary depending on the input device. It's not ideal for us to be hard-coding the translation from evdev event code to InputPlumber Capability. I think the best way for us to handle this would be to implement specifying the raw evdev event codes in the capability map, as @blindedone1458 mentioned, and the source device implementation could use that to translate the event codes into the correct InputPlumber Capability. Maybe something like:

```yaml
mapping:
  - name: Guide
    source_events:
      - evdev:
          type: KEY
          code: KEY_F16
    target_event:
      gamepad:
        button: Guide
```

Expanding on this further, we could even update the capability map to add support for [SDL2 Gamepad Mappings](https://github.com/mdqinc/SDL_GameControllerDB/tree/master) and/or support hidraw input report decoding. Something like:

```yaml
mapping:
  - name: Guide
    source_events:
      - hidraw:
          report_id: 0x02
          input_type: u8
          byte_start: 8
          bits_offset: 2
    target_event:
      gamepad:
        button: Guide
```

ShadowApex (approved):

pastaq:
> In general, I strongly believe that a `Capability` in InputPlumber should be actual real capabilities of an input device. So in other words, I don't don't think we should ever have a `Capability::TriggerHappyX`. We should probably come to a consensus and codify this somewhere, but I think that we should try to follow these rules when adding new capabilities to InputPlumber:
> 
>     1. The capability should describe a real function of the device (i.e. no `Button1`, `TriggerHappy1`, etc.)
> 
>     2. The capability should describe either:
>        a. The input type and its location (e.g. `Axis::LeftStick`, `Axis::RightStick`, `Touchpad::Center`, etc.), or
>        b. The input type and its function (e.g. >`Button::Screenshot`, `Button::Guide`, etc.)

I'll just point out that a/b/x/y are also arbitrary alphabetical indexes and mapped differently to north/south/east/west depending on the manufacturer, and we still don't have support for c/z events, despite of them being next in the series and being used by dinput. The only difference is they aren't renamed in the kernel. BTN_C is just as valid a capability as BTN_NORTH. We need a unique way to identify these events so they can an be translated into a meaningful purpose. I.e. what do we do if a device produces left/right paddle events from an hidraw device driver and trigger happy 5-8 events from an evdev dpad. You run into collisions in the Capability map for unique identification. I'd much rather see us handle these edge cases by pushing unique capabilities and using capability maps then shoehorning in the first example we come across and dealing with conflicts later. I don't understand this mentality where it must provide functionality described by a noun for it to be a capability when we see examples of this not being the case in real world hardware.

blindedone1458:
A few suggestions to add to the discussion, you're both correct in my opinion. You have a need for both:

Defined capabilities based on real devices that are commonly used and thus have predictable and intuitive default mappings for devices. A user could reasonably discern that the X button on a DS4 controller is equivalent to South, but it may be confusing if a new Xbox controller's side buttons were defined as `BTN_23` instead of something like `LeftSide1`.

Additionally a pool of extraneous arbitrary capabilities with no defined defaults which allows user customization for their unique edge cases. Again a user could reasonably discern they wouldn't want to use  `gamepad:button: CustomBtn2` with `xb360` as their Target Device, but may want to emit an event that would only be recognized by their application. This would require an additional target-side mapping configuration to raw evdev / hidraw codes to truly be useful.

ShadowApex:
> I'll just point out that a/b/x/y are also arbitrary alphabetical indexes and mapped differently to north/south/east/west depending on the manufacturer, and we still don't have support for c/z events, despite of them being next in the series and being used by dinput. The only difference is they aren't renamed in the kernel. BTN_C is just as valid a capability as BTN_NORTH.

Right, which is why we don't have `Button::X`, `Button::Y`, `Button:A`, etc. defined as capabilities. I don't think we should be considering `BTN_NORTH`, `BTN_C`, etc. as capabilities, either. These are evdev event codes that we want to map to something meaningful: input that a virtual device is capable of emitting; i.e. a capability.

> We need a unique way to identify these events so they can an be translated into a meaningful purpose. I.e. what do we do if a device produces left/right paddle events from an hidraw device driver and trigger happy 5-8 events from an evdev dpad. You run into collisions in the Capability map for unique identification.

I don't see how this might be a problem. Based on your example, the capability map would look like this:

```yaml
mapping:
  - name: Left Paddle
    source_events:
      - hidraw:
          report_id: 0x02
          input_type: bool
          byte_start: 8
          bits_offset: 2
    target_event:
      gamepad:
        button: LeftPaddle
        
  - name: Dpad Left
    source_events:
      - evdev:
          type: KEY
          code: BTN_TRIGGER_HAPPY_5
    target_event:
      gamepad:
        button: DPadLeft
```

> I'd much rather see us handle these edge cases by pushing unique capabilities and using capability maps then shoehorning in the first example we come across and dealing with conflicts later.

This is true right now, but if we update our capability maps to allow mapping evdev event codes directly in the capability map, then there will be no conflicts. The evdev source device implementation can look at the capability map to determine how to translate these evdev event codes into InputPlumber capabilities.

> I don't understand this mentality where it must provide functionality described by a noun for it to be a capability when we see examples of this not being the case in real world hardware.

The main thing I think is  `Capability != Event Code`. Event codes, hidraw input reports, imu data are all ways hardware can communicate that some input is happening, but a `Capability` is the meaning/intent that we give to those inputs that we can map in a device independent way.

pastaq:
Okay, we'll need to refactor the source device implementations and add a lot to the config module to do that, but that's outside the scope for this PR. I suppose *for now* this change isn't adding any more of these problematic capabilities since it was just an omission when we added it at first. We can merge this and work on a more comprehensive solution after.

PR #297, opened 2025-02-23
Implement LEDs devices (Updated)

ShadowApex:
This PR includes @NeroReflex 's  LED implementation from #268 with review fixes so we can try to merge this in.

### DIFF
--- a/rootfs/usr/share/inputplumber/devices/50-rog_ally.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-rog_ally.yaml
@@ -53,6 +53,9 @@ source_devices:
         x: [1, 0, 0]
         y: [0, -1, 0]
         z: [0, 0, -1]
+  - group: led
+    led:
+      name: ASUS ROG Ally Config
   #- group: touchscreen
   #  udev:
   #    properties:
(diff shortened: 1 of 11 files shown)


NeroReflex:
> This PR includes @NeroReflex 's  LED implementation from #268 with review fixes so we can try to merge this in.

There is one thing @flukejones suggested me: to review basically how I get the multi_index part: originally I appended the path, but it makes probably more sense to get that as a driver property....

Anyway, sorry for my delay, I'm currently stuck in having to do different things..  So thank you for evolving my PR!

ShadowApex:
> Anyway, sorry for my delay, I'm currently stuck in having to do different things.. So thank you for evolving my PR!

No worries! I just wanted to get your changes in, so we can build on this more. Thanks for actually implementing this!

pastaq (approved):
Looks good for now. We can address the multi_index path and the MiltiColorChasssis naming in a follow on PR when we plumb this in with target devices and a dbus interface.

flukejones:
I'm okay to build on this with further PR. The important thing I think is to get it in to use quickly so we can then iterate on it.

PR #307, opened 2025-03-05
feat(Networking): add websocket client/server support

ShadowApex:
This change adds websocket source/target devices that use the [UCIS standard](https://github.com/ShadowBlip/UnifiedControllerInputSpec) for its wire protocol. To use this feature, you can now use the `websocket` target device which exposes an `org.shadowblip.Input.Websocket` dbus interface that can be used to connect to a remote instance of InputPlumber to route input over the network to another device.

To start a websocket server (i.e. a websocket source device), an entry can be added to a composite device config with the address/port to listen on:

```yaml
source_devices:
  - group: gamepad
    websocket:
      address: "0.0.0.0"
      port: 12907
```

Then using either the dbus interface, or the cli, you can connect remote composite device:

```bash
inputplumber device 0 connect ws://x.x.x.x:12907
```

This change required some amount of refactoring to use a new device enum instead of our previous assumption that all source devices would be udev devices.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -230,6 +230,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "block-buffer"
+version = "0.10.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3078c7629b62d3f0439517fa394996acacc5cbc91c5a20d8c658e77abd503a71"
+dependencies = [
+ "generic-array",
+]
+
+[[package]]
 name = "bumpalo"
 version = "3.16.0"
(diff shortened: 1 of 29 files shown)


R1kaB3rN:
As a future and optional enhancement to do this, what do you think about creating a secure channel protocol between the client/server? I don’t imagine ensuring confidentiality to be a requirement, but I think strong, mutual authentication would be great. The noise protocol framework as used by WireGuard makes this pretty straightforward to design and implement one.

pastaq:
Aside from a rebase, what is this waiting on?

PR #309, opened 2025-03-09
chore: add debian packaging

NeroReflex:
This PR add support for building .deb packages on both debian 12 and ubuntu 24.04 LTS

I also added the authors field otherwise cargo deb will fail since .deb packages requires a maintainer field.

I tested the CI on my fork coping both blocks into the "build" hook.


pastaq (approved):

pastaq:
Can you link the successful CI from your branch?

0001-chore-add-debian-package-to-workflow.patch
```diff
@@ -0,0 +1,66 @@
+From 73a7363ea1bbb8622ded0507dc12ba0483523894 Mon Sep 17 00:00:00 2001
+From: Denis Benato <benato.denis96@gmail.com>
+Date: Sun, 9 Mar 2025 16:11:05 +0100
+Subject: [PATCH 1/2] chore: add debian package to workflow
```
ShadowApex, line 4: We shouldn't be adding these patches to source control here.

.github/workflows/release.yaml
```diff
@@ -138,3 +138,95 @@ jobs:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install basic dependencies
+        id: install-software
+        run: |
+          export DEBIAN_FRONTEND=noninteractive
+          apt-get update
+          apt-get install -y wget git lsb-release wget software-properties-common gnupg curl build-essential devscripts debhelper
```
ShadowApex, line 160: I would prefer if we put these steps inside a `Dockerfile` and adding a new target in our `Makefile` to build the debian package so it's easy for us to build the deb package without the GitHub action.

This also makes it easier for us to switch to a different CI solution other than GitHub Actions if we ever need to do so.

ShadowApex (changes requested):
Few comments, but thank you for these changes!

0001-chore-add-debian-package-to-workflow.patch
NeroReflex: Damn. My bad.

NeroReflex:
> Can you link the successful CI from your branch?

https://github.com/NeroReflex/InputPlumber/actions/runs/13750158722/job/38450112628

.github/workflows/release.yaml
NeroReflex: I have created a dist-deb very similar to the dist-rpm.
I was unable to make use of an in-repo Dockerfile without giving direct docker commands. I personally find that a bit ugly and unnecessary given dependencies are listed there, but if you want that I can put those command in the github CI, just be aware the file won't look particularly nice afterward

pastaq: What is the plan for this?

PR #311, opened 2025-03-10
fix(Generic Gamepads): re-enable support for generic gamepads

ShadowApex:
This change re-enables support for generic gamepads. This config will only be used if `manage-all` is set to `true` by an external process.


pastaq (approved):
Please have ChimeraOS folks test this before we push to main. I don't want to break them again.

@NeroReflex

NeroReflex:
Opened the overlay with the rog ally button and external joystock goes through steam: ogui ignores it completely but that's maybe because it's not the thing the opened the menu?
I don't even know how to open the menu with my xbox one controller. It doesn't have the share button and the cental xbox button opens steam's menu.
How do I check if it has manage-all set?

PR #320, opened 2025-03-20
fix(Legion Go S): Fix evdev detection

pastaq:
- Removes PID glob matching for the event devices on the Go S. It is causing issues blocking all evdevs.


ShadowApex (approved):

XinoAssassin:
The config checked pass in our tests.

PR #324, opened 2025-03-24
fix(MSI Claw): add support for setting XInput mode for systems without kernel support

ShadowApex:
This change adds support for switching the MSI Claw into XInput mode for systems that do not have kernel support. Once Nero's driver has been upstreamed to the kernel, we can update this method to use the sysfs interface when it is available.

### DIFF
--- a/rootfs/usr/share/inputplumber/schema/composite_device_v1.json
+++ b/rootfs/usr/share/inputplumber/schema/composite_device_v1.json
@@ -486,23 +486,23 @@
           }
         }
       }
-    }
-  },
-  "Color": {
-    "type": "object",
-    "additionalProperties": false,
-    "properties": {
-      "r": {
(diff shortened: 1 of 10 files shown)


rootfs/usr/share/inputplumber/schema/composite_device_v1.json
pastaq, line 1: Why was this changed in this pr?

pastaq (approved):
Seems fine, but we should avoid this pattern if we can. One issue with getting something working in userspace is the incentive to get it working in the kernel is diminished. This should work with or without inputplumber.

rootfs/usr/share/inputplumber/schema/composite_device_v1.json
ShadowApex: Schema validation was failing when I was updating the device config.

PR #326, opened 2025-03-30
fix(Manager): add whitelist to manage specific virtual devices

ShadowApex:
This change adds a whitelist of virtual devices that InputPlumber can manage. Namely Sunshine virtual gamepads to allow sunshine/moonlight streaming.

In the future, we may want to move this whitelist into a configuration file.

### DIFF
--- a/src/input/manager.rs
+++ b/src/input/manager.rs
@@ -2,7 +2,6 @@ use core::panic;
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::error::Error;
-use std::fs;
 use std::path::PathBuf;
 use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


src/input/manager.rs
```diff
@@ -1210,11 +1210,26 @@ impl Manager {
                         }
                     };
 
-                    if !is_bluetooth {
+                    // Some virtual gamepads we DO want to manage
+                    const VIRT_DEVICE_WHITELIST: &[&str] = &[
```
pastaq, line 1214: I there somewhere else we can keep this const, such as the top of the file or in udev.rs? Doing it inline here will make it harder to keep track of to add more and it is only a few bytes.

pastaq (changes requested):
Seems fine, but I'd prefer to declare the cost somewhere else.

pastaq (approved):

PR #308, opened 2025-03-05
feat(Flydigi): Add support for flydigi vader 3/4 pro

Fabianoshz:
This should add support for both flydigi vader 3 and 4.

Fixes #208

PR #327, opened 2025-04-02
Add support for Flydigi Vader 4 Pro 

winghugs:
This allows InputPlumber to pick up and read the Vader 4 Pro controller as a Steam Deck controller, including the two extra buttons on the Vader 4 Pro being read as track pad presses so they are fully SteamInput mappable.

Three of the buttons (HAPPY11-HAPPY13) were not listed to be picked up, and I mapped them to be specifically redundant to the two HAPPY buttons listed that this controller does not pick up, as well as adding the Fn button on the controller as the QuickAccess menu.

I specifically target the Vader 4 Pro in the device config, so there should not be any false positives with this mapping.

This relies on this xpad fork here which allows the buttons to be detectable in evdev: https://github.com/ahungry/xpad/tree/feature/flydigi-vader3-extra-buttons


pastaq:
This appears to be a duplicate of #308.

In both cases, we should wait for capability map v2 to be finished. The current issue is that there's no standard for what the happy trigger events do, so by shoehorning in a capability to them we canonicalize something that isn't a standard, which will inevitably have conflicts in the future. Capability map v2 will allow us to directly translate the evdev events per source device into capabilities. That will allow us to map unique devices like this without introducing potential conflicts with other devices.

winghugs:
>This appears to be a duplicate of https://github.com/ShadowBlip/InputPlumber/pull/308.

This catches C and Z buttons as well allowing them to be mapped.

> In both cases, we should wait for capability map v2 to be finished.

I agree, I will close this and will make a capability map for this controller when that arrives.

PR #328, opened 2025-04-03
feat(SteamDeck): Add Steam Deck UHID target device

pastaq:


src/input/target/steam_deck_uhid.rs
```diff
@@ -0,0 +1,851 @@
...
+        let Some(report_id) = data.get(1) else {
+            return Ok(vec![]);
+        };
+        self.current_report = (*report_id).try_into()?;
+        log::debug!("True Report ID: {:#04x}", (*report_id));
+        let output_events = match self.current_report {
+            ReportType::TriggerHapticCommand => {
+                let buf = data.as_slice().try_into()?;
+
+                let packed_haptic_report = PackedHapticReport::unpack(buf)?;
```
ShadowApex, line 499: It's possible that this data buffer also has a preceding `0x00` byte, so we may want to take that into account if these haptic/rumble reports don't work.

ShadowApex (approved):
Looks good!

src/input/target/steam_deck_uhid.rs
pastaq: I was thinking the same

src/input/target/steam_deck_uhid.rs
```diff
@@ -0,0 +1,851 @@
...
+        &mut self,
+        id: u32,
+        _report_number: u8,
+        _report_type: uhid_virt::ReportType,
+        data: Vec<u8>,
+    ) -> Result<Vec<OutputEvent>, Box<dyn Error>> {
+        let Some(report_id) = data.get(1) else {
+            return Ok(vec![]);
+        };
+        self.current_report = (*report_id).try_into()?;
```
ShadowApex, line 493:
```suggestion
        self.current_report = match (*report_id).try_into() {
            Ok(id) => id,
            Err(err) => {
                log::warn("Unknown report id: {report_id}");
                return Ok(());
            }
        };
```

ShadowApex (changes requested):

src/input/target/steam_deck_uhid.rs
pastaq: Removing the first byte of the data buffer fixed haptics.

src/input/target/steam_deck_uhid.rs
pastaq: I did something similar, but this had some issues.

ShadowApex (approved):

ShadowApex (approved):

src/input/target/mod.rs
```diff
@@ -59,7 +61,7 @@ pub mod xbox_series;
 /// Possible errors for a target device client
 #[derive(Error, Debug)]
 pub enum InputError {
-    #[error("error occurred running device")]
+    #[error("error occurred running input device")]
```
ShadowApex, line 64: Since you're touching this, could you update this to ensure we also print the downstream error message?:
```suggestion
    #[error("error occurred running input device: `{0}`")]
```

(Usage docs for `thiserror`: https://github.com/dtolnay/thiserror?tab=readme-ov-file#details)

src/input/target/mod.rs
```diff
@@ -103,9 +105,9 @@ impl From<ClientError> for InputError {
 #[derive(Error, Debug)]
 pub enum OutputError {
     #[allow(dead_code)]
-    #[error("behavior is not implemented")]
+    #[error("output behavior is not implemented")]
     NotImplemented,
-    #[error("error occurred running device")]
+    #[error("error occurred running output device")]
```
ShadowApex, line 110: Same here:
```suggestion
    #[error("error occurred running output device: `{0}`")]
```

ShadowApex (changes requested):

ShadowApex (approved):

PR #329, opened 2025-04-04
feat(Capability Map): add generic evdev translator

ShadowApex:
This change implements a generic evdev event translator that uses the new Capability Map v2 configuration to allow mapping specific evdev events to native InputPlumber events.

Example:

```yaml
# yaml-language-server: $schema=https://raw.githubusercontent.com/ShadowBlip/InputPlumber/main/rootfs/usr/share/inputplumber/schema/capability_map_v2.json
version: 2
kind: CapabilityMap
name: Example
id: example

# List of mapped events
mapping:
  - name: A to B
    source_events:
      - evdev:
          event_type: KEY
          event_code: BTN_SOUTH
          value_type: button
    target_event:
      gamepad:
        button: East

  - name: B to A
    source_events:
      - evdev:
          event_type: KEY
          event_code: BTN_EAST
          value_type: button
    target_event:
      gamepad:
        button: South

  - name: X + Y to Start
    mapping_type:
      evdev: chord
    source_events:
      - evdev:
          event_type: KEY
          event_code: BTN_WEST
          value_type: button
      - evdev:
          event_type: KEY
          event_code: BTN_NORTH
          value_type: button
    target_event:
      gamepad:
        button: Start

  - name: X to Y delay
    mapping_type:
      evdev: delayed_chord
    source_events:
      - evdev:
          event_type: KEY
          event_code: BTN_NORTH
          value_type: button
    target_event:
      gamepad:
        button: West
```

It also includes device/capability map configs for the 8BitDo and PS4 controllers.


rootfs/usr/share/inputplumber/capability_maps/dinput_generic.yaml
pastaq, line 1: Was this verified on a dinput device? I believe many of the default events are dissimilar to xinput events and don't translate intuitively.

pastaq (approved):
This seems fine for now, but we need to discuss what we're going to do with other event device types. Keyboard device types for example aren't on this pattern and will need a unique implementation to ensure non-mapped input is blocked. Touchscreen and touchpad might work with this implementation, but need to be tested. It would be preferable to not need to reinvent this 3 more times.

rootfs/usr/share/inputplumber/capability_maps/dinput_generic.yaml
ShadowApex: I tested this on the 8BitDo Pro 2 in dinput mode.

pastaq (approved):

PR #331, opened 2025-04-08
fix(Device Hiding): add ignore for libinput

ShadowApex:
This change updates the hiding rules to include hiding devices from libinput:
https://wayland.freedesktop.org/libinput/doc/latest/ignoring-devices.html


pastaq (approved):

pastaq:
What is this waiting on?

ShadowApex:
I would like to do more testing with this change to make sure it doesn't introduce any unintended consequences.

PR #332, opened 2025-04-09
Add driver for the Flydigi Vader 4 Pro - DInput Mode

winghugs:
This is a driver written for the Flydigi Vader 4 Pro to allow for all features to be used in InputPlumber without any additional driver installs or changes. This driver is specifically written to work with the controller in DInput mode only, plugged in or via the included dongle. The XInput mode should not be caught up in the detection at all, so can still be used.

Rumble and motion controls are not implemented at the moment, but are planned.

Two important notes if folks stumble on this:
- The Fn button (currently set to the quick action menu) puts the device in a "mouse" mode on a hardware level. This doesn't change most device functionality, except the furthest left back button starts only being pressed on repeat when held. There may be other quirks - basically if you need to hit this button, make sure you hit it a second time before getting back to your game to disable the mouse mode.

- The C and Z buttons are mapped specifically to touching the left and right Steam Deck trackpads. I would recommend mapping the trackpads to single button, and using the touch input to map these.

This can easily be changed to the Dualsense Edge or Xbox Elite Controller should you need those devices for particular games, you will just want to load a profile for those games if you want C and Z functionality. They are hard mapped to LeftPaddle3 (C) and RightPaddle3 (Z)


src/drivers/flydigi_vader_4_pro/hid_report_test.rs
```diff
@@ -0,0 +1,35 @@
...
+use super::hid_report::PackedInputDataReport;
+
+#[tokio::test]
+async fn test_flydigi_vader_4_pro() -> Result<(), Box<dyn Error>> {
+    let report = PackedInputDataReport::unpack_from_slice(&DATA_A).unwrap();
+    println!("{report}");
+    assert!(report.a, "should be pressed");
+
+    Ok(())
+}
```
ShadowApex, line 14: Since this is not a valid passing test, you should remove this file or create a working test.

ShadowApex (changes requested):
Thanks for your work on this! In order for us to merge this in, could you run `cargo-fmt` to format all your changes according to the standard? Once this is done, and the testing pipeline passes, we should be able to merge this.

Thanks!

src/drivers/flydigi_vader_4_pro/mod.rs
```diff
@@ -2,5 +2,4 @@ pub mod driver;
 pub mod event;
 pub mod hid_report;
 #[cfg(test)]
```
ShadowApex, line 4: You also need to remove this `#[cfg(test)]` line here.

winghugs:
> Thanks for your work on this! In order for us to merge this in, could you run `cargo-fmt` to format all your changes according to the standard? Once this is done, and the testing pipeline passes, we should be able to merge this.
> 
> Thanks!

this is done - cargo-fmt has no output, so I'm assuming we're golden?

I'll get another PR in for gyro when I finally get it working.

ShadowApex:
> this is done - cargo-fmt has no output, so I'm assuming we're golden?
> 
> I'll get another PR in for gyro when I finally get it working.

After running `cargo-fmt` from the project directory, you'll need to commit the changes it made. It looks like the formatted version has not been committed yet.

winghugs:
> After running `cargo-fmt` from the project directory, you'll need to commit the changes it made. It looks like the formatted version has not been committed yet.

got it, was expecting it to give me some kind of output. those changes are pushed.

ShadowApex (approved):

bugQ:
awesome work, tho now my pad does not appear at all, whereas with just the standard Linux kernel DInput driver, it at least presented most of its inputs. do i need to configure something ? apologies as somebody new to InputPlumber

winghugs:
> awesome work, tho now my pad does not appear at all, whereas with just the standard Linux kernel DInput driver, it at least presented most of its inputs. do i need to configure something ? apologies as somebody new to InputPlumber

This configures the controller to be read as a steam deck, which also means steam input is needed to configure it. An alternative is using it as a dualsense edge or xbox elite controller, however you will lose C and Z functionality. I do plan on figuring out proper dualsense edge emulation that keeps C and Z but I need to adjust the target in order to feasibly do that. 

Does steam not pick up the controller?

bugQ:
> Does steam not pick up the controller?

yeah, nor does it appear in in the Game Controller module of ksystemsettings anymore

winghugs:
> > Does steam not pick up the controller?
> 
> yeah, nor does it appear in in the Game Controller module of ksystemsettings anymore

are you able to run it in debug mode and provide the log output when you connect the controller?

`sudo LOG_LEVEL=debug inputplumber 2>&1 | tee .inputplumber.log`

bugQ:
sure thing, here:

> > > Does steam not pick up the controller?
> > 
> > 
> > yeah, nor does it appear in in the Game Controller module of ksystemsettings anymore
> 
> are you able to run it in debug mode and provide the log output when you connect the controller?
> 
> `sudo LOG_LEVEL=debug inputplumber 2>&1 | tee .inputplumber.log`

you got it: [inputplumber.log](https://github.com/user-attachments/files/20008785/inputplumber.log)

winghugs:
> > > > Does steam not pick up the controller?
> > > 
> > > 
> > > yeah, nor does it appear in in the Game Controller module of ksystemsettings anymore
> > 
> > 
> > are you able to run it in debug mode and provide the log output when you connect the controller?
> > `sudo LOG_LEVEL=debug inputplumber 2>&1 | tee .inputplumber.log`
> 
> you got it: [inputplumber.log](https://github.com/user-attachments/files/20008785/inputplumber.log)

thank you!

could you give me some information about your OS? do you have any udev rules that might be picking this up?

also on the controller - the switch on the back is all the way to the right and it's plugged in either wired or dongle, correct? the light is also blue in the bottom middle, not white? this doesn't work for bluetooth yet.

bugQ:
> > you got it: [inputplumber.log](https://github.com/user-attachments/files/20008785/inputplumber.log)
> 
> thank you!

:+1:

> could you give me some information about your OS? do you have any udev rules that might be picking this up?

it's NixOS unstable (i.e. the branch soon to be stabilized as nixos-2025.05). the inputplumber package is at version 0.53.0 and was installed by setting `services.inputplumber.enable = true;`. here's a udevadm test on the device in question: [udevadm-test.log](https://github.com/user-attachments/files/20009737/udevadm-test.log)

> also on the controller - the switch on the back is all the way to the right and it's plugged in either wired or dongle, correct? the light is also blue in the bottom middle, not white? this doesn't work for bluetooth yet.

if you mean the "right" relative to the face of the controller, then yes; it's the setting marked by an icon of the dongle. i was using a USB cable for the given log.

bugQ:
>   systemd.services.inputplumber.serviceConfig = {
>     Environment="PATH=/run/current-system/sw/bin";
>   };

it turns out this is the only part i needed! thank you! this should really be part of the default config for this service, since that path is where nearly all executables go on NixOS. time for a PR...

PR #336, opened 2025-04-17
[draft]: Test zotac zone driver

flukejones:


pastaq:
You're using capability map v2, but have it set as the composite device capability map. V2 maps are per source device, so you need to attach it to the source device(s) that produces the events.

I.E.
https://github.com/ShadowBlip/InputPlumber/blob/9c13d3c751da7adb12e531c7378db4bef471872b/rootfs/usr/share/inputplumber/devices/60-ps4_gamepad.yaml#L26

flukejones:
Ah thanks. That solves it. Now to correctly remap things

hphilm:
BOSCH IMU would be:

```
  # IMU
   - group: imu
     iio:
      name: i2c-BMI0160:00
```

hphilm:
We have some conflicting events when one mapping file is used for same events:

Seems we have two events, one for the left dial `REL_HWHEEL` and one for the right dial `REL_WHEEL`. The later will conflict with the left touchpad (mouse) when only one mapping file is used as the event is the same.

```
  - name: Left Dial Counter-clockwise
    source_events:
      - evdev:
          event_type: REL
          event_code: REL_HWHEEL
          axis_direction: negative
          value_type: button
    target_event:
      keyboard: KeyVolumeDown
  - name: Left Dial Clockwise
    source_events:
      - evdev:
          event_type: REL
          event_code: REL_HWHEEL
          axis_direction: positive
          value_type: button
    target_event:
      keyboard: KeyVolumeDown
  - name: Right Dial Counter-clockwise
    source_events:
      - evdev:
          event_type: REL
          event_code: REL_WHEEL
          axis_direction: negative
          value_type: button
    target_event:
      keyboard: KeyVolumeDown
  - name: Right Dial Clockwise
    source_events:
      - evdev:
          event_type: REL
          event_code: REL_WHEEL
          axis_direction: positive
          value_type: button
    target_event:
      keyboard: KeyVolumeDown
```

Event `REL_WHEEL` is also issued by the dials. So a separate mapping file might be needed for those.

``` 
 - name: Left Touchpad
    source_events:
      - evdev:
          event_type: REL
          event_code: REL_WHEEL # horizontal plus vertical
          value_type: button
    target_event:
      mouse:
        motion:
          direction: horizontal
```

hphilm:
From William:

For the dials, you need to:
- Add a new "gamepad dial" capability to inputplumber for the left and right dials
- Map the dial events to the new dial capability
- Use `value_type: axis` instead of `value_type: button` so it can interpret the value correctly
- Update the default input profile to map the dial input to whatever you want (e.g. volume up/down)

For the touchpads, you need to:
- Look at the hidraw input reports for the absolute values of the touches
- Add that to the Zotac Zone inputplumber driver so it can read those touch events instead of reading the REL events from evdev.

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
```diff
@@ -0,0 +1,145 @@
...
+      keyboard: KeyVolumeDown
+  - name: Right Dial Clockwise
+    source_events:
+      - evdev:
+          event_type: REL
+          event_code: REL_WHEEL
+          axis_direction: positive
+          value_type: button
+    target_event:
+      keyboard: KeyVolumeDown
```
ShadowApex, line 105: The capability map is meant to be used to fix devices that emit the wrong input events, and map them to the _real_ capabilities of the input device. The Zotac Zone dials are a new type of input, not volume buttons, so these should be mapped to a new "Dial" input capability instead:

```suggestion
      gamepad:
        dial: RightStickDial
```

To do this, you'll need to add the new capability in `capability.rs` and in the config module.

The reason we want to create a new capability for this new type of input is it allows us to distinguish between volume buttons and these dials, so they can be mapped properly in an input profile. Without a new capability, we would not be able to make that distinction.

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
```diff
@@ -0,0 +1,145 @@
...
+          value_type: button
+    target_event:
+      mouse:
+        button: Middle
+  - name: Left Touchpad
+    source_events:
+      - evdev:
+          event_type: REL
+          event_code: REL_WHEEL # horizontal plus vertical
+          value_type: button
```
ShadowApex, line 138: Using `value_type: button` will make InputPlumber interpret the given value as a boolean, which is not what you want for touchpads. Touchpads should also be `ABS` input, instead of `REL` input, as it's impossible to translate relative input to absolute input in order to correctly map these to other touch devices. Does the hidraw input report expose the touchpad inputs as absolute values?

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
```diff
@@ -0,0 +1,145 @@
...
+    target_event:
+      mouse:
+        motion:
+          direction: horizontal
```
ShadowApex, line 142: The target mapping here should be a touchpad, not a mouse.

ShadowApex (changes requested):

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
```diff
@@ -0,0 +1,145 @@
...
+          value_type: button
+    target_event:
+      keyboard: KeyVolumeDown
+  - name: Right Dial Clockwise
+    source_events:
+      - evdev:
+          event_type: REL
+          event_code: REL_WHEEL
+          axis_direction: positive
+          value_type: button
```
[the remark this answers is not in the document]
flukejones: This doesn't work.

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
flukejones: Ah, I see now. I'll add this.

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
flukejones: Removed touchpads. Left one has something funky going on and right works as-is.

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
flukejones: Except these are relative events so it seems to break expectations in code somewhere. They are removed now anyways.

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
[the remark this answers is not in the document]
pastaq: It would be a new value type, so it needs to be added.

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
pastaq: We should revisit that in the future. The touchpads can be mapped to many different things in steam input and the deck target enables haptic feedback.

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
```diff
@@ -0,0 +1,111 @@
...
+  - name: Right Dial Clockwise
+    source_events:
+      - evdev:
+          event_code: REL_WHEEL
+          event_type: REL
+          axis_direction: positive
+          value_type: joystick_x
+    target_event:
+      dial:
+        direction: counter_clockwise
```
ShadowApex, line 108:
```suggestion
      gamepad:
        dial:
          name: RightStickDial
```

flukejones:
I added a Dial capability but I have zero idea if this is what was expected, or what it's expected to be used for, or if it even works. I imagine it would be best used with UCIS?

hphilm:
I will test drive the inputplumber configuration as given without dials and mouse support over the weekend. I see dials work has moved to #339

flukejones:
Close in favour of #339

PR #337, opened 2025-04-17
fix(Manager): emit gamepad order changed signal on order change

ShadowApex:
This change updates the `Manager` and `ManagerInterface` to emit dbus signals whenever the gamepad order has changed. It also caches the gamepad order in the `ManagerInterface`, so dbus queries do not need to communicate with the `Manager` directly to get the current gamepad order.

In the future we should try to follow this pattern of caching data in the dbus interface struct so queries to dbus do not affect the performance of the underlying implementation. i.e. before caching, a client could send a bunch of dbus requests to get the gamepad order which the `Manager` would need to handle, which would block it from doing other work to process those requests.

### DIFF
--- a/src/dbus/interface/manager.rs
+++ b/src/dbus/interface/manager.rs
@@ -1,11 +1,12 @@
 use std::time::Duration;
 
 use tokio::sync::mpsc;
-use zbus::fdo;
+use zbus::{fdo, Connection};
 use zbus_macros::interface;
 
 use crate::{
     config::CompositeDeviceConfig,
+    constants::BUS_PREFIX,
     input::{manager::ManagerCommand, target::TargetDeviceTypeId},
 };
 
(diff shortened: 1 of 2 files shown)


pastaq (approved):

pastaq:
> In the future we should try to follow this pattern of caching data in the dbus interface struct so queries to dbus do not affect the performance of the underlying implementation. i.e. before caching, a client could send a bunch of dbus requests to get the gamepad order which the Manager would need to handle, which would block it from doing other work to process those requests.

Agreed. This is a good first step.

We should consider if we want or need to have suspend hooks anymore. If we depreciate the vchi deck target there isn't really a need we're filling anymore and it does confuse games a lot more when everything disconnects/reconnects.

pastaq (approved):

PR #338, opened 2025-04-18
feat(Legion Go S): Add touchpad support

pastaq:
- Adds touchpad support for the Legion Go S.
- Splits IMU, Touchpad, and Xinput devices for Legion Go S. Source of bad
  data was identified as the input buffer filling when one of the
  three input devices was grabbed. Since we were reading up to 32b for
  xinput it would incorrectly detect as an xinput event. Splitting the
  source devices ensures the buffer can be limited to a single event of each size.
- Fixes deck targets click to select in desktop mode by enabling touchpad force events on targets.


src/drivers/legos/imu_driver.rs
```diff
@@ -0,0 +1,180 @@
...
+pub const DINPUT_PID: u16 = 0xe311;
+pub const PIDS: [u16; 2] = [XINPUT_PID, DINPUT_PID];
+pub const IID: i32 = 0x5;
+
+// Input report sizes
+const INERTIAL_PACKET_SIZE: usize = 9;
+const HID_TIMEOUT: i32 = 10;
+
+// Input report axis ranges
+pub const GYRO_SCALE: i16 = 2;
```
ShadowApex, line 23: Is there any information about this scale value?

src/drivers/legos/touchpad_driver.rs
```diff
@@ -0,0 +1,170 @@
...
+                Ok(events) => events,
+                Err(e) => {
+                    log::error!("Got error processing TouchInputDataReport: {e:?}");
+                    vec![]
+                }
+            };
+        }
+
+        // There is no release event, so check to see if we are still touching.
+        if self.is_touching && (self.last_touch.elapsed() > Duration::from_millis(25)) {
```
ShadowApex, line 86: Can we define this `Duration` as a constant at the top of this file?

src/drivers/legos/touchpad_driver.rs
```diff
@@ -0,0 +1,170 @@
...
+            return events;
+        };
+
+        // Binary events
+        if state.pressed != old_state.pressed {
+            events.push(Event::Button(ButtonEvent::RPadPress(BinaryInput {
+                pressed: state.pressed,
+            })));
+            events.push(Event::Trigger(TriggerEvent::RpadForce(TriggerInput {
+                value: 32 * state.pressed as u8,
```
ShadowApex, line 150: Is there a reason to multiply this by 32?

ShadowApex (approved):

src/drivers/legos/touchpad_driver.rs
pastaq: Pressed is a bool. I'm emulating about 1/4 max pressure which, from my testing, is about what a normal couch on the deck is. There's no pressure sensor on the go s.

src/drivers/legos/imu_driver.rs
pastaq: IIRC it was added to fix the deg/s to more closely match the Deck. It's been a while and in this case was just moved.

PR #340, opened 2025-04-18
fix(Target Touchpad): use higher resolution and polling rate to fix slow, jittery touch behavior

ShadowApex:
This change fixes slowness, jitter, and overall bad feeling of virtual touchpads. An example of this can be found in the following review video:

[ETA Prime Video](https://www.youtube.com/embed/lU9LkJYfd68?start=137&end=158)
https://www.youtube.com/embed/lU9LkJYfd68?start=137&end=158

The main change that fixed this issue was to change the polling rate of the target touchpad device from 13ms to 3ms.

### DIFF
--- a/src/input/target/mod.rs
+++ b/src/input/target/mod.rs
@@ -694,7 +694,7 @@ impl TargetDevice {
             "touchpad" => {
                 let device = TouchpadDevice::new()?;
                 let options = TargetDriverOptions {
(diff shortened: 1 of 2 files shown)


pastaq (approved):

pastaq:
Was this tested in both game mode and desktop?

ShadowApex:
> Was this tested in both game mode and desktop?

Yes, I have tested this in both modes.

PR #342, opened 2025-04-23
feat(LED): add dbus interface to control source LED color

ShadowApex:
This change adds a new `org.shadowblip.Output.Source.LED` dbus interface on any source device that supports the `OutputCapability::LED(LED::Color)` capability to allow LED control over dbus.

```bash
$ busctl introspect org.shadowblip.InputPlumber /org/shadowblip/InputPlumber/devices/source/hidraw5
NAME                                     TYPE      SIGNATURE RESULT/VALUE                             FLAGS
org.shadowblip.Input.Source.UdevDevice   interface -         -                                        -
.DevicePath                              property  s         "/dev/hidraw5"                           emits-change
.IdBustype                               property  s         "3"                                      emits-change
.IdProduct                               property  s         "0ce6"                                   emits-change
.IdVendor                                property  s         "054c"                                   emits-change
.IdVersion                               property  s         "33041"                                  emits-change
.Name                                    property  s         "Sony Interactive Entertainment DualSen… emits-change
.PhysPath                                property  s         ""                                       emits-change
.Properties                              property  a{ss}     15 "CURRENT_TAGS" ":uaccess:seat:" "DEV… emits-change
.Subsystem                               property  s         "hidraw"                                 emits-change
.SysfsPath                               property  s         "/devices/pci0000:00/0000:00:08.3/0000:… emits-change
org.shadowblip.Output.Source.LED         interface -         -                                        -
.SetColor                                method    yyy       -                                        -
```

The LED color can be set by calling the `SetColor` method on this interface:

```bash
busctl call org.shadowblip.InputPlumber \
    /org/shadowblip/InputPlumber/devices/source/hidraw5 \
    org.shadowblip.Output.Source.LED \
    SetColor "yyy" 255 0 255
```

### DIFF
--- a/src/dbus/interface/composite_device.rs
+++ b/src/dbus/interface/composite_device.rs
@@ -9,6 +9,7 @@ use zbus_macros::interface;
 
 use crate::{
     config::DeviceProfile,
+    constants::BUS_SOURCES_PREFIX,
     input::{
         capability::{Capability, Gamepad, Mouse},
         composite_device::{client::CompositeDeviceClient, InterceptMode},
@@ -321,6 +322,26 @@ impl CompositeDeviceInterface {
         }
 
         Ok(capability_strings)
+    }
+
+    /// Source devices that this [CompositeDevice] is managing
+    #[zbus(property)]
+    async fn source_devices(&self) -> fdo::Result<Vec<String>> {
+        let paths = self
+            .composite_device
+            .get_source_device_paths()
+            .await
+            .map_err(|e| fdo::Error::Failed(e.to_string()))?;
+        let paths = paths
+            .into_iter()
+            .filter_map(|p| {
+                let last = p.split("/").last()?;
+                let name = last.replace(':', "_");
+                Some(format!("{BUS_SOURCES_PREFIX}/{name}"))
+            })
+            .collect();
+
+        Ok(paths)
     }
 
     /// List of source devices that this composite device is processing inputs for
--- a/src/dbus/interface/source/mod.rs
+++ b/src/dbus/interface/source/mod.rs
@@ -2,4 +2,5 @@ pub mod evdev;
 pub mod hidraw;
 pub mod iio_imu;
 pub mod led;
+pub mod output;
 pub mod udev;
--- a/src/dbus/interface/source/output.rs
+++ b/src/dbus/interface/source/output.rs
@@ -1,0 +1,87 @@
+use crate::{
+    constants::BUS_SOURCES_PREFIX,
+    input::{output_event::OutputEvent, source::command::SourceCommand},
+};
+use std::{error::Error, time::Duration};
+use tokio::sync::mpsc::Sender;
+use zbus::{fdo, Connection};
+use zbus_macros::interface;
+
+/// DBusInterface exposing LED output
+pub struct SourceOutputLedInterface {
+    tx: Sender<SourceCommand>,
+}
+
+impl SourceOutputLedInterface {
+    fn new(tx: Sender<SourceCommand>) -> SourceOutputLedInterface {
+        SourceOutputLedInterface { tx }
+    }
+
+    /// Creates a new instance of the source led interface on DBus.
+    pub fn listen_on_dbus(
+        conn: Connection,
+        device_id: &str,
+        tx: Sender<SourceCommand>,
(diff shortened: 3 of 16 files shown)


pastaq (commented):
We need a brightness value. Led_mc provides a multi_intensity fd for seeing color and a brightness fd for scaling the color to s max brightness. Only setting one won't work if the brightness is 0.

ShadowApex:
> We need a brightness value. Led_mc provides a multi_intensity fd for seeing color and a brightness fd for scaling the color to s max brightness. Only setting one won't work if the brightness is 0.

As far as I understand, an overall brightness value is not a real thing in LEDs. The brightness fd in the kernel interface just provides some ergonomics for adjusting the overall color intensity of each RGB value. E.g.  `{R: 255, G: 255, B: 255, Brightness: 128}`  is the equivalent to `{R: 128, G: 128, B: 128}`.

pastaq:
> > We need a brightness value. Led_mc provides a multi_intensity fd for seeing color and a brightness fd for scaling the color to s max brightness. Only setting one won't work if the brightness is 0.
> 
> As far as I understand, an overall brightness value is not a real thing in LEDs. The brightness fd in the kernel interface just provides some ergonomics for adjusting the overall color intensity of each RGB value. E.g. `{R: 255, G: 255, B: 255, Brightness: 128}` is the equivalent to `{R: 128, G: 128, B: 128}`.

1.) That isn't universally true. Some devices have a distinct brightness value where the MCU firmware does this. Go S does this as an example.
2.) Even in the cases where that isn't true, the kernel drivers all expect a distinct brightness value and will scale it internally. 

With those issues combined, failing to set a brightness to sysfs when it is lower than 255 will fail to turn on the LED's (0) or set it to the desired brightness (1-254).

ShadowApex:
> 1.) That isn't universally true. Some devices have a distinct brightness value where the MCU firmware does this. Go S does this as an example.

That is the device firmware choosing to do those same color intensity calculations as you would in software. There is no "brightness" wire that connects to a physical LED as far as I have ever seen. There's no "backlight" in an LED.

>2.) Even in the cases where that isn't true, the kernel drivers all expect a distinct brightness value and will scale it internally.
> 
> With those issues combined, failing to set a brightness to sysfs when it is lower than 255 will fail to turn on the LED's (0) or set it to the desired brightness (1-254).

These are implementation details that should be left to each individual implementation. What we should concentrate on here is how we want to expose this functionality to the user. The individual LED implementation in InputPlumber is responsible for setting the correct sysfs values based on the color intensity values. Exposing the RGB values alone will afford the exact same array of LED color that you could with RGB + Brightness. Our DBus interface does not need to be identical to the kernel interface.

If we _really_ want to support the widest array of LED hardware we can with this interface, it would be better for this interface to support any array of colors, instead of just RGB, but I just wanted to create a simple interface to get something working. 

If you think we should do this instead, and/or have strong feeling about a `brightness` value, I can scrap this PR and refactor it.

pastaq:
> That is the device firmware choosing to do those same color intensity calculations as you would in software. There is no "brightness" wire that connects to a physical LED as far as I have ever seen. There's no "backlight" in an LED.

Sure, but that seems tangential to how it is implemented. UX is the primary concern.

> These are implementation details that should be left to each individual implementation. What we should concentrate on here is how we want to expose this functionality to the user. The individual LED implementation in InputPlumber is responsible for setting the correct sysfs values based on the color intensity values. Exposing the RGB values alone will afford the exact same array of LED color that you could with RGB + Brightness. Our DBus interface does not need to be identical to the kernel interface.

If you just want to assume max value, then it will work, but UX will be compromised. IMO it would be better (and consistent with existing userspace software which has brightness sliders) to provide a brightness value through the dbus interface. In the event an interface does _not_ have a brightness value, that implementation could do the calculation for setting RGB. This ensures consistency for userspace software, where special cases don't need to be made in the software to determine if the levels need to be scaled down or not before they are sent over.

The reason brightness is used as a trait is that it simplifies selecting a color, then setting its intensity. The software is slightly more complex with 4 sliders instead of 3, but it allows the user to more easily tweak the color they want. For example, finding teal with r 50, g 255, b 190, then lowering the brightness to 10% is much easier than finding teal with those same values, then determining 10% of that should be r 19, g 100, b 75 and manually moving them.

PR #344, opened 2025-04-25
feat(systemd): flag the service as dbus type

NeroReflex:
Make systemd aware of the service being a dbus service.

### DIFF
--- a/rootfs/usr/lib/systemd/system/inputplumber.service
+++ b/rootfs/usr/lib/systemd/system/inputplumber.service
@@ -2,6 +2,8 @@
 Description=InputPlumber Service
 
 [Service]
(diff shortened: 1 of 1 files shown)


pastaq (changes requested):
I briefly looked at the discord discussion. For the record here, can you edit the commit message to include the problem this is solving?

NeroReflex:
> I briefly looked at the discord discussion. For the record here, can you edit the commit message to include the problem this is solving?

This isn't solving any problem nor modifying the behaviour in amy way. It's just the systemd recommended way to run dbus services.

pastaq (approved):

ShadowApex (approved):

PR #345, opened 2025-04-25
fix(Composite Device): refactors target device behavior into its own struct

ShadowApex:
Currently the main composite device file is sitting at 2312 lines, encapsulating a _lot_ of functionality, making it difficult to work on and debug. This change refactors the `CompositeDevice` struct to pull out all target device state and handling into its own `CompositeDeviceTargets` struct.

In an upcoming PR, I also plan on doing the same for source device handling by creating a `CompositeDeviceSources` struct and moving the source device behavior there to further reduce the amount of state and behavior in `CompositeDevice`.

### DIFF
--- a/src/input/composite_device/mod.rs
+++ b/src/input/composite_device/mod.rs
@@ -1,5 +1,6 @@
 pub mod client;
 pub mod command;
+pub mod targets;
 
 use std::{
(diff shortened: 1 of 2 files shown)


pastaq (approved):
Good idea.

PR #346, opened 2025-04-27
feat(Legion Go S): Add Config Support

pastaq:
- Adds support for configuring the Legion Go S Gamepad and Touchpad devices. Depends on lenovo-legos-hid driver.

### DIFF
--- a/rootfs/usr/share/inputplumber/devices/50-legion_go_s.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-legion_go_s.yaml
@@ -23,6 +23,39 @@ matches:
 # One or more source devices to combine into a single virtual device. The events
 # from these devices will be watched and translated according to the key map.
 source_devices:
+  # Config, used for setting attributes on load
+  - group: gamepad
+    udev:
+      attributes:
+        - name: bInterfaceNumber
+          value: "03"
+        - name: idVendor
+          value: "1a86"
+        - name: idProduct
+          value: "e310"
+      driver: lenovo-legos-hid
+      subsystem: hidraw
+  - group: gamepad
+    udev:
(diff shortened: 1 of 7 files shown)


rootfs/usr/share/inputplumber/devices/50-legion_go_s.yaml
```diff
@@ -23,6 +23,17 @@ matches:
...
+  - group: gamepad
+    hidraw:
+      vendor_id: 0x1a86
+      product_id: 0xe310
+      interface_num: 3
+  - group: gamepad
+    hidraw:
+      vendor_id: 0x1a86
+      product_id: 0xe311
+      interface_num: 3
```
ShadowApex, line 36: Can we use udev matches here to also match on the driver? E.g.

```suggestion
  # Config, used for setting attributes on load
  - group: gamepad
    udev:
      attributes:
        - name: bInterfaceNumber
          value: "03"
        - name: idVendor
          value: "1a86"
        - name: idProduct
          value: "e310"
      driver: lenovo-legos-hid
      subsystem: hidraw
  - group: gamepad
    udev:
      attributes:
        - name: bInterfaceNumber
          value: "03"
        - name: idVendor
          value: "1a86"
        - name: idProduct
          value: "e311"
      driver: lenovo-legos-hid
      subsystem: hidraw
```

ShadowApex (changes requested):
One small request, but otherwise looks good!

rootfs/usr/share/inputplumber/devices/50-legion_go_s.yaml
pastaq: Works well. Thanks.

PR #348, opened 2025-04-29
fix(Steam Deck): Add GetChipId handling.

pastaq:
- A recent update to the Steam client causes any Steam Deck target device that isn't a Deck or Legion Go S to be treated like a Steam Controller. The Steam Controller (and now the Deck targets) receives a GetChipId (0xBA) Set/GetReport sequence. This message is spammed 500 times and takes approx 1-2 mins for all 500 requests to time out. This behavior can be seen in the following video: https://youtu.be/8oCHWmL6XzQ?t=137

  Responding with a correct ID resolves this issue.

- While at it, Correct the name of or add missing all the ReportType entries as found in the SDL constants for Valve controllers: https://github.com/libsdl-org/SDL/blob/510c7edd9b53c6e1b826a2a009de24fcc4dfb2c1/src/joystick/hidapi/steam/controller_constants.h


ShadowApex (approved):
Looks good!

PR #352, opened 2025-05-01
fix(MSI Claw A1M): send MKeys setting when switching to xinput mode

ShadowApex:
This change fixes an issue where sometimes the gamepad would fail to switch into xinput mode. Tested on two different devices. It will also read and print the current gamepad mode if debug logging is enabled.


pastaq (approved):
Not a huge fan of doing these types of workarounds outside the kernel. TBS I understand the claw driver is a WIP so I'll let it go ahead for now.

PR #353, opened 2025-05-02
feat(LedMultiColor): Refactor RGB interface

pastaq:
- Rename RGB Driver, Implementation, and Errors to be less opinionated about the specific details of the hardware.
- Add support for reading/writing brightness.
- Add support for reading multicolor intensity.
- Add support for multi_intensity that provides 'red green blue' vice only 'rgb'
- Remove requirement for udev device matching to require a devnode.
- Use sysname instead of name for leds as many led devices don't have a name.
- Add configuration for all known kernel drivers that support led_mc.


src/input/source/led/multicolor.rs
```diff
@@ -0,0 +1,272 @@
...
+                break; //TODO: Support more than one color for this type.
+            }
+            IndexType::Red => current_color.r = color_vals[pos] as u8,
+            IndexType::Green => current_color.g = color_vals[pos] as u8,
+            IndexType::Blue => current_color.b = color_vals[pos] as u8,
+        }
+    }
+
+    Ok(current_color)
+}
```
ShadowApex, line 229: With this design, right now reading and writing to LEDs are limited to only RGB values, when in reality multicolor LEDs can have an arbitrary array of colors. Some LEDs might only be able to support one color, two colors, or 5 colors. 

On my laptop, for example, there are 6 different colors in the color index:
```
/sys/class/leds/chromeos:multicolor:charging $ cat multi_index
red green blue yellow white amber
```

Instead of limiting the color to RGB values, we should instead support an arbitrary array of colors.

src/input/source/led/multicolor.rs
```diff
@@ -0,0 +1,272 @@
...
+                    Err(e.into())
+                }
+            }
+        }
+        Err(e) => {
+            log::error!("Failed to read_brightness");
+            Err(e.into())
+        }
+    }
+}
```
ShadowApex, line 250: Reduce indentation:

```suggestion
/// Reads the brightness path and returns a u8 with the current brightness setting
fn read_brightness(path: &Path) -> Result<u8, Box<dyn Error + Send + Sync>> {
    let mut brightstring = read_to_string(path)?;
    brightstring = brightstring.trim().to_string();
    log::debug!("Got brightness value: {brightstring}");
    Ok(brightstring.parse()?)
}
```

src/input/source/led/multicolor.rs
```diff
@@ -0,0 +1,272 @@
...
+                current_color.b = rgb[3];
+                break; //TODO: Support more than one color for this type.
+            }
+            IndexType::Red => current_color.r = color_vals[pos] as u8,
+            IndexType::Green => current_color.g = color_vals[pos] as u8,
+            IndexType::Blue => current_color.b = color_vals[pos] as u8,
+        }
+    }
+
+    Ok(current_color)
```
ShadowApex, line 228: Instead of returning a hard-coded RGB value, we should instead return a vector of color values (i.e. `Vec<u8>`)

src/input/source/led/multicolor.rs
```diff
@@ -0,0 +1,272 @@
...
+            result.current_color.r,
+            result.current_color.g,
+            result.current_color.b,
+        )?;
+
+        Ok(result)
+    }
+
+    /// Writes the given red, green, and blue values to the device.
+    fn write_color(&self, r: u8, g: u8, b: u8) -> Result<(), Box<dyn Error + Send + Sync>> {
```
ShadowApex, line 114: To support non-RGB multi-color LEDs, I think we should change this method to take a hashmap of values instead of hard-coding RGB values. E.g.

```suggestion
    fn write_color(&self, color: HashMap<String, u8>) -> Result<(), Box<dyn Error + Send + Sync>> {
```

Then write the values according to the `multi_index` mapping.

pastaq: This function is still useful for target interfaces that only support RGB writing. A more comprehensive write_colors function was added in v2.

src/input/source/led/multicolor.rs
```diff
@@ -0,0 +1,321 @@
...
+            "orange" => multi_index_map.push(ColorType::Orange),
+            "pink" => multi_index_map.push(ColorType::Pink),
+            "purple" => multi_index_map.push(ColorType::Purple),
+            "red" => multi_index_map.push(ColorType::Red),
+            "rgb" => multi_index_map.push(ColorType::Rgb),
+            "violet" => multi_index_map.push(ColorType::Violet),
+            "white" => multi_index_map.push(ColorType::White),
+            "yellow" => multi_index_map.push(ColorType::Yellow),
+            _ => return Err(Box::new(LedMcError::UnsupportedIndex(idx.clone()))),
+        }
```
ShadowApex, line 317: Instead of doing this here, we should implement `TryFrom<String>` for `ColorType`. Then you can just do: `multi_index_map.push(idx.try_into()?)`

E.g.
```rust
impl TryFrom<String> for ColorType {
    type Error = LedMcError;

    fn try_from(value: String) -> Result<Self, Self::Error> {
        match value.as_str() {
            "amber" => Ok(ColorType::Amber),
            "blue" => Ok(ColorType::Blue),
            "cyan" => Ok(ColorType::Cyan),
            "green" => Ok(ColorType::Green),
            "infrared" => Ok(ColorType::Infrared),
            "lime" => Ok(ColorType::Lime),
            "multi" => Ok(ColorType::Multi),
            "orange" => Ok(ColorType::Orange),
            "pink" => Ok(ColorType::Pink),
            "purple" => Ok(ColorType::Purple),
            "red" => Ok(ColorType::Red),
            "rgb" => Ok(ColorType::Rgb),
            "violet" => Ok(ColorType::Violet),
            "white" => Ok(ColorType::White),
            "yellow" => Ok(ColorType::Yellow),
            _ => Err(LedMcError::UnsupportedIndex(value)),
        }
    }
}
```

src/input/source/led/multicolor.rs
```diff
@@ -0,0 +1,321 @@
...
+
+/// Reads the brightness path and returns a u8 with the current brightness setting
+fn read_brightness(path: &Path) -> Result<u8, Box<dyn Error + Send + Sync>> {
+    let mut brightstring = read_to_string(path)?;
+    brightstring = brightstring.trim().to_string();
+    log::debug!("Got brightness value: {brightstring}");
+    Ok(brightstring.parse()?)
+}
+
+/// Reads the multi_index path and returns a Vec<IndexType> of the configured device.
```
ShadowApex, line 291:
```suggestion
/// Reads the multi_index path and returns a Vec<ColorType> of the configured device.
```

src/input/source/led/multicolor.rs
```diff
@@ -0,0 +1,321 @@
...
+            }
+            _ => {
+                return Ok(());
+            }
+        }
+        Ok(())
+    }
+}
+
+/// Reads the multi_intensity path and returns an LedFixedColor with the current color setting
```
ShadowApex, line 248:
```suggestion
/// Reads the multi_intensity path and returns a Vec<HashMap<ColorType, u8>> with the current color setting
```

src/input/source/led/multicolor.rs
```diff
@@ -0,0 +1,321 @@
...
+fn read_color(
+    path: &Path,
+    index: &[ColorType],
+) -> Result<Vec<HashMap<ColorType, u8>>, Box<dyn Error + Send + Sync>> {
+    let mut current_color: Vec<HashMap<ColorType, u8>> = Vec::new();
+
+    let color_contents = read_to_string(path).map_err(|err| Box::new(LedMcError::Read(err)))?;
```
ShadowApex, line 255: If we have defined error types, then we should use that in the result instead of a `Box<dyn Error>`:

```suggestion
fn read_color(path: &Path, index: &[ColorType]) -> Result<Vec<HashMap<ColorType, u8>>, LedMcError> {
    let mut current_color: Vec<HashMap<ColorType, u8>> = Vec::new();

    let color_contents = read_to_string(path).map_err(|err| LedMcError::Read(err))?;
```

src/input/source/led/multicolor.rs
```diff
@@ -0,0 +1,321 @@
...
+        read_brightness(self.brightness_path.as_path())
+    }
+
+    /// Writes the given Vec<u32> color mappiing to the device. The color mapping should be
+    /// created following the multi_index_map of the device. Ensure that Multi and Rgb ColorType
+    /// indexes are accounted for if they were translated to Red Green Blue ColorType before
+    /// sending to this function.
+    fn write_colors(&self, colors: Vec<u32>) -> Result<(), Box<dyn Error + Send + Sync>> {
+        // Loop through all the colors, Then add the values of each to a string.
+        let mut raw_values = Vec::new();
```
ShadowApex, line 168: Small optimization:

```suggestion
        let mut raw_values = Vec::with_capacity(colors.len());
```

src/input/source/led/multicolor.rs
```diff
@@ -0,0 +1,321 @@
...
+            self.multi_index_map.as_ref(),
+        )
+    }
+
+    /// Reads the brightness path and returns a u8 with the current brightness setting
+    fn read_brightness(&self) -> Result<u8, Box<dyn Error + Send + Sync>> {
+        read_brightness(self.brightness_path.as_path())
+    }
+
+    /// Writes the given Vec<u32> color mappiing to the device. The color mapping should be
```
ShadowApex, line 162:
```suggestion
    /// Writes the given Vec<u32> color mapping to the device. The color mapping should be
```

ShadowApex (changes requested):

ShadowApex (approved):
LGTM! Just need to rebase off main to fix the CI pipeline.

PR #357, opened 2025-05-07
fix(Deck): Add generic target PID

pastaq:
- Ensure only the Steam Deck is detected with the valid Steam Deck PID.
- Add generic PID so devices that are not a Deck can work with deck-uhid.

Tested on OrangePi NEO


ShadowApex (approved):

pastaq:
@ShadowApex since the deck PID will not work in UHID, should we use the generic PID in that one case?

ShadowApex:
> @ShadowApex since the deck PID will not work in UHID, should we use the generic PID in that one case?

Yes, that's true, we should. The deck PID will only work with the virtual usb deck target.

PR #359, opened 2025-05-10
feat(Autostart): Add autostart rules/hwdb for InputPlumber

pastaq:
- Modified from steamos-customizations, the new udev hwdb and udev rules allow for automatic startup of inputplumber on boot. This allows immutable systems to ship inputplumber and not need clever hacks or a blanket service enabling when not needed. This will also improve the level of support out of the box in SteamOS.


ShadowApex (changes requested):
The only consideration here is if someone has InputPlumber installed, but does _not_ want it to autostart when it detects one of these devices, though that seems like an unlikely case if someone installs InputPlumber.

We will also need to update the RPM spec file (used for COPR builds):
https://github.com/ShadowBlip/InputPlumber/blob/main/pkg/rpm/inputplumber.spec#L42

And the `Cargo.toml` which builds the RPM for GitHub releases:
https://github.com/ShadowBlip/InputPlumber/blob/main/Cargo.toml#L18

pastaq:
> The only consideration here is if someone has InputPlumber installed, but does _not_ want it to autostart when it detects one of these devices, though that seems like an unlikely case if someone installs InputPlumber.

In such a case they can mask the service.

ShadowApex (approved):

PR #361, opened 2025-05-12
fix(Makefile): use target triple from rustc for TARGET_ARCH

hack3ric:
Sometimes `uname -m` does not match target triple's arch (like riscv64, which is riscv64gc-*), sometimes we don't have `gnu` but `gnueabi` or `musl`. Use the [method](https://wiki.archlinux.org/title/Rust_package_guidelines#Prepare) used on Arch Linux to directly get target triple from rustc.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 NAME := $(shell grep 'name =' Cargo.toml | head -n 1 | cut -d'"' -f2)
 VERSION := $(shell grep '^version =' Cargo.toml | cut -d'"' -f2)
-ARCH ?= $(shell uname -m)
-TARGET_ARCH ?= $(ARCH)-unknown-linux-gnu
+TARGET_ARCH ?= $(shell rustc -vV | sed -n 's/host: //p')
+ARCH := $(shell echo "$(TARGET_ARCH)" | cut -d'-' -f1)
 DBUS_NAME := org.shadowblip.InputPlumber
 ALL_RS := $(shell find src -name '*.rs')
 ALL_ROOTFS := $(shell find rootfs -type f)
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -28,4 +28,4 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - run: |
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -1,7 +1,7 @@
 NAME := $(shell grep 'name =' Cargo.toml | head -n 1 | cut -d'"' -f2)
 VERSION := $(shell grep '^version =' Cargo.toml | cut -d'"' -f2)
 ARCH ?= $(shell uname -m)
-TARGET_ARCH ?= $(ARCH)-unknown-linux-gnu
+TARGET_ARCH ?= $(shell rustc -vV | sed -n 's/host: //p')
```
ShadowApex, line 4: This change may produce undesirable results if someone forgets to set both `ARCH` and `TARGET_ARCH` when trying to cross-compile using the Makefile. Ideally we should set the `ARCH` based on the user-provided `TARGET_ARCH` to prevent this issue:

```suggestion
TARGET_ARCH ?= $(shell rustc -vV | sed -n 's/host: //p')
ARCH := $(shell echo "$(TARGET_ARCH)" | cut -d'-' -f1)
```

ShadowApex (changes requested):
Thank you for this change! This will be a good improvement to the Makefile, but currently this change will break our CI which builds the x86_64 binary and cross-compiles the aarch64 binary. Could you also include an update in your PR that fixes the CI pipeline?

**.releaserc.yaml**

```yaml
      prepareCmd: |
        make in-docker TARGET='dist update-pkgbuild-hash'
        make in-docker TARGET='dist' TARGET_ARCH="aarch64-unknown-linux-gnu"
```

Thank you!

ShadowApex:
The test job will also have to be updated to accommodate this change:

**.github/workflows/test.yaml**

```yaml
  build-aarch64:
    name: Run aarch64 build
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v3
      - run: |
          make in-docker TARGET='dist' BUILD_TYPE='debug' TARGET_ARCH='aarch64-unknown-linux-gnu'
```

hack3ric:
Thanks for the quick review, I've integrated the changes :)

PR #364, opened 2025-05-13
fix(GPD): Support upstream gyro driver

justinweiss:
The upstream driver has a different device name and needs to be detected using the new name.

I added `bmi260` literally because `bmi*` seemed like it could be too generic, but I'm happy to change `bmi*-imu` to `bmi*` if you'd prefer.


ShadowApex (approved):
Looks good! If we end up finding more bmiXXX drivers in the wild, we can consider using `bmi*`, but `bmi260` seems fine to me.

Thanks for this change!

PR #371, opened 2025-06-03
feat(Event): add event context with performance metrics

ShadowApex:
This change adds a new `EventContext` property of the `NativeEvent` which can hold performance metrics and metadata related to an input event. Performance metrics can be used to show the processing latency for each event.

Performance/latency can be monitored with the CLI:

```bash
inputplumber device 0 target monitor xbox-series
```

Where "xbox-series" is the target device type to monitor.

image

### DIFF
--- a/src/input/source/mod.rs
+++ b/src/input/source/mod.rs
@@ -1,5 +1,6 @@
 use std::{
     collections::HashSet,
+    env,
     error::Error,
     str::FromStr,
     sync::{Arc, Mutex, MutexGuard},
@@ -22,7 +23,7 @@ use self::{
 use super::{
     capability::Capability,
     composite_device::client::CompositeDeviceClient,
-    event::{native::NativeEvent, Event},
+    event::{context::EventContext, native::NativeEvent, Event},
     output_event::OutputEvent,
 };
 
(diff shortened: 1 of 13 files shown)


src/input/source/mod.rs
```diff
@@ -348,6 +359,13 @@ impl<T: SourceInputDevice + SourceOutputDevice + Send + 'static> SourceDriver<T>
                         {
                             continue;
                         }
+                        let mut context = context.clone();
+                        let send_span = context
+                            .metrics_mut()
+                            .create_child_span("root", "source_send");
+                        send_span.start();
+                        // TODO: Only allocate the event context if setting enabled
```
pastaq: What are the implications of this not being done?

pastaq (commented):
I'll test it soon. 

Is it planned at all to generate some sort of data log csv or some other file to be able to reproduce a run and compare graphs?

src/input/source/mod.rs
ShadowApex: There is some small extra overhead allocating these bytes on the heap for every event. Overall the effect is very small, but with many events per second, this effect could add up, so it would be slightly better for performance if we only perform these heap allocations when metrics are enabled.

ShadowApex:
> Is it planned at all to generate some sort of data log csv or some other file to be able to reproduce a run and compare graphs?

That's certainly an option. Right now I was planning on just comparing different things by comparing the graphs from the CLI (`inputplumber device <num> target monitor <target dev>`).

PR #377, opened 2025-06-09
chore(Makefile): cache crates when running in docker

ShadowApex:
This change updates the `in-docker` target in the Makefile to re-use the cargo cache between runs. This ~~increases~~* decreases the build times dramatically for incremental builds with Docker.


pastaq (approved):
Increases?

PR #379, opened 2025-06-13
fix(FlyDigi Vader 4 Pro): Finish gyro and improve poll rate

winghugs:
this add gyro support for the vader 4 pro

the polling rate has also been adjusted as several testers reported a better feel with a faster poll rate.


src/input/source/hidraw.rs
```diff
@@ -423,7 +423,7 @@ impl HidRawDevice {
             DriverType::Vader4Pro => {
                 let device = Vader4Pro::new(device_info.clone())?;
                 let options = SourceDriverOptions {
-                    poll_rate: Duration::from_millis(1),
```
pastaq, line 426: What is the CPU load at 0ms poll rate? This will run as fast as possible, code execution is the only limiting factor. 1ms is already 1000Hz. Have you tried 750/500us?

pastaq (changes requested):
Fan you change the commit title to match our semantic versioning CI?
E.g.
`fix(FlyDigi Vader 4 Pro): Finish gyro and improve poll rate`

You can also squash these commits

src/input/source/hidraw.rs
winghugs: I haven't noticed any significant CPU load as a result of this, and it is perceivable between 1ms and 0ms (for some reason.)

I'm happy to set it back to 1 if the 0 here makes you nervous, it still feels okay at 1. Maybe this setting can be exposed to the user at some point in the future.

ShadowApex: With the current implementation, this will do a blocking read and wait up to `HID_TIMEOUT` milliseconds (10ms currently) for data from the device. Since it does a blocking read and not a non-blocking one, this won't burn extra CPU cycles, but it will run at the speed of the device; which seems OK to me. 

If we ever update this to use a non-blocking read, we will definitely need to change this, though.

ShadowApex:
Thanks for this @winghugs !

PR #381, opened 2025-06-15
feat(ds5-bt): allow emulating a bluetooth dualsense

NeroReflex:
Allow emulation of a bluetooth DualSense or DualSense Edge.


ShadowApex (changes requested):
These changes require adding an evdev check in `manager.rs` like we do with hidraw to ensure we don't add virtual evdev devices that are not real bluetooth devices.

PR #384, opened 2025-06-19
docs(README): add troubleshooting guide

LukeShortCloud:
This includes the most common troubleshooting steps for InputPlumber along with real-world examples from my original ASUS ROG Ally.

I also added a table of contents to make it a bit easier to navigate the README.md file.

Let me know if I should add or change anything!

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,19 @@
   <br>
 </p>
 
+Table of Contents:
+- [About](#about)
+    - [Features](#features)
+- [Install](#install)
+- [Documentation](#documentation)
+- [Usage](#usage)
+    - [Input Profiles](#input-profiles)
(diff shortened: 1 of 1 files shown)


pastaq:
Looks good. I don't have any issues with it as is, but I think some "context" would be useful for each. The "how" is covered, but the "why" is not unless you're already familiar with the tools.

pastaq (approved):

LukeShortCloud:
Sorry, I've been so busy with Playtron. I thought that after the GameOS 1.0 release, things would calm down but I got busier more than ever before lol.

Starting soon, I'll have more free time again. I hope to get back to addressing your feedback in the near future.

PR #389, opened 2025-07-04
xpad uhid d-pad fix

alkazar:
I noticed that the d-pad on my xbox series x controller was not working correctly when connected via bluetooth with xpad.

There was a typo/oopsie in the state transition code mixing up a direction.


alkazar:
Well, xpad is the default... and is simpler. It seems to work well enough.

pastaq:
> Well, xpad is the default... and is simpler. It seems to work well enough.

Xpad doesn't support Bluetooth

PR #390, opened 2025-07-10
feat(ForceFeedback): add dbus interface for rumble

ShadowApex:
This change adds a new `org.shadowblip.Output.ForceFeedback` dbus interface which can be used as a simple interface for sending rumble output events over dbus. To facilitate support for this, this change also fleshes out the design for advertising and fetching output capabilities for devices. It also includes the ability for the composite device to notify target devices if any input/output capabilities have changed.

If a device supports force feedback, you can do the following to make it rumble:

```bash
busctl call org.shadowblip.InputPlumber \
  /org/shadowblip/InputPlumber/devices/target/dbus0 \
  org.shadowblip.Output.ForceFeedback \
  Rumble "d" 1.0
```

During testing of this change, I ran into a deadlock issue, which caused the device to stop responding, so I have also included a separate commit that adds timeouts for any messages that get sent through a client channel in order to help better detect deadlocks in the future.


src/dbus/interface/force_feedback.rs
```diff
@@ -0,0 +1,94 @@
...
+use zbus_macros::interface;
+
+use crate::{
+    drivers::steam_deck::hid_report::PackedRumbleReport,
+    input::{composite_device::client::CompositeDeviceClient, output_event::OutputEvent},
+};
+
+use super::NamedInterface;
+
+/// [ForceFeedbacker] is any device that can implement force feedback
```
pastaq, line 12: TIHI

src/dbus/interface/force_feedback.rs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+impl ForceFeedbacker for CompositeDeviceClient {
+    async fn rumble(&mut self, value: f64) -> Result<(), Box<dyn Error>> {
+        let value = value.min(1.0);
+        let value = value.max(0.0);
+        log::debug!("Sending rumble event with value: {value}");
+        let report = PackedRumbleReport {
+            intensity: (value * u8::MAX as f64) as u8,
+            left_speed: Integer::from_primitive((value * u16::MAX as f64) as u16),
```
pastaq, line 25: This should be strong/weak magnitude. Steam deck does this in a non standard way and some evdev based devices implement it this way, but it's not a guarantee and is opinionated against standard.

src/dbus/interface/mod.rs
```diff
@@ -1,5 +1,12 @@
 pub mod composite_device;
+pub mod force_feedback;
 pub mod manager;
 pub mod performance;
 pub mod source;
 pub mod target;
+
+/// Used to return the name of a DBus interface
```
pastaq: This comment is missing on some of the other uses

src/dbus/interface/force_feedback.rs
ShadowApex: Thanks, I hate it? :laughing:

src/dbus/interface/mod.rs
ShadowApex: I will actually be removing this because I discovered that every zbus `Interface` has a `name()` method that does this exact thing.

src/dbus/interface/force_feedback.rs
ShadowApex: Not sure what you're asking here. You want this function to take two values; one for strong and weak magnitude?

pastaq: I think that is strange we're using the steam input rumble event as our "defacto" rumble when it's not a standard hid spec output event. This seems very opinionated towards the steam deck and steam input rather than being based on the hid spec. It probably doesn't matter in the long run technically because we translate between both of them which is why I didn't ask for changes, I just wanted to understand the reasoning.

ShadowApex: Yes, I chose to use the Deck output event because it was simple to implement and will ultimately be translated depending on the source device. If I chose to send evdev output events, I would need to add extra logic to include the effect upload dance, whereas the Deck and DS5 output events handle this already.

pastaq (approved):

DocMAX:
Absolutely love seeing your active development. Keep up the great work.

PR #392, opened 2025-07-12
fix: remove quick access capability from source dualsense

imLinguin:
the dualsense gamepad doesn't have such key and this only causes problems for apps that rely on checking for this capability

unsure how the key-combo use case of that capability behaves now (as I have not tested that), I believe however the target driver should work as expected

### DIFF
--- a/src/input/source/hidraw/dualsense.rs
+++ b/src/input/source/hidraw/dualsense.rs
@@ -469,7 +469,6 @@ pub const CAPABILITIES: &[Capability] = &[
     Capability::Gamepad(Gamepad::Button(GamepadButton::LeftStickTouch)),
     Capability::Gamepad(Gamepad::Button(GamepadButton::LeftTrigger)),
     Capability::Gamepad(Gamepad::Button(GamepadButton::North)),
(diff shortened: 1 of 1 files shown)


src/input/source/hidraw/dualsense.rs
```diff
@@ -469,7 +469,6 @@ pub const CAPABILITIES: &[Capability] = &[
     Capability::Gamepad(Gamepad::Button(GamepadButton::LeftStickTouch)),
     Capability::Gamepad(Gamepad::Button(GamepadButton::LeftTrigger)),
     Capability::Gamepad(Gamepad::Button(GamepadButton::North)),
-    Capability::Gamepad(Gamepad::Button(GamepadButton::QuickAccess)),
```
pastaq, line 472: This looks like it was supposed to be
`    Capability::Gamepad(Gamepad::Button(GamepadButton::Mute)),` instead

pastaq (changes requested):

src/input/source/hidraw/dualsense.rs
imLinguin: oh interesting, I missed that

PR #394, opened 2025-07-13
fix(DBus): add interface manager

ShadowApex:
This change adds a new `DBusInterfaceManager` which is responsible for starting and stopping dbus interfaces for a particular object/struct. It ensures that dbus interfaces are removed for a particular path if the `DBusInterfaceManager` is ever freed.


DocMAX:
Looks like we have an issue with this... Doesn't detect Gamepads/Inputs anymore... I see this in log:
```
Jul 16 22:47:43 steamdeck inputplumber[738]: [2025-07-16T20:47:43Z ERROR inputplumber::input::composite_device] Failed running device: "InputError occurred running source device: Failed to fetch events: Os { code: 19, kind: Uncategorized, message: \"No such device\" }"
Jul 16 22:47:43 steamdeck inputplumber[738]: [2025-07-16T20:47:43Z ERROR inputplumber::input::composite_device] Failed to send source devices changed signal: FDO(Failed("device no longer exists"))

```

pastaq:
May be related to #401

PR #405, opened 2025-07-19
fix(InputManager): Fix duplicate target creation.

pastaq:
- InputManager does not use the same device queue process when starting a composite device that we use for setting the target devices after it is started. If an external process connects to the dbus interface and uses set_target_devices before the InputManager has finished starting all targets it can pass all checks against de-duplication and start multiple conflicting targets. As a remedy, instead of doing a custom one off method at init, grab the ComposteDeviceClient and set the target devices the same way.

Tested on Legion Go S and Ally X, SteamOS.

### DIFF
--- a/src/input/manager.rs
+++ b/src/input/manager.rs
@@ -813,13 +813,23 @@ impl Manager {
             .unwrap();
         sources.push(source_device);
 
-        device.listen_on_dbus().await?;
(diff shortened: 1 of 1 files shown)


ShadowApex (approved):
Great change, looks good!

PR #406, opened 2025-07-20
fix(MSI Claw A8):Add support for MSI Claw A8.

honjow:
Also need update udev hwdb

### DIFF
--- a/rootfs/usr/share/inputplumber/devices/50-msi_claw_a8_bz2e.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-msi_claw_a8_bz2e.yaml
@@ -1,0 +1,60 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ShadowBlip/InputPlumber/main/rootfs/usr/share/inputplumber/schema/composite_device_v1.json
+# Schema version number
+version: 1
+
+# The type of configuration schema
+kind: CompositeDevice
+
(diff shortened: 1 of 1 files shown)


ShadowApex:
Thanks for this @honjow! We just need to correct the following error from our tests and we should be able to merge this in:

```
Errors:
  * Unable to find pattern 'dmi:*svnMicro-StarInternationalCo.,Ltd.:*pnClawA8BZ2EM:*' generated from config "./rootfs/usr/share/inputplumber/devices/50-msi_claw_a8_bz2e.yaml" in hwdb file: ./rootfs/usr/lib/udev/hwdb.d/60-inputplumber-autostart.hwdb
Total errors: 1

Configs with failures:
  "./rootfs/usr/share/inputplumber/devices/50-msi_claw_a8_bz2e.yaml"

ERROR: The above device configurations have `auto_manage: true`, but do not have a matching entry in the `inputplumber-autostart.hwdb` file. Please add an entry to the hwdb file so the inputplumber service will start when the device is detected.
```

ShadowApex (changes requested):
You should be able to add the pattern `dmi:*svnMicro-StarInternationalCo.,Ltd.:*pnClawA8BZ2EM:*` to "inputplumber-autostart.hwdb"

ShadowApex (approved):
Thank you @honjow !

pastaq (approved):

PR #410, opened 2025-07-26
Fix(Hardware Support): Passthrough Zotac Zone Touchpads

pastaq:
Users on SteamOS are reporting that touchpads do not work on the Zotac Zone. Steamos-manager currently only creates a deck-uhid target so the rel events produced have no target device to emit from. Passthrough the events to reintroduce this functionality. This may need to be changed in the future if touchpad firmware is updated to include an abs event mode, though it would likely have a different report descriptor and name at that point.

### DIFF
--- a/rootfs/usr/share/inputplumber/devices/50-zotac-zone.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-zotac-zone.yaml
@@ -41,12 +41,13 @@ source_devices:
       product_id: "1590"
       handler: event*
     capability_map_id: zone1
-  - group: mouse
-    evdev:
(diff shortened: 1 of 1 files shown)


ShadowApex:
This change, like #411, would make inputplumber read events from the touchpad device, but not grab or hide it, which can result in double input. Since we currently don't have source implementations for evdev mouse/touchpad devices right now, this may not immediately cause problems, but it will in the future if/when we decide to add source implementations for evdev mice/touchpads.

If we don't want inputplumber to manage the touchpad, then we should remove or comment out the source device block in the config instead of enabling passthrough.

ShadowApex (approved):

PR #411, opened 2025-07-26
Fix(Hardware Support): Passthrough Ayn Volume Events

pastaq:
Users on SteamOS are reporting that volume keys do not work on Ayn devices. Steamos-manager currently only creates a deck-uhid target so the keyboard events produced have no target device to emit from. Passthrough the events to reintroduce this functionality.

### DIFF
--- a/rootfs/usr/share/inputplumber/devices/50-ayn_loki_max.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-ayn_loki_max.yaml
@@ -28,6 +28,7 @@ source_devices:
       phys_path: usb-0000:74:00.0-1/input0
       handler: event*
   - group: keyboard
+    passthrough: true
     evdev:
       name: AT Translated Set 2 keyboard
       phys_path: isa0060/serio0/input0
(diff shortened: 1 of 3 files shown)


ShadowApex:
Setting `passthrough: true` would cause double input if the user uses a keyboard target device. This would break that use-case.

ShadowApex:
> Have you actually tested this or is this conjecture? Since both "down" and "up" events will happen at the same time there shouldn't be a repeated key as the kernel hasn't been signaled the key was released. I don't think this will happen.

This is not "conjecture". With passthrough enabled, inputplumber will not hide or grab the source device, so there will be two different keyboard devices both emitting the same keyboard events. How these inputs are used is up to the userspace application and may or may not be problematic depending on how they decide to handle volume input from two different keyboards at the same time.

Using passthrough in this case _also_ won't consume the key combo(s), which will bleed through to the focused application. i.e. apps will see CTRL+ALT+SHIFT+T

The proper solution to this issue is to update SteamOS manager to also set a target keyboard device.

Issue #202, opened 2024-09-14
Feature: Support running InputPlumber unprivileged


R1kaB3rN:
### Description

I'm interested in making InputPlumber available for clients through umu-launcher to fill in the gap of users unable to have their game controllers work, with the goal of the service being started by the client or umu-launcher at game launch in an unprivileged context.

### Problem

InputPlumber requires to be started and enabled as a privileged user to fulfill all of its tasks properly, making it both inaccessible for Flatpak applications and inconvenient as it would require clients to prompt their users for elevated privileges in the non-Flatpak case. Attempting to execute the `inputplumber` binary unprivileged results in:
```

[2024-09-14T19:03:19Z INFO  inputplumber] Starting InputPlumber v0.35.2
[2024-09-14T19:03:19Z ERROR inputplumber::input::manager] Failed to run task to list device configs: JoinError::Cancelled(Id(25))
[2024-09-14T19:03:19Z ERROR inputplumber::input::manager] Failed to run task to list device configs: JoinError::Cancelled(Id(27))
[2024-09-14T19:03:19Z ERROR inputplumber::input::manager] Failed to run task to list device configs: JoinError::Cancelled(Id(29))
[2024-09-14T19:03:19Z ERROR inputplumber::input::manager] Failed to run task to list device configs: JoinError::Cancelled(Id(31))
[2024-09-14T19:03:19Z ERROR inputplumber::input::manager] Failed to run task to list device configs: JoinError::Cancelled(Id(33))
[2024-09-14T19:03:19Z ERROR inputplumber::input::manager] Failed to run task to list device configs: JoinError::Cancelled(Id(35))
[2024-09-14T19:03:19Z ERROR inputplumber::input::manager] Failed to run task to list device configs: JoinError::Cancelled(Id(37))
[2024-09-14T19:03:19Z ERROR inputplumber::input::manager] Failed to run task to list device configs: JoinError::Cancelled(Id(39))
[2024-09-14T19:03:19Z ERROR inputplumber::input::manager] Failed to run task to list device configs: JoinError::Cancelled(Id(41))
[2024-09-14T19:03:19Z ERROR inputplumber::input::manager] Failed to run task to list device configs: JoinError::Cancelled(Id(43))
[2024-09-14T19:03:19Z ERROR inputplumber::input::manager] Failed to run task to list device configs: JoinError::Cancelled(Id(45))
[...]

PR #413, opened 2025-07-28
feat(security): polkit support

Pandapip1:
Addresses the use case described in https://github.com/ShadowBlip/InputPlumber/issues/202 (even if inputplumber itself still has to be run as root for the most part)

Stuff TODO:

- [x] Add missing `dbus-xml` files/entries
  - [x] `org.shadowblip.Input.Source.IIOIMUDevice`
  - [x] `org.shadowblip.Input.Source.LEDDevice`
  - [x] `org.shadowblip.Input.Source.UdevDevice`
  - [x] `org.shadowblip.Input.Debug`
  - [x] `org.shadowblip.Input.Touchscreen`
  - [x] `org.shadowblip.Output.ForceFeedback`
  - [x] `org.shadowblip.Input.Metrics`
  - [x] The ones I did before opening this PR
- [x] Add missing actions to the polkit actions file
  - [x] `org.shadowblip.Input.Source.IIOIMUDevice`
  - [x] `org.shadowblip.Input.Source.LEDDevice`
  - [x] `org.shadowblip.Input.Source.UdevDevice`
  - [x] `org.shadowblip.Input.Debug`
  - [x] `org.shadowblip.Input.Touchscreen`
  - [x] `org.shadowblip.Output.ForceFeedback`
  - [x] `org.shadowblip.Input.Metrics`
  - [x] The ones I did before opening this PR
- [x] Add missing polkit checks
  - [x] `org.shadowblip.Input.Source.IIOIMUDevice`
  - [x] `org.shadowblip.Input.Source.LEDDevice`
  - [x] `org.shadowblip.Input.Source.UdevDevice`
  - [x] `org.shadowblip.Input.Debug`
  - [x] `org.shadowblip.Input.Touchscreen`
  - [x] `org.shadowblip.Output.ForceFeedback`
  - [x] `org.shadowblip.Input.Metrics`
  - [x] `org.shadowblip.Input.Source.EventDevice`
  - [x] `org.shadowblip.Input.Source.HIDRawDevice`
  - [x] `org.shadowblip.Input.Mouse`
  - [x] `org.shadowblip.Input.Manager`
  - [x] `org.shadowblip.Input.Keyboard`
  - [x] `org.shadowblip.Input.Gamepad`
  - [x] `org.shadowblip.Input.DBusDevice`
  - [x] `org.shadowblip.Input.CompositeDevice`
- [x] Fix commit messages

### DIFF
--- a/pkg/rpm/inputplumber.spec
+++ b/pkg/rpm/inputplumber.spec
@@ -10,7 +10,7 @@ URL:            https://github.com/ShadowBlip/InputPlumber
 
 BuildRequires:  libevdev-devel libiio-devel git make cargo libudev-devel llvm-devel clang-devel
 Requires:       libevdev libiio
-Recommends:     steam gamescope-session linuxconsoletools
+Recommends:     steam gamescope-session linuxconsoletools polkit
 Provides:       inputplumber
 Conflicts:      hhd
 
@@ -36,6 +36,7 @@ mkdir -p %{buildroot}/usr/share/inputplumber/capability_maps
 mkdir -p %{buildroot}/usr/share/inputplumber/devices
 mkdir -p %{buildroot}/usr/share/inputplumber/profiles
 mkdir -p %{buildroot}/usr/share/inputplumber/schema
+mkdir -p %{buildroot}/usr/share/polkit-1/actions
 
 install -D -m 755 %{_builddir}/InputPlumber/target/%{_arch}-unknown-linux-gnu/release/inputplumber %{buildroot}/usr/bin/inputplumber
 install -D -m 644 %{_builddir}/InputPlumber/rootfs/usr/share/dbus-1/system.d/org.shadowblip.InputPlumber.conf %{buildroot}/usr/share/dbus-1/system.d/org.shadowblip.InputPlumber.conf
@@ -47,6 +48,7 @@ install -D -m 644 %{_builddir}/InputPlumber/rootfs/usr/share/inputplumber/capabi
 install -D -m 644 %{_builddir}/InputPlumber/rootfs/usr/share/inputplumber/devices/* %{buildroot}/usr/share/inputplumber/devices/
 install -D -m 644 %{_builddir}/InputPlumber/rootfs/usr/share/inputplumber/profiles/* %{buildroot}/usr/share/inputplumber/profiles/
 install -D -m 644 %{_builddir}/InputPlumber/rootfs/usr/share/inputplumber/schema/* %{buildroot}/usr/share/inputplumber/schema/
+install -D -m 644 %{_builddir}/InputPlumber/rootfs/usr/share/polkit-1/actions/org.shadowblip.InputPlumber.policy %{buildroot}/usr/share/polkit-1/actions/org.shadowblip.InputPlumber.policy
 
 %post
 udevadm control --reload-rules
@@ -71,6 +73,7 @@ systemctl disable inputplumber.service
 /usr/share/inputplumber/devices/*.yaml
 /usr/share/inputplumber/profiles/*.yaml
 /usr/share/inputplumber/schema/*.json
+/usr/share/polkit-1/actions/org.shadowblip.InputPlumber.policy
 
 %changelog
 * Tue Aug 6 2024 William Edwards [0.33.1-0]
--- a/rootfs/usr/share/dbus-1/system.d/org.shadowblip.InputPlumber.conf
+++ b/rootfs/usr/share/dbus-1/system.d/org.shadowblip.InputPlumber.conf
@@ -4,10 +4,15 @@
 <busconfig>
   <!-- Only root can own the service -->
   <policy user="root">
+    <allow receive_sender="org.shadowblip.InputPlumber"/>
     <allow own="org.shadowblip.InputPlumber"/>
   </policy>
-  <!-- Anyone can send messages to the owner of org.shadowblip.InputPlumber -->
+  <!--
+   Anyone can send messages to the owner of org.shadowblip.InputPlumber,
+   but deny receiving signals by default as it leaks timing information
+   -->
   <policy context="default">
+    <deny receive_sender="org.shadowblip.InputPlumber"/>
     <allow send_destination="org.shadowblip.InputPlumber"/>
   </policy>
 </busconfig>
--- a/rootfs/usr/share/dbus-1/system.d/org.shadowblip.InputPlumber.conf
+++ b/rootfs/usr/share/dbus-1/system.d/org.shadowblip.InputPlumber.conf
@@ -12,7 +12,9 @@
    but deny receiving signals by default as it leaks timing information
    -->
   <policy context="default">
-    <deny receive_sender="org.shadowblip.InputPlumber"/>
+    <!-- For the time being, allow receiving signals anyway for backwards compatibilty. -->
+    <!-- TODO: Uncomment this when an alternative API is made available -->
+    <!-- <deny receive_sender="org.shadowblip.InputPlumber"/> -->
     <allow send_destination="org.shadowblip.InputPlumber"/>
   </policy>
 </busconfig>
--- a/src/dbus/polkit.rs
+++ b/src/dbus/polkit.rs
@@ -1,11 +1,9 @@
-use std::collections::HashMap;
-use zbus::zvariant::{OwnedValue, Value};
-use zbus::{fdo, message::Header, Connection, Proxy};
+use zbus::{fdo, message::Header, Connection};
 
 pub async fn check_polkit(hdr: Option<Header<'_>>, action_id: &str) -> fdo::Result<()> {
     let connection = Connection::system()
         .await
-        .map_err(|e| fdo::Error::Failed(format!("Failed to acquire dbus connection: {}", e)))?;
+        .map_err(|e| fdo::Error::Failed(format!("Failed to acquire dbus connection: {e}")))?;
     check_polkit_with_connection(hdr, action_id, connection).await
 }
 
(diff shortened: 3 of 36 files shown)


Pandapip1:
Well this *should* be working now. I'll do some testing.

Pandapip1:
Note also that I've disabled receiving to signals for unprivileged apps, since it can't be controlled by polkit. I think the most appropriate way to do this with unprivileged apps is probably a new interface that creates a One-to-One D-Bus connection (which is a thing that apparently exists) over a pair of sockets created with `socketpair`, where InputPlumber keeps one of the sockets and sends the FD to the program that requested to receive inputs, and then otherwise does signals normally.

pkg/rpm/inputplumber.spec
```diff
@@ -47,6 +48,7 @@ install -D -m 644 %{_builddir}/InputPlumber/rootfs/usr/share/inputplumber/capabi
 install -D -m 644 %{_builddir}/InputPlumber/rootfs/usr/share/inputplumber/devices/* %{buildroot}/usr/share/inputplumber/devices/
 install -D -m 644 %{_builddir}/InputPlumber/rootfs/usr/share/inputplumber/profiles/* %{buildroot}/usr/share/inputplumber/profiles/
 install -D -m 644 %{_builddir}/InputPlumber/rootfs/usr/share/inputplumber/schema/* %{buildroot}/usr/share/inputplumber/schema/
+install -D -m 644 %{_builddir}/InputPlumber/rootfs/usr/share/polkit-1/actions/org.shadowblip.Input.policy %{buildroot}/usr/share/polkit-1/actions/org.shadowblip.Input.policy
```
ShadowApex:
```suggestion
install -D -m 644 %{_builddir}/InputPlumber/rootfs/usr/share/polkit-1/actions/org.shadowblip.InputPlumber.policy %{buildroot}/usr/share/polkit-1/actions/org.shadowblip.InputPlumber.policy
```

pkg/rpm/inputplumber.spec
```diff
@@ -71,6 +73,7 @@ systemctl disable inputplumber.service
 /usr/share/inputplumber/devices/*.yaml
 /usr/share/inputplumber/profiles/*.yaml
 /usr/share/inputplumber/schema/*.json
+/usr/share/polkit-1/actions/org.shadowblip.Input.policy
```
ShadowApex:
```suggestion
/usr/share/polkit-1/actions/org.shadowblip.InputPlumber.policy
```

src/dbus/polkit.rs
```diff
@@ -0,0 +1,83 @@
...
+        ))
+    })?;
+
+    let sender = hdr
+        .sender()
+        .ok_or_else(|| fdo::Error::Failed("Missing sender".into()))?;
+
+    let connection = Connection::system()
+        .await
+        .map_err(|e| fdo::Error::Failed(format!("Failed to acquire dbus connection: {}", e)))?;
```
ShadowApex: Instead of creating a new dbus connection for every request, I think it might be better to pass the connection as an argument to allow the caller to decide to re-use an existing connection or create a new one.

ShadowApex (changes requested):
Overall great changes. Thanks again for this! I still need to think about our strategy for merging this in in a way that won't break those that depend on us (i.e. SteamOS, OpenGamepadUI, etc.), but there were a few minor issues below.

We also need the following changes to ensure the new polkit policy is included in any packages we build:

```diff
diff --git a/Cargo.toml b/Cargo.toml
index 1729dc3..66e3c9e 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ authors = [
 assets = [
   { source = "target/release/inputplumber", dest = "/usr/bin/inputplumber", mode = "755" },
   { source = "rootfs/usr/share/dbus-1/system.d/org.shadowblip.InputPlumber.conf", dest = "/usr/share/dbus-1/system.d/org.shadowblip.InputPlumber.conf", mode = "644" },
+  { source = "rootfs/usr/share/polkit-1/actions/org.shadowblip.InputPlumber.policy", dest = "/usr/share/polkit-1/actions/org.shadowblip.InputPlumber.policy", mode = "644" },
   { source = "rootfs/usr/lib/systemd/system/inputplumber.service", dest = "/usr/lib/systemd/system/inputplumber.service", mode = "644" },
   { source = "rootfs/usr/lib/systemd/system/inputplumber-suspend.service", dest = "/usr/lib/systemd/system/inputplumber-suspend.service", mode = "644" },
   { source = "rootfs/usr/share/inputplumber/devices/*.yaml", dest = "/usr/share/inputplumber/devices/", mode = "644" },
@@ -54,6 +55,11 @@ assets = [
     "usr/share/dbus-1/system.d/org.shadowblip.InputPlumber.conf",
     "644",
   ],
+  [
+    "rootfs/usr/share/polkit-1/actions/org.shadowblip.InputPlumber.policy",
+    "usr/share/polkit-1/actions/org.shadowblip.InputPlumber.policy",
+    "644",
+  ],
   [
     "rootfs/usr/lib/systemd/system/*",
     "usr/lib/systemd/system/",
```

```diff
diff --git a/Makefile b/Makefile
index 9a04956..5dd424c 100644
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,8 @@ install: build ## Install inputplumber to the given prefix (default: PREFIX=/usr
                rootfs/usr/lib/udev/hwdb.d/*
        install -D -m 644 -t $(PREFIX)/lib/udev/rules.d/ \
                rootfs/usr/lib/udev/rules.d/*
+       install -D -m 644 -t $(PREFIX)/share/polkit-1/actions/ \
+               rootfs/usr/share/polkit-1/actions/*
        install -D -m 644 -t $(PREFIX)/share/$(NAME)/devices/ \
                rootfs/usr/share/$(NAME)/devices/*
        install -D -m 644 -t $(PREFIX)/share/$(NAME)/schema/ \
@@ -77,6 +79,7 @@ install: build ## Install inputplumber to the given prefix (default: PREFIX=/usr
 uninstall: ## Uninstall inputplumber
        rm $(PREFIX)/bin/$(NAME)
        rm $(PREFIX)/share/dbus-1/system.d/$(DBUS_NAME).conf
+       rm $(PREFIX)/share/polkit-1/actions/$(DBUS_NAME).policy
        rm $(PREFIX)/lib/systemd/system/$(NAME).service
        rm $(PREFIX)/lib/systemd/system/$(NAME)-suspend.service
        rm $(PREFIX)/lib/udev/hwdb.d/59-inputplumber.hwdb
```

src/dbus/polkit.rs
Pandapip1: This is fair. I'll implement this as a separate function that takes a connection and keep the existing function (with its existing function signature) so I don't have to do a massive refactor.

rootfs/usr/share/dbus-1/system.d/org.shadowblip.InputPlumber.conf
```diff
@@ -4,10 +4,15 @@
...
+    <allow receive_sender="org.shadowblip.InputPlumber"/>
     <allow own="org.shadowblip.InputPlumber"/>
   </policy>
-  <!-- Anyone can send messages to the owner of org.shadowblip.InputPlumber -->
+  <!--
+   Anyone can send messages to the owner of org.shadowblip.InputPlumber,
+   but deny receiving signals by default as it leaks timing information
+   -->
   <policy context="default">
+    <deny receive_sender="org.shadowblip.InputPlumber"/>
```
ShadowApex: Unfortunately listening to signals is required in order to not break compatibility with existing software (e.g. OpenGamepadUI). We'll need to allow receiving signals by default until there's a better compatibility path forward:

```suggestion
  <!--
   Anyone can send messages to the owner of org.shadowblip.InputPlumber,
   but deny receiving signals by default as it leaks timing information
   -->
  <policy context="default">
    <!-- <deny receive_sender="org.shadowblip.InputPlumber"/> -->
```

src/dbus/polkit.rs
```diff
@@ -0,0 +1,90 @@
+use std::collections::HashMap;
+use zbus::zvariant::{OwnedValue, Value};
+use zbus::{fdo, message::Header, Connection, Proxy};
+
+pub async fn check_polkit(hdr: Option<Header<'_>>, action_id: &str) -> fdo::Result<()> {
+    let connection = Connection::system()
+        .await
+        .map_err(|e| fdo::Error::Failed(format!("Failed to acquire dbus connection: {}", e)))?;
+    check_polkit_with_connection(hdr, action_id, connection).await
+}
```
ShadowApex: We'll need an escape hatch to ensure we don't break compatibility with existing clients (e.g. SteamOS, OpenGamepadUI, etc.). I think using an environment variable to enable polkit auth (that defaults to "off") would be a good way for us to pull this feature in and open a path forward for more secure defaults once these clients adopt polkit auth:

```suggestion
pub async fn check_polkit(hdr: Option<Header<'_>>, action_id: &str) -> fdo::Result<()> {
    let polkit_enabled = match std::env::var("ENABLE_POLKIT_AUTH") {
        Ok(value) => value.as_str() == "1",
        Err(_) => false,
    };
    if !polkit_enabled {
        return Ok(());
    }

    let connection = Connection::system()
        .await
        .map_err(|e| fdo::Error::Failed(format!("Failed to acquire dbus connection: {}", e)))?;
    check_polkit_with_connection(hdr, action_id, connection).await
}
```

ShadowApex (changes requested):
I did some testing and it works well! Thanks again for these changes! 

The only thing we need to do before we can pull these changes in is to make this feature optional and off by default to make sure we ensure compatibility for existing clients of InputPlumber (SteamOS, OpenGamepadUI, etc.). Once we have this feature available and these clients are ready, we can push to enable these polkit policies by default.

src/dbus/polkit.rs
Pandapip1: What about a compile-time cargo feature flag?

ShadowApex: A compile-time flag also works. Thanks!

Pandapip1:
Done!

ShadowApex (approved):

PR #415, opened 2025-07-31
fix(Legion Go): Improve new PID support

pastaq:
- Clean up old and unused events from Legion Go driver. Driver has been remodeled to be similar to the newer Go S driver.
- Only grab the Touchpad and Gamepad targets, allowing FPS mode to function properly.
- Add events unique to the Go 2.
- Combine Go and Go 2 configs as the now use the same PID for controllers.


src/drivers/lego/hid_report.rs
```diff
@@ -275,8 +309,8 @@ pub struct XInputDataReport {
     #[packed_field(bytes = "8")]
     pub r_controller_mode0: u8, // 64 - 71
 
-    #[packed_field(bytes = "9")]
-    pub gamepad_mode: u8,
+    #[packed_field(bytes = "9", ty = "enum")]
+    pub gamepad_mode: GamepadMode,
```
ShadowApex, line 313: Nice

src/input/source/hidraw/lego_xinput.rs
```diff
@@ -262,12 +233,16 @@ fn translate_event(event: event::Event) -> NativeEvent {
...
                 InputValue::Bool(value.pressed),
             ),
+            event::GamepadButtonEvent::ShowDesktop(value) => NativeEvent::new(
+                Capability::Gamepad(Gamepad::Button(GamepadButton::Keyboard)),
+                InputValue::Bool(value.pressed),
+            ),
+            event::GamepadButtonEvent::AltTab(value) => NativeEvent::new(
+                Capability::Gamepad(Gamepad::Button(GamepadButton::RightPaddle2)),
+                InputValue::Bool(value.pressed),
+            ),
```
ShadowApex, line 243: Should we create a new button capability for this? Maybe like `GamepadButton::SwitchWindow`? Not critical, we can always revisit this later.

ShadowApex (commented):
Looks good to me! Will do some testing on the Go 1 to see if there are any other issues.

src/input/source/hidraw/lego_xinput.rs
pastaq: I discussed with Pierre-Loup. Until we're switching to UCIS this prevents us from over tasking the deck hid report and makes the button more easily accessible. On both models the rightmost button is difficult to hit.

ShadowApex (approved):

PR #416, opened 2025-07-31
fix(Tester): set dbus path to allow debug interface to emit events

### LINKED ISSUE #409: Input Tester not working
In the latest version of InputPlumber (0.60.7), the CLI input test no longer seems to be working.

Image

ShadowApex:
This change fixes the CLI input tester which was broken in v0.60.4.

Fixes #409

### DIFF
--- a/src/input/target/debug.rs
+++ b/src/input/target/debug.rs
@@ -180,6 +180,7 @@ impl TargetInputDevice for DebugDevice {
     ) {
         let iface = TargetDebugInterface::new();
         dbus.register(iface);
+        self.dbus_path = Some(dbus.path().to_string());
     }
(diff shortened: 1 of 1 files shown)


pastaq (approved):

LukeShortCloud:
+1 I built and tested this locally. The test tool is working again.

A separate issue, probably related to my screen size, but I couldn't test my Select and Dpad-Right buttons on my GPD Win4 2025 edition. More specifically, they were not in the list of controller keys. For example, in the picture below, look to the very right. There are 3 (not 4) "Dpad" buttons. One for up, down, and left (not right).

PXL_20250802_034015477~2

PR #419, opened 2025-08-01
feat(GPD Win4): add 2025 model

LukeShortCloud:


pastaq:
The bmi260 has a kennel driver, but it requires out of tree "firmware" to be installed for the device init.

This is the file:
https://share.google/0pLieG4sQJAGF1TuQ

Must distros have been packaging this directly. You'll need to look at the upstream bmi270 driver for the install location as I don't remember off the top of my head.

rootfs/usr/share/inputplumber/devices/50-gpd_win4.yaml
```diff
@@ -36,8 +37,24 @@ source_devices:
...
+      phys_path: usb-0000:66:00.0-2.1/input0
+      handler: event*
+  - group: keyboard
+    evdev:
+      name: "  Mouse for Windows"
+      handler: event*
+      phys_path: usb-0000:66:00.0-2.2/input1
+  - group: imu
+    iio:
+      name: i2c-BMI0260:00
```
pastaq, line 56: This is the format for the out of tree driver. Add the upstream naming convention as well.

LukeShortCloud: I'm not sure what you mean (I'm still a newb here haha). Where can I look to find that information? I'm using a kernel from Nobara, should I boot into a vanilla kernel?

LukeShortCloud:
Good news! With the driver you linked, the IIO device is picked up and I verified via the InputPlumber test tool that both the accelerometer and gyro work. :-)

```
$ sudo inputplumber device 0 info
╭────┬──────────┬──────────────┬──────────────────────────────────────────────────────────────────╮
│ Composite Device                                                                                │
├────┼──────────┼──────────────┼──────────────────────────────────────────────────────────────────┤
│ Id │ Name     │ Profile Name │ Source Devices                                                   │
├────┼──────────┼──────────────┼──────────────────────────────────────────────────────────────────┤
│ 0  │ GPD Win4 │ Default      │ ["/dev/input/event12", "/dev/input/event11", "/dev/iio:device0"] │
╰────┴──────────┴──────────────┴──────────────────────────────────────────────────────────────────╯
```

rootfs/usr/share/inputplumber/devices/50-gpd_win4.yaml
LukeShortCloud: Based off your earlier comment, I guess I still need to figure out how to get the `bmi270` driver in the Linux kernel to load and attach to this device.

pastaq: https://github.com/torvalds/linux/blob/a6923c06a3b2e2c534ae28c53a7531e76cc95cfa/drivers/iio/imu/bmi270/bmi270_i2c.c#L55

ShadowApex, line 56: To match both the out-of-tree and in-tree driver, you can make this change:

```suggestion
      name: {i2c-BMI0260:00,bmi260}
```

pastaq: I don't remember the exact name. We can revisit later.

pastaq (approved):

rootfs/usr/share/inputplumber/devices/50-gpd_win4.yaml
LukeShortCloud: Are you saying I need to literally add the driver/module name? @ShadowApex mentioned to me earlier today there are ways to specify more than one parameter. So I would need something like this?

`name: "{i2c-BMI0260:00,bmi270_i2c}"`

LukeShortCloud: Ah, @ShadowApex just confirmed, yeah. Thanks! I'll get that updated tomorrow.

LukeShortCloud: Alright, I won't make any changes to the PR and leave it as-is for now.

PR #341, opened 2025-04-22
fix(ASUS Ally): don't crash out of driver init if attrs fail

flukejones:
Instead of bombing out of the driver init process allow IP to keep working as the majority of parts should still be usable.

PR #425, opened 2025-08-11
fix(Ally X): add limited functionality if asus_ally_hid driver is not available

ShadowApex:
This change adds limited support for the Ally X controller if Luke Jone's `asus_ally_hid` driver is not present.


pastaq:
While we're touching this, we should also add backported support for hid-asus-ally. The refactored asus-ally-hid won't get accepted upstream and has issues that caused SteamOS, CachyOS, etc to go back to the earlier version.

ShadowApex:
> While we're touching this, we should also add backported support for hid-asus-ally. The refactored asus-ally-hid won't get accepted upstream and has issues that caused SteamOS, CachyOS, etc to go back to the earlier version.

Isn't that already supported? According to Luke it should work with both?:
https://github.com/ShadowBlip/InputPlumber/pull/341/commits/b03624b0236d34da5b12a7007e9fa80f8d009d8f

pastaq (approved):

PR #426, opened 2025-08-15
fix(Generic Gamepad): ignore Framework Laptop 16 keyboard

LukeShortCloud:


LukeShortCloud:
In case it's useful for the future, here's the full `udevadm info` for that device.

[udevadm-info-framework-16-keyboard-ansi-system-control.txt](https://github.com/user-attachments/files/21804103/udevadm-info-framework-16-keyboard-ansi-system-control.txt)

PR #432, opened 2025-09-08
fix(Legion Go/2): Improve Touchpad Support

pastaq:
- Delay release events when using tap to click. This enables the use of Steam Input overlays by preventing the overlay from deactivating before the click event is registered.
- Improve rejecting the firmware "hold to click" mode.
- While at it, do some housekeeping with variable names and logging.


rootfs/usr/share/inputplumber/devices/50-legion_go_2.yaml
```diff
@@ -0,0 +1,129 @@
...
+      product_id: "{61e[b-d]}" # Don't grab 5/e, thats FPS mode Keyboard events
+      handler: event*
+  # Don't block evdev gamepad, used for rumble
+  - group: gamepad
+    unique: false
+    evdev:
+      name: "{Lenovo Legion Controller for Windows,Generic X-Box pad}"
+      vendor_id: "17ef"
+      product_id: "{61e[b-e]}"
+      handler: event*
```
ShadowApex, line 68: Do we want to filter out these events? E.g.

```yaml
    ...
    events:
      # Block input events, but use device for output events
      exclude:
        - "*"
```

src/drivers/iio_imu/driver.rs
```diff
@@ -1,8 +1,16 @@
...
+    fs::File,
+    io::{self, BufRead, BufReader},
+};
 
 use industrial_io::{Channel, ChannelType, Device, Direction};
 
-use crate::drivers::iio_imu::info::MountMatrix;
+use crate::{
+    drivers::iio_imu::info::MountMatrix,
+    input::capability::{Capability, Source},
```
ShadowApex, line 12: Ideally drivers should be standalone modules that do not rely on any InputPlumber data types. I would like to eventually move each of these drivers into their own crate that does not have any InputPlumber dependencies.

src/drivers/lego/driver.rs
```diff
@@ -0,0 +1,628 @@
...
+
+use crate::input::capability::Source;
+use crate::udev::device::UdevDevice;
+use crate::{
+    drivers::lego::{
+        event::{ImuAxisInput, TouchAxisInput, TouchButtonEvent},
+        hid_report::GamepadMode,
+        CLICK_DELAY, PAD_FORCE_NORMAL, RELEASE_DELAY, XINPUT_COMMAND_ID,
+    },
+    input::capability::Capability,
```
ShadowApex, line 17: Same thing here, we should try not to rely on InputPlumber data types inside the driver implementations.

src/drivers/lego/driver.rs
```diff
@@ -0,0 +1,628 @@
...
+        let Some(parent) = device.parent() else {
+            return Ok(HashSet::from([
+                Capability::Accelerometer(Source::Left),
+                Capability::Accelerometer(Source::Right),
+                Capability::Accelerometer(Source::Center),
+                Capability::Gyroscope(Source::Left),
+                Capability::Gyroscope(Source::Right),
+                Capability::Gyroscope(Source::Center),
+            ]));
+        };
```
ShadowApex, line 92: If you want, you can reduce the amount of duplication here by defining a constant:

```suggestion
        const DEFAULT_FILTER: [Capability; 6] = [
            Capability::Accelerometer(Source::Left),
            Capability::Accelerometer(Source::Right),
            Capability::Accelerometer(Source::Center),
            Capability::Gyroscope(Source::Left),
            Capability::Gyroscope(Source::Right),
            Capability::Gyroscope(Source::Center),
        ];
        let device = self.udev_device.get_device()?;
        let Some(parent) = device.parent() else {
            return Ok(HashSet::from(DEFAULT_FILTER));
        };
```

src/drivers/lego/driver.rs
```diff
@@ -0,0 +1,628 @@
...
+                    y: state.touch_y,
+                })));
+            }
+        }
+        events
+    }
+
+    fn start_click(&mut self) -> Vec<Event> {
+        if self.is_clicked {
+            log::debug!("Regecting extra click");
```
ShadowApex:
```suggestion
            log::debug!("Rejecting extra click");
```

src/input/event/value.rs
```diff
@@ -132,28 +118,28 @@ impl InputValue {
...
                             },
-                            // Gamepad Button -> Keyboard
                             Capability::Keyboard(_) => Ok(self.clone()),
-                            // Gamepad Button -> Touchpad
                             Capability::Touchpad(touch) => match touch {
                                 Touchpad::LeftPad(_) => Err(TranslationError::NotImplemented),
                                 Touchpad::RightPad(_) => Err(TranslationError::NotImplemented),
                                 Touchpad::CenterPad(_) => Err(TranslationError::NotImplemented),
                             },
-                            // Gamepad Button -> Touchscreen
```
ShadowApex, line 150: Why were these comments removed?

src/input/source/evdev/gamepad.rs
```diff
@@ -524,6 +524,8 @@ impl SourceOutputDevice for GamepadEventDevice {
                     capabilities.push(OutputCapability::ForceFeedback);
                     capabilities.push(OutputCapability::ForceFeedbackUpload);
                     capabilities.push(OutputCapability::ForceFeedbackErase);
+                    capabilities.push(OutputCapability::Haptics(Haptic::TrackpadLeft));
+                    capabilities.push(OutputCapability::Haptics(Haptic::TrackpadRight));
```
ShadowApex, line 528: Is there a reason these output capabilities are added for all evdev gamepads that support force feedback...?

src/input/source/mod.rs
```diff
@@ -9,6 +9,7 @@ use std::{
 };
 
 use ::evdev::FFEffectData;
+use clap::error::Result;
```
ShadowApex:
```suggestion
```

ShadowApex (changes requested):

src/drivers/iio_imu/driver.rs
pastaq: ~~I know, but I'm not going to replicate our massive match tree of all capabilities now for something we're not going to do for a while. If we decide to actually do that we can translate to driver events and send those.~~

Thinking on it more, I can just match on the supported capabilities list instead. That will eliminate unused events and limit depth.

src/input/event/value.rs
pastaq: CA to fill missing members removed them

src/input/source/evdev/gamepad.rs
pastaq: Because they are translated in the driver to evdev ffb, but it won't ever get there because they aren't reported as supported. Without this you get no rumble from deck targets to evdev sources

src/drivers/lego/driver.rs
pastaq: Thanks, I meant to go back and do that

src/input/source/evdev/gamepad.rs
pastaq: https://github.com/ShadowBlip/InputPlumber/blob/206d05b8bd35e20484be499b773e80514bbffed7/src/input/source/evdev/gamepad.rs#L581-L587

src/drivers/iio_imu/driver.rs
pastaq: This will require a refactor of how Event is handled entirely. Currently, Events and their values create a type. (e.g. Event::Axis(AxisEvent::LeftAccel(ImuAxisValue))), which causes the match {} to include the value type and value members. To allow Capability -> Event and Event -> Capability will require using a new DeviceEvent {Event, Value} paradigm. I have added an expository TODO for the IMU and Legion Go drivers where Capability is used.

ShadowApex (approved):

PR #435, opened 2025-09-10
fix(OrangePi NEO): Improve Touchpad Support

pastaq:
- Apply similar changes as #432 to improve the OrangePi NEO in the same way.

### DIFF
--- a/src/drivers/opineo/driver.rs
+++ b/src/drivers/opineo/driver.rs
@@ -7,7 +7,10 @@ use std::{
 use hidapi::HidDevice;
 use packed_struct::{types::SizedInteger, PackedStruct};
 
-use crate::udev::device::UdevDevice;
+use crate::{
+    drivers::opineo::event::{TriggerEvent, TriggerInput},
+    udev::device::UdevDevice,
+};
 
 use super::{
     event::{BinaryInput, Event, TouchAxisEvent, TouchButtonEvent},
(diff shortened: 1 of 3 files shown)


src/input/source/hidraw/opineo.rs
```diff
@@ -120,6 +120,16 @@ fn normalize_axis_value(event: event::TouchAxisEvent) -> InputValue {
     }
 }
 
+/// Normalize the trigger value to something between 0.0 and 1.0 based on the
+/// Legion Go's maximum axis ranges.
```
ShadowApex:
```suggestion
/// Orange Pi's maximum axis ranges.
```

src/input/source/hidraw/opineo.rs
```diff
@@ -162,18 +170,33 @@ fn translate_event(event: event::Event, touchpad_side: TouchpadSide) -> NativeEv
...
     Capability::Touchpad(Touchpad::RightPad(Touch::Button(TouchButton::Press))),
     Capability::Touchpad(Touchpad::RightPad(Touch::Motion)),
 ];
 
-pub const OUTPUT_CAPABILITIES: &[OutputCapability] = &[
-    OutputCapability::ForceFeedback,
-    OutputCapability::LED(LED::Color),
-];
+/// List of all output capabilities that the OrangePi NEO supports
+pub const OUTPUT_CAPABILITIES: &[OutputCapability] = &[OutputCapability::ForceFeedback];
```
ShadowApex, line 202: Does the OrangePi Neo not support LEDs?

src/drivers/opineo/driver.rs
```diff
@@ -197,15 +221,38 @@ impl Driver {
...
+        log::trace!("Last touch elapsed: {:?}", self.last_touch.elapsed());
+        self.is_clicked = false;
+        let mut events = Vec::new();
+        let event = Event::TouchButton(TouchButtonEvent::Left(BinaryInput { pressed: false }));
+        events.push(event);
+        // The touchpad doesn't have a force sensor. The deck target wont produce a "click"
+        // event in desktop or lizard mode without a force value. Simulate a 1/4 press to work
+        // around this.
+        let event = Event::Trigger(TriggerEvent::PadForce(TriggerInput { value: 0 }));
+        events.push(event);
```
ShadowApex, line 255: Instead of emulating the touch force in the source device, can we instead handle this in the target device(s)?

src/input/source/hidraw/opineo.rs
pastaq: It has RGB that is only controlled through button combo's on the device. OrangePi did not want to spend the money for an API for the MCU.

src/drivers/opineo/driver.rs
pastaq: I'd rather not manage source device edge cases in target devices.

PR #436, opened 2025-09-10
fix(polkit): check feature flag before creating new dbus connection

ShadowApex:
Small improvement to new polkit logic.


src/dbus/polkit.rs
```diff
@@ -1,13 +1,17 @@
-use zbus::{fdo, message::Header, Connection};
+use std::collections::HashMap;
+use zbus::zvariant::{OwnedValue, Value};
```
pastaq, line 2: Where are these new imports used?

ShadowApex: In the code in this file. These weren't used before because the entire function was conditionally removed based on a feature flag.

PR #437, opened 2025-09-12
feat(Force Feedback): add interface to toggle force feedback on/off

ShadowApex:
This change adds a new `enabled` property to the `org.shadowblip.Output.ForceFeedback` interface that can allow users to toggle on/off force feedback at the composite device level. This interface will only be available if a source device exists that supports force feedback.

It also fixes a longstanding issue where we do not remove capabilities when a source device is removed.

### DIFF
--- a/src/input/composite_device/mod.rs
+++ b/src/input/composite_device/mod.rs
@@ -718,7 +718,7 @@ impl CompositeDevice {
 
     /// Process a single output event from a target device.
     async fn process_output_event(&mut self, event: OutputEvent) -> Result<(), Box<dyn Error>> {
-        //log::trace!("Received output event: {:?}", event);
+        log::trace!("Received output event: {:?}", event);
 
         // Handle any output events that need to upload FF effect data
         if let OutputEvent::Uinput(uinput) = event.borrow() {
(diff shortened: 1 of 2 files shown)


src/input/composite_device/mod.rs
```diff
@@ -785,6 +807,12 @@ impl CompositeDevice {
             return Ok(());
         }
 
+        // If force feedback is disabled at the composite device level, don't
+        // forward any other FF events to source devices.
+        if !self.ff_enabled && event.is_force_feedback() {
+            return Ok(());
+        }
+
```
pastaq, line 815: Side note: this should be doable now.

src/input/composite_device/mod.rs
```diff
@@ -1610,25 +1683,44 @@ impl CompositeDevice {
...
+            }
+            self.output_capabilities_by_source
+                .insert(id.clone(), output_capabilities);
+
+            // Determine if the FF dbus interface should be created
+            let supports_ff = self
+                .output_capabilities
+                .contains(&OutputCapability::ForceFeedback);
+            let ff_iface_name = ForceFeedbackInterface::<CompositeDeviceClient>::name();
+            if supports_ff && !self.dbus.has_interface(&ff_iface_name) {
```
pastaq, line 1730: We should check if supports_ff is false and we've previously made the interface. That will cover if the ff device is removed and so are all ff capabilities.

src/input/composite_device/mod.rs
pastaq: Uhg, not showing the line I selected to comment. Refers to Todo for routing events only to capable devices

src/input/composite_device/mod.rs
ShadowApex: Can you elaborate more on this? I don't think I understand. Right now, if a source device is added that supports force feedback, we will add the force feedback dbus interface. When a source device is removed, we check if any surviving source devices support force feedback and if they don't, we remove the force feedback dbus interface.

pastaq: Oh, it happens at 1599. I missed that the first time I reviewed.

pastaq (approved):

PR #440, opened 2025-09-15
feat(udev): Use udev rules for configuring devices at startup

### LINKED ISSUE #82: Feature: Add turbo takeover function for OneXPlayer devices
Multiple versions of OneXPlayer devices have a turbo takeover function, where the turbo button becomes a momentary key press keyboard key. We already have this button mapped as the quick access button on multiple devices but we don't have a way for this functionality to be activated when a OneXPlayer device is resolved.

Adding the ability to have config specific scripts run could also help with #76.

pastaq:
- Adds 99-inputplumber-device-setup.rules, which contains rules on how to set up a device. This has a few advantages as we won't need to create polling device drivers for configuration interfaces or add entire new subsystems to device discovery.
- Adds rule for AOKZOE and OXP devices to take over the turbo button, fixes #82.
- Removes configuration driver for Legion Go S and replaces it with a simple rule.
- Adds commented out rules for ROG Ally devices. These rules have been tested and they work, but the driver creates sysfs paths before the device has initialized. This causes any udev rules that match the device to not be applied unless udevadm is re-triggered. Retaining for posterity if the issue can be fixed in the kernel.


pkg/rpm/inputplumber.spec
```diff
@@ -41,9 +41,8 @@ mkdir -p %{buildroot}/usr/share/polkit-1/actions
 install -D -m 755 %{_builddir}/InputPlumber/target/%{_arch}-unknown-linux-gnu/release/inputplumber %{buildroot}/usr/bin/inputplumber
 install -D -m 644 %{_builddir}/InputPlumber/rootfs/usr/share/dbus-1/system.d/org.shadowblip.InputPlumber.conf %{buildroot}/usr/share/dbus-1/system.d/org.shadowblip.InputPlumber.conf
 install -D -m 644 %{_builddir}/InputPlumber/rootfs/usr/lib/systemd/system/* %{buildroot}/usr/lib/systemd/system/
-install -D -m 644 %{_builddir}/InputPlumber/rootfs/usr/lib/udev/hwdb.d/59-inputplumber.hwdb %{buildroot}/usr/lib/udev/hwdb.d/59-inputplumber.hwdb
-install -D -m 644 %{_builddir}/InputPlumber/rootfs/usr/lib/udev/hwdb.d/60-inputplumber-autostart.hwdb %{buildroot}/usr/lib/udev/hwdb.d/60-inputplumber-autostart.hwdb
-install -D -m 644 %{_builddir}/InputPlumber/rootfs/usr/lib/udev/rules.d/90-inputplumber-autostart.rules %{buildroot}/usr/lib/udev/rules.d/90-inputplumber-autostart.rules
+install -D -m 644 %{_builddir}/InputPlumber/rootfs/usr/lib/udev/hwdb.d/* %{buildroot}/usr/lib/udev/hwdb.d/
```
ShadowApex, line 44: You will also need to make a similar change in `Cargo.toml` to ensure that this new udev rule is packaged.

ShadowApex (changes requested):

ShadowApex (approved):

PR #191, opened 2024-08-16
fix(OneXPlayer X1): Add OneXPlayer X1 series.

pastaq:

PR #441, opened 2025-09-16
fix(Hardware): Add Right Paddle support to F1 series, unify config.

pastaq:
- The M2 button is mapped by default to Guide. No other button emits this so map it to RightPaddle1
- Combine the configs for F1 series. 

TODO:
- Phys path for F1 OLED and F1 EVA-02 are pending.


src/drivers/oxfly_tty/event.rs
```diff
@@ -0,0 +1,77 @@
...
+    pub pressed: bool,
+}
+
+/// Axis input contain (x, y) coordinates
+#[derive(Clone, Debug)]
+pub struct JoyAxisInput {
+    pub x: i16,
+    pub y: i16,
+}
+/// Trigger input contains non-negative integars
```
ShadowApex, line 21:
```suggestion
}

/// Trigger input contains non-negative integars
```

src/drivers/oxfly_tty/driver.rs
```diff
@@ -0,0 +1,268 @@
...
+                pressed: state.status == ButtonStatus::Pressed,
+            })),
+            ButtonId::M2 => Event::GamepadButton(GamepadButtonEvent::M2(BinaryInput {
+                pressed: state.status == ButtonStatus::Pressed,
+            })),
+        }
+    }
+    /// Unpacks the buffer into a [JoystickDataReport] structure and updates
+    /// the internal button_state
+    fn handle_joystick_report(
```
ShadowApex, line 193:
```suggestion
    }
    
    /// Unpacks the buffer into a [JoystickDataReport] structure and updates
    /// the internal button_state
    fn handle_joystick_report(
```

src/drivers/oxfly_tty/driver.rs
```diff
@@ -0,0 +1,268 @@
...
+        buf: [u8; INPUT_REPORT_SIZE],
+    ) -> Result<Vec<Event>, Box<dyn Error + Send + Sync>> {
+        let input_report = ButtonDataReport::unpack(&buf)?;
+
+        // Print input report for debugging
+        //log::debug!("--- ButtonDataReport ---");
+        //log::debug!("{input_report}");
+        //log::debug!(" ---- ButtonDataReport ----");
+
+        //// Translate the input report into an event
```
ShadowApex:
```suggestion
        // Translate the input report into an event
```

src/drivers/oxfly_tty/driver.rs
```diff
@@ -0,0 +1,268 @@
...
+    // If the device needs to return to front-end mode or not
+    takeover_needed: bool,
+    // Timestamp of the last unkown data report.
+    last_data: Instant,
+    /// State for the internal gamepad controller joysticks
+    joystick_state: Option<JoystickDataReport>,
+}
+
+impl Driver {
+    pub fn new(udev_device: UdevDevice) -> Result<Self, Box<dyn Error + Send + Sync>> {
```
ShadowApex: In order to try to maintain standalone drivers that do not rely on inputplumber data types, it would be preferable if the devnode and port were passed to this function instead of an inputplumber `UdevDevice`:

```suggestion
    pub fn new(devnode: &str, port_number: u16) -> Result<Self, Box<dyn Error + Send + Sync>> {
```

src/drivers/oxfly_tty/driver.rs
```diff
@@ -0,0 +1,268 @@
...
+                        self.last_data = Instant::now();
+                        Vec::new()
+                    }
+                }
+            }
+            Err(_e) => {
+                //log::debug!("Got error: {e}");
+                vec![]
+            }
+        };
```
ShadowApex: nit: it would be nice to reduce the amount of nesting here:
```suggestion
        if let Err(_e) = self.port.read_exact(&mut buf) {
            //log::debug!("Got error: {e}");
            return Ok(vec![]);
        }

        //log::trace!("Got bytes: {:02x?}", &buf);
        let report_id = buf[1];
        let events = match report_id {
            BUTTON_DATA_REPORT => {
                let slice = &buf[..INPUT_REPORT_SIZE];
                let sized_buf = slice.try_into()?;
                self.handle_button_report(sized_buf)?
            }
            JOYSTICK_DATA_REPORT => {
                let slice = &buf[..INPUT_REPORT_SIZE];
                let sized_buf = slice.try_into()?;
                self.handle_joystick_report(sized_buf)?
            }
            TAKEOVER_DATA_REPORT => Vec::new(), // IGNORE ME
            _ => {
                // Mouse mode sends packets of a different size, so the data is jumbled.
                // We don't care about them, but when it exits the device will return to
                // normal gamepad mode. Flag the state change so we can automatically send
                // the mode switch command again.
                if !self.takeover_needed {
                    log::debug!("Unkown data report: {:02x?}", report_id);
                    self.takeover_needed = true;
                }
                self.last_data = Instant::now();
                Vec::new()
            }
        };
```

src/drivers/oxp_tty/driver.rs
```diff
@@ -0,0 +1,268 @@
...
+            if state.right_stick_x != old_state.right_stick_x
+                || state.right_stick_y != old_state.right_stick_y
+            {
+                events.push(Event::Axis(AxisEvent::RStick(JoyAxisInput {
+                    x: state.right_stick_x,
+                    y: state.right_stick_y,
+                })));
+            }
+        }
+        events
```
ShadowApex, line 261: nit: you can reduce nesting here with the `let Some(thing) = thing else { };` pattern:
```suggestion
        // Translate state changes into events if they have changed
        let Some(old_state) = old_state else {
            return events;
        };
        if state.left_trigger != old_state.left_trigger {
            events.push(Event::Axis(AxisEvent::TriggerL(TriggerInput {
                value: state.left_trigger,
            })));
        }
        if state.right_trigger != old_state.right_trigger {
            events.push(Event::Axis(AxisEvent::TriggerR(TriggerInput {
                value: state.right_trigger,
            })));
        }
        if state.left_stick_x != old_state.left_stick_x
            || state.left_stick_y != old_state.left_stick_y
        {
            events.push(Event::Axis(AxisEvent::LStick(JoyAxisInput {
                x: state.left_stick_x,
                y: state.left_stick_y,
            })));
        }
        if state.right_stick_x != old_state.right_stick_x
            || state.right_stick_y != old_state.right_stick_y
        {
            events.push(Event::Axis(AxisEvent::RStick(JoyAxisInput {
                x: state.right_stick_x,
                y: state.right_stick_y,
            })));
        }
        events
```

src/input/source/mod.rs
```diff
@@ -13,7 +13,7 @@ use led::LedDevice;
 use thiserror::Error;
 use tokio::sync::mpsc::{self, error::TryRecvError};
 
-use crate::config;
+use crate::{config, input::source::tty::TtyDevice};
```
ShadowApex: nit: we should move this import into the `use self::` statement below for consistency

src/input/source/tty.rs
```diff
@@ -0,0 +1,128 @@
...
+    input::{
+        capability::Capability,
+        composite_device::client::CompositeDeviceClient,
+        info::DeviceInfoRef,
+        output_capability::OutputCapability,
+        source::{tty::oxflyserial::OneXFlySerial, SourceDriverOptions},
+    },
+    udev::device::UdevDevice,
+};
+/// List of available drivers
```
ShadowApex, line 24:
```suggestion
};

/// List of available drivers
```

src/input/source/tty.rs
```diff
@@ -0,0 +1,128 @@
...
+enum DriverType {
+    Unknown,
+    OneXFlySerial,
+}
+/// [TtyDevice] represents an input device using the tty subsystem.
```
ShadowApex, line 31:
```suggestion
}

/// [TtyDevice] represents an input device using the tty subsystem.
```

src/input/source/tty.rs
```diff
@@ -0,0 +1,128 @@
...
+            "Finding driver for tty interface: {name}, port: {:04x?}",
+            port_number
+        );
+        if port_number == oxfly_tty::SERIAL_PORT {
+            return DriverType::OneXFlySerial;
+        }
+        DriverType::Unknown
+    }
+}
+/// Returns the DBus path for an [TtyDevice] from a device id (E.g. tty://ttyS0)
```
ShadowApex, line 180:
```suggestion
}

/// Returns the DBus path for an [TtyDevice] from a device id (E.g. tty://ttyS0)
```

src/input/source/tty/oxflyserial.rs
```diff
@@ -0,0 +1,244 @@
...
+
+        for event in events {
+            let new_event = self.translate_event(event);
+            new_events.push(new_event);
+        }
+
+        new_events
+    }
+
+    /// Translate the given Legion Go event into a native event
```
ShadowApex:
```suggestion
    /// Translate the given oxfly event into a native event
```

src/input/source/tty/oxflyserial.rs
```diff
@@ -0,0 +1,244 @@
...
+    if event_value >= 0.0 {
+        let maximum = max - mid;
+        event_value / maximum
+    } else {
+        let minimum = min - mid;
+        let value = event_value / minimum;
+        -value
+    }
+}
+// Returns a value between 0.0 and 1.0 based on the given value with its
```
ShadowApex, line 181:
```suggestion
}

// Returns a value between 0.0 and 1.0 based on the given value with its
```

src/drivers/oxfly_tty/driver.rs
pastaq: There is a complication with sending specific data like that. The X1 and G1 models use a very similar serial interface, so much so that I can reuse the driver with minor modifications and just have two source implementations that report the different capabilities. (See https://github.com/ShadowBlip/InputPlumber/pull/191#issuecomment-3624973520 for details). The problem is that the tty device reports as ttyUSB0 and doesn't have a "port" attribute. Instead it has a VID/PID and bInterfaceNumber we need to match on. What I think I will do is separate the validation logic from the new() logic, call the validation logic from the inputplumber side tty source type discovery so we only need to validate it once. if it returns true we can then call new(), without needing to re-validate on the driver side as we've done previously.

ShadowApex: Sure, that works. But the validation logic shouldn't need the `UdevDevice` directly. You can pull that logic out of the driver and into the source device implementation.

src/drivers/oxp_tty/driver.rs
ShadowApex: @pastaq this was not updated

pastaq: Woops, I mistook it as the other nest reduction.

ShadowApex (approved):

PR #443, opened 2025-09-26
feat(Composite Device): expose serial for a persistent identifier

ShadowApex:
This change exposes a new `Serial` property on the `CompositeDevice` dbus interface which can be used to retrieve a persistent unique identifier that can be used to identify a specific device. This field is meant to be used for things like profile traveling, where you can correctly identify a specific controller and apply some settings to it.

 This should provide a good starting point for being able to uniquely identify a specific controller, but there are some challenges we will need to think about:
 
 1. Composite devices can have multiple source devices; which serial number should we use?
 2. Source devices may change during runtime; how should we handle this exactly?
 3. Some poorly implemented hardware may have no serial number at all; should we try to use some semi-unique identifier like product id + vendor id?

### DIFF
--- a/bindings/dbus-xml/org.shadowblip.Input.CompositeDevice.xml
+++ b/bindings/dbus-xml/org.shadowblip.Input.CompositeDevice.xml
@@ -41,6 +41,10 @@
      -->
     <property name="Name" type="s" access="read"/>
     <!--
+     Persistent identifier for the device
+     -->
+    <property name="Serial" type="s" access="read"/>
+    <!--
      Name of the currently loaded profile
      -->
     <property name="ProfileName" type="s" access="read"/>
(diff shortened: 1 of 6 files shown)


pastaq (commented):
My only concern is that serial is a bit ambiguous. I understand the properties exposed in the kernel all use serial in the title, but serial on it's own could be misinterpreted as a data line, or be confusing once we add serial input devices (like OXP use on recent hardware)

ShadowApex:
> My only concern is that serial is a bit ambiguous. I understand the properties exposed in the kernel all use serial in the title, but serial on it's own could be misinterpreted as a data line, or be confusing once we add serial input devices (like OXP use on recent hardware)

Do you have any suggestions? I didn't want to use `SerialNumber` either because it's not really a number.

pastaq:
UniqueID?

PR #445, opened 2025-09-30
feat: add mouse wheel support

LukeShortCloud:
for both horizontal and vertical movement.

### DIFF
--- a/rootfs/usr/share/inputplumber/profiles/mouse_keyboard_wasd.yaml
+++ b/rootfs/usr/share/inputplumber/profiles/mouse_keyboard_wasd.yaml
@@ -177,7 +177,13 @@ mapping:
         button: RightBumper
     target_events:
       - mouse:
-          button: WheelUp
+          wheel:
+            name: Vertical
+            direction: up
+      #- mouse:
+      #    wheel:
+      #      name: Horizontal
+      #      direction: right
 
   - name: LB
     source_event:
@@ -185,4 +191,10 @@ mapping:
         button: LeftBumper
     target_events:
       - mouse:
-          button: WheelDown
+          wheel:
+            name: Vertical
+            direction: down
+      #- mouse:
+      #    wheel:
+      #      name: Horizontal
+      #      direction: left
--- a/src/input/event/value.rs
+++ b/src/input/event/value.rs
@@ -6,6 +6,7 @@ use crate::{
 use super::dbus::Action;
 
 /// Possible errors while doing input value translation
+#[derive(Debug)]
 pub enum TranslationError {
     /// Translation not yet implemented
     NotImplemented,
@@ -138,6 +139,8 @@ impl InputValue {
                                 Mouse::Motion => Err(TranslationError::NotImplemented),
                                 // Gamepad Button -> Mouse Button
                                 Mouse::Button(_) => Ok(self.clone()),
+                                // Gamepad Button -> Mouse Wheel
+                                Mouse::Wheel(_) => self.translate_button_to_wheel(target_config),
                             },
                             // Gamepad Button -> Keyboard
                             Capability::Keyboard(_) => Ok(self.clone()),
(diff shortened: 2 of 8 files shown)


ShadowApex (approved):
Looks great! Thanks for this!

rootfs/usr/share/inputplumber/profiles/mouse_keyboard_wasd.yaml
```diff
@@ -177,12 +177,24 @@ mapping:
         button: RightBumper
     target_events:
       - mouse:
-          button: WheelUp
+          wheel:
+            name: Vertical
```
pastaq, line 181: What value is the name attribute providing? I.e. I can't have a horizontal up or vertical left? Plane is essentially assumed.

src/input/event/value.rs
```diff
@@ -678,6 +687,37 @@ impl InputValue {
...
+                "No mouse config found".to_string(),
+            ));
+        };
+        let Some(direction) = wheel.direction.as_ref() else {
+            return Err(TranslationError::InvalidTargetConfig(
+                "No wheel direction found to translate button to mouse wheel".to_string(),
+            ));
+        };
+        match direction.as_str() {
+            "up" => Ok(InputValue::Float(1.0)),
```
pastaq, line 711: Cases with the same return value can be combined.

```
"up" | "right => Ok(InputValue::Float(1.0)),
 "down" | "left" => Ok(InputValue::Float(-1.0)),      
```

pastaq:
@LukeShortCloud have you had a chance to look at my nits?

rootfs/usr/share/inputplumber/profiles/mouse_keyboard_wasd.yaml
LukeShortCloud: What do you mean? Are you saying there would be issues with trying to use both at the same time? The normal scroll up and down function and pushing left or right feel very different (at least on my mice). I think it would be clunky to try and and go horizontal up or vertical left.

src/input/event/value.rs
LukeShortCloud: Nice, I didn't know about that Rust functionality. Thanks! I'll update the code.

rootfs/usr/share/inputplumber/profiles/mouse_keyboard_wasd.yaml
pastaq: I think the name is redundant. Up and down will always be vertical, so it's not adding value. I agree that vertical left would be weird/clunky and doesn't make sense.

LukeShortCloud: Ah, I understand now. The `target_events.mouse.wheel.name` and `target_events.mouse.wheel.direct` attributes are somewhat redundant. So maybe I can consolidate into something like this.

Before:

```
      - mouse:
          wheel:
            name: Vertical
            direction: up
```

After:

```
      - mouse:
          wheel: Up
```

to align with similar existing target events. For example:

```
      - mouse:
          button: Right
```

ShadowApex: @pastaq The `name` is the name of the capability, and the direction is used for mapping a particular direction of the wheel if the user doesn't want to map the whole axis. Relying only on the direction kind of breaks convention with how we define mapping every other type of event.

E.g. axes are mapped with this convention:

```yaml
- gamepad:
    axis:
      name: LeftStick
      direction: left
```

pastaq: > @pastaq The `name` is the name of the capability, and the direction is used for mapping a particular direction of the wheel if the user doesn't want to map the whole axis. Relying only on the direction kind of breaks convention with how we define mapping every other type of event.
> 
> E.g. axes are mapped with this convention:
> 
> ```yaml
> - gamepad:
>     axis:
>       name: LeftStick
>       direction: left
> ```

This isn't an analogous situation. Vertical and horizontal aren't Capabilities, they are directions. Mouse:Wheel is a Capability that can do up down left right, just like Gamepad:LeftStick. If your analogy was accurate it would be Gamepad:LeftStick: Vertical:Up, which would be equally silly. There's also not a conflict with the button click as that should be handled by Mouse:Button:Middle

ShadowApex: So you would rather see `Mouse:Wheel` changed into a `Vector2` value? That will require some minor refactoring.

PR #449, opened 2025-10-19
Introduces quadratic scaling for more precise control with small stick movement.

fewtarius:
This PR changes analog input scaling to scale quadratically instead of linearly which makes small movement less abrupt for more precise control.

### DIFF
--- a/src/config/capability_map.rs
+++ b/src/config/capability_map.rs
@@ -226,6 +226,8 @@ pub struct AxisCapability {
     pub direction: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub deadzone: Option<f64>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub quadratic_scaling: Option<bool>,
 }
 
 #[derive(Debug, Deserialize, Serialize, Clone, JsonSchema, PartialEq)]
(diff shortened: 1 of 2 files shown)


fewtarius:
> I think this would be a good option to have, but we should make this configurable (and off by default for now).
> 
> We could add this as a capability config option:
> 
> [capability_map.rs](https://github.com/ShadowBlip/InputPlumber/blob/main/src/config/capability_map.rs#L194-L200)
> 
> ```rust
> #[derive(Debug, Deserialize, Serialize, Clone, JsonSchema, PartialEq)]
> #[serde(rename_all = "snake_case")]
> pub struct AxisCapability {
>     pub name: String,
>     pub direction: Option<String>,
>     pub deadzone: Option<f64>,
>     pub quadratic_scaling: Option<bool>,
> }
> ```
> 
> Then we can check for the option in the translate function:
> 
> [value.rs](https://github.com/ShadowBlip/InputPlumber/blob/main/src/input/event/value.rs#L175)
> 
> ```rust
> // Axis -> Axis
> Gamepad::Axis(_) => {
>     let use_scaling = target_config
>         .gamepad
>         .as_ref()
>         .and_then(|gamepad| gamepad.axis.as_ref())
>         .and_then(|axis| axis.quadratic_scaling)
>         .unwrap_or(false);
> 
>     match use_scaling {
>         false => Ok(self.clone()),
>         true => match self {
>             InputValue::Vector2 { x, y } => {
>                 let scaled_x = x.map(|v| v * v.abs());
>                 let scaled_y = y.map(|v| v * v.abs());
>                 Ok(InputValue::Vector2 {
>                     x: scaled_x,
>                     y: scaled_y,
>                 })
>             }
>             _ => Ok(self.clone()),
>         },
>     }
> }
> ```

Alright, I'll submit the changes when I can.

PR #461, opened 2025-11-26
fix(Service): use 'ProtectSystem=full' in service definition

ShadowApex:
This change adds `ProtectSystem=full` to give InputPlumber a read-only view of important parts of the filesystem. InputPlumber should not have to write anything to most parts of the filesystem.


pastaq (approved):
We might consider using strict here, but this is fine. In the future I'd like InputPlumber to be able to create device profiles and place them in etc, so if we did use strict we'd need to add an approved path for /etc/inputplumber, along with the /run paths we use for udev.

PR #463, opened 2025-12-08
fix(build): Fix local 'make test' build errors

pastaq:
- Fix errors from clippy and rust-analyzer from newer versions
- While at it, update deps
With this update `make test` will pass again when run locally.

I'm not super happy with the allow for large errors. The canonical true fix seems to be using a Box, but I wasn't able to figure out a simple fix for the errors from outside libs without rewriting all the client functions.

### DIFF
--- a/src/input/event/evdev.rs
+++ b/src/input/event/evdev.rs
@@ -936,13 +936,15 @@ fn input_event_from_value(
             // the minimum and maximum values for that axis depending on the
             // axis direction. This is typically done for DPad button input that
             // needs to be translated to an ABS_HAT axis input.
-            if axis_info.is_some() && axis_direction.is_some() {
-                let info = axis_info.unwrap();
-                let direction = axis_direction.unwrap();
-                match direction {
-                    AxisDirection::None => Some(value),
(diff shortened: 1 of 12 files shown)


src/input/event/evdev.rs
```diff
@@ -936,13 +936,15 @@ fn input_event_from_value(
...
             // axis direction. This is typically done for DPad button input that
             // needs to be translated to an ABS_HAT axis input.
-            if axis_info.is_some() && axis_direction.is_some() {
-                let info = axis_info.unwrap();
-                let direction = axis_direction.unwrap();
-                match direction {
-                    AxisDirection::None => Some(value),
-                    AxisDirection::Positive => Some(info.maximum() * value),
-                    AxisDirection::Negative => Some(info.minimum() * value),
+            if let Some(info) = axis_info {
```
pastaq, line 939: According to the docs, rust 2024 should allow if let chaining. 
https://doc.rust-lang.org/edition-guide/rust-2024/let-chains.html

That would be a nicer fix, but rust-analyzer errors on my system.

PR #476, opened 2025-12-16
fix(RPM): add missing PolKit policy

LukeShortCloud:
Without this, there are DBus authentication errors when using the 0.69.0 RPM.


LukeShortCloud:
Tested and working on my SuiPlay0X1.

ShadowApex (approved):

PR #478, opened 2025-12-18
feat(Hardware Support): Add GPD Win5 Support

honjow:
Currently, the Paddle button on win5, in addition to the user-configured hotkey, also emits f14 and f15, so it is captured as an input source. 
The top-right mouse/keyboard mode toggle button outputs f13 on short press and is temporarily mapped to Keyboard (although Keyboard hotkeys don't seem to work well right now). 
The home key is mapped to guide — although this duplicates GPD's `menu` button, that menu button position is indeed hard to press. The bottom-right keyboard button is mapped to the qam button. The home and keyboard buttons output different values on long press, and both have been handled uniformly to facilitate using the guide combination key.

### DIFF
--- a/rootfs/usr/share/inputplumber/capability_maps/gpd_menu.yaml
+++ b/rootfs/usr/share/inputplumber/capability_maps/gpd_menu.yaml
@@ -1,0 +1,24 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ShadowBlip/InputPlumber/main/rootfs/usr/share/inputplumber/schema/capability_map_v2.json
+# Schema version number
+version: 2
+
+# The type of configuration schema
+kind: CapabilityMap
+
+# Name for the device event map
+name: GPD Menu
+
+# Unique identifier of the capability mapping
+id: gpd_menu
+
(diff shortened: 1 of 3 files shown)


honjow:
> Would it make more sense to map mode switch to QAM and keyboard to keyboard since the keyboard button has a keyboard picture? Also, did you want to remap menu to something else since it's function is now duplicated in home? Qam2 or screenshot might make sense there

What is the difference between qam and qam2? Mapping the menu button to a screenshot might be good.I can swap qam and keyboard, although I actually prefer invoking qam from the bottom right more.

pastaq:
Yes. You'll need a v2 config attached directly to the source device itself.

Example:
https://github.com/ShadowBlip/InputPlumber/blob/55d5edd622d2acefa5924514acc2c1feb8662e9a/rootfs/usr/share/inputplumber/devices/60-LogitechDualAction.yaml#L30

https://github.com/ShadowBlip/InputPlumber/blob/main/rootfs/usr/share/inputplumber/capability_maps/logitech_dualaction.yaml

Make sure you increment the version at the top and not just the schema

PR #487, opened 2026-01-01
QOL: centralize normalize/denormalize functions in src/input/event/value.rs

### LINKED ISSUE #465: QOL: add repeated normalize/denormalize functions to InputValue
In our source and target device implementations, it is normal for us to re-implement the same [de]normalize_[un]signed_value functions. These are all called from a driver specific normalize_[axis|trigger]_value function. As the code is identical in every instance, we can reduce the amount of code we need to copy/paste if we include these translators into src/input/event/value.rs and import it where needed.

blakesmittty:
Fixes #465 

This refactor moves the repeated implementations of the [de]normalize_[un]signed_value functions into 'src/input/event/value.rs' so they can be reused by the source and target implementations.

No functional changes.

### DIFF
--- a/src/input/target/dualsense.rs
+++ b/src/input/target/dualsense.rs
@@ -35,6 +35,7 @@ use crate::{
         event::{
             native::{NativeEvent, ScheduledNativeEvent},
             value::InputValue,
+            value::{denormalize_signed_value_u8, denormalize_unsigned_value_u8},
         },
         output_capability::{OutputCapability, LED},
         output_event::OutputEvent,
@@ -439,11 +440,13 @@ impl DualSenseDevice {
                     GamepadAxis::LeftStick => {
                         if let InputValue::Vector2 { x, y } = value {
                             if let Some(x) = x {
-                                let value = denormalize_signed_value(x, STICK_X_MIN, STICK_X_MAX);
+                                let value =
+                                    denormalize_signed_value_u8(x, STICK_X_MIN, STICK_X_MAX);
                                 state.joystick_l_x = value
                             }
                             if let Some(y) = y {
-                                let value = denormalize_signed_value(y, STICK_Y_MIN, STICK_Y_MAX);
+                                let value =
+                                    denormalize_signed_value_u8(y, STICK_Y_MIN, STICK_Y_MAX);
                                 state.joystick_l_y = value
                             }
                         }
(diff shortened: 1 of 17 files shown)


src/input/target/dualsense.rs
```diff
@@ -440,23 +441,23 @@ impl DualSenseDevice {
                         if let InputValue::Vector2 { x, y } = value {
                             if let Some(x) = x {
                                 let value = denormalize_signed_value(x, STICK_X_MIN, STICK_X_MAX);
-                                state.joystick_l_x = value
+                                state.joystick_l_x = value as u8
```
pastaq: These should already be a u8, is there a specific reason you added a cast?

https://github.com/ShadowBlip/InputPlumber/blob/ffd726a9a450eb83df270b579348e48780b81dcb/src/drivers/dualsense/hid_report.rs#L207-L214
https://github.com/ShadowBlip/InputPlumber/blob/ffd726a9a450eb83df270b579348e48780b81dcb/src/drivers/dualsense/event.rs#L44-L49

src/input/target/horipad_steam.rs
```diff
@@ -126,23 +127,23 @@ impl HoripadSteamDevice {
                         if let InputValue::Vector2 { x, y } = value {
                             if let Some(x) = x {
                                 let value = denormalize_signed_value(x, JOY_AXIS_MIN, JOY_AXIS_MAX);
-                                self.state.joystick_l_x = value
+                                self.state.joystick_l_x = value as u8
```
pastaq: Same for these

https://github.com/ShadowBlip/InputPlumber/blob/ffd726a9a450eb83df270b579348e48780b81dcb/src/drivers/horipad_steam/hid_report.rs#L68-L76
https://github.com/ShadowBlip/InputPlumber/blob/ffd726a9a450eb83df270b579348e48780b81dcb/src/drivers/horipad_steam/event.rs#L84-L89

pastaq (changes requested):
I don't think we need the u8 casts. Aside from that it looks good.

pastaq:
@blakesmittty 
Thanks, that's much cleaner.

One other thing I didn't catch before but would be a good change would be to reduce the number of commits. As this is a simple change just one commit should be sufficient. The commit title for your first commit is actually perfect for capturing the scope of work. If you wanted to add the amplifying information from the other commits for all the specific changes to each source and target implementation  below the commit title, that would be okay, but it's not necessary.

blakesmittty:
@pastaq done 👍

pastaq (approved):

PR #489, opened 2026-01-05
feat(ForceFeedback): add dbus interface for scale

ShadowApex:
Control the sensitivity of force feedback.


pastaq:
I'm not convinced about this approach. I think it makes more sense to handle at the source device level instead of at the event level.

 Some devices have a hardware scale built in. How should we handle that? I.e. setting it to 1.0 when the hardware is at half intensity already could lead to confusion. Do we just hard enable max intensity?


What if a device has both haptics and rumble sensors, but one of the two sensors is far to aggressive while the other is fairly good. By scaling it globally you just inverted the problem instead.

PR #493, opened 2026-01-08
WIP add Steam Deck touchpad functionality and enable the device profile

GloriousEggroll:
this makes it so left and right touchpad mimic SteamOS functionality, with one minor improvement.

SteamOS behavior:
left pad = dpad, tactile bump when navigating menus. right pad = nothing

Enhanced behavior provided:
left pad = dpad, tactile bump when navigating menus. right pad = standard touchpad with cursor movement, left click select with tactile bump

Note: all controls are fully re-mappable per-game within steam, this just provides functionality to match steamos when in a steamos session instead of doing nothing like they currently do.

this also enables the device profile and changes its target to deck so that the driver is used.

requires https://github.com/ShadowBlip/OpenGamepadUI/pull/500 for OGUI to be able to apply profiles with leftpad/rightpad support

### DIFF
--- a/rootfs/usr/share/inputplumber/devices/50-steam_deck.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-steam_deck.yaml
@@ -61,7 +61,7 @@ options:
 
 # The target input device(s) to emulate by default
 target_devices:
-  - xbox-elite
+  - deck
   - mouse
   - keyboard
   #- touchscreen
--- a/src/input/source/hidraw/steam_deck.rs
+++ b/src/input/source/hidraw/steam_deck.rs
@@ -4,7 +4,7 @@ use std::{
     fmt::Debug,
     sync::{Arc, Mutex},
     thread,
-    time::Duration,
+    time::{Duration, Instant},
 };
 
 use evdev::{FFEffectData, FFEffectKind, InputEvent};
@@ -47,7 +47,17 @@ pub struct DeckController {
     lizard_mode_started: bool,
     lizard_mode_running: Arc<Mutex<bool>>,
     ff_evdev_effects: HashMap<i16, FFEffectData>,
+    left_click_until: Option<Instant>,
+    right_click_until: Option<Instant>,
+    rumble_last: (u16, u16),
+    leftpad_dpad: DpadState,
+    leftpad_pressed: bool,
+    leftpad_last_xy: Option<(f64, f64)>,
 }
+
+const CLICK_MS: u64 = 25;
+const CLICK_STRENGTH: u16 = u16::MAX;
+const HAPTIC_MAX: u16 = 32767;
 
 impl DeckController {
     /// Create a new Deck Controller source device with the given udev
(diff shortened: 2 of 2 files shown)


rootfs/usr/share/inputplumber/devices/50-steam_deck.yaml
```diff
@@ -57,11 +57,11 @@ options:
   # If true, InputPlumber will automatically try to manage the input device. If
   # this is false, InputPlumber will not try to manage the device unless an
   # external service enables management of the device. Defaults to 'false'
-  auto_manage: false
+  auto_manage: true
```
pastaq: This change is something we might not be able to ship. Since InputPlumber is shipping on SteamOS/deck hardware Valve explicitly do not want InputPlumber to run on the deck. With the udev rule (IIRC?) not including the Deck it shouldn't start , but this would need to be verified on SteamOS for regressions.

src/input/source/hidraw/steam_deck.rs
```diff
@@ -587,6 +658,162 @@ fn translate_event(event: steam_deck::event::Event) -> NativeEvent {
...
 
+
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq)]
+struct DpadState {
+    up: bool,
+    down: bool,
+    left: bool,
+    right: bool,
+}
+
```
pastaq, line 681: Please add a function description that specifies the reason for the function and its expected behavior.

src/input/source/hidraw/steam_deck.rs
```diff
@@ -587,6 +658,162 @@ fn translate_event(event: steam_deck::event::Event) -> NativeEvent {
...
+    leftpad_dpad: &mut DpadState,
+    leftpad_pressed: &mut bool,
+    leftpad_last_xy: &mut Option<(f64, f64)>,
+) -> Vec<NativeEvent> {
+    let mut out = Vec::new();
+
+    for e in events {
+        match e {
+            // --- LEFT PAD PRESS: gate D-pad output ---
+            steam_deck::event::Event::Button(steam_deck::event::ButtonEvent::LPadPress(v)) => {
```
pastaq, line 790: We can probably `use crate steam_deck::event` to clean these up a bit

src/input/source/hidraw/steam_deck.rs
```diff
@@ -587,6 +658,162 @@ fn translate_event(event: steam_deck::event::Event) -> NativeEvent {
...
+                    if new_state != old {
+                        *leftpad_dpad = new_state;
+                        out.extend(emit_dpad_diffs(old, new_state));
+                    }
+                }
+
+                continue;
+            }
+
+            // Optional: swallow left pad touch so apps never see it
```
pastaq: How does this stay mappable in Steam if we eat the events?

pastaq (changes requested):
I'm not convinced this is the best approach, but I'm not really able to articulate why at this time. It feels like a lot of what is happening here is workarounds that should either be in the driver, target device, or exposed as configurable to OpenGamepadUI. I'd like @ShadowApex to review before we merge this, but I do have a couple nits in the meantime.

src/input/source/hidraw/steam_deck.rs
GloriousEggroll: its not swallowing them, thats a leftover comment, i removed the code that swallows the events, forgot to remove the comment

rootfs/usr/share/inputplumber/devices/50-steam_deck.yaml
pastaq: Update: build tests fail without a corresponding update to udev, so this can't ship.

GloriousEggroll: resolved per convo, remains disabled for now

PR #504, opened 2026-01-19
fix(Device Hiding): move hidden device nodes to /dev/inputplumber/sources

### LINKED ISSUE #502: Feature: move /dev/input/event* to /dev/inputplumber
On system where there is only a root user, changing the permissions of the `/dev/input/event` files is not enough.
ie: SDL still see the sources devives grabbed by InputPlumber

Moving them away ie to `/dev/inputplumber` does the trick (eventually controlled by an option)

ShadowApex:
This change updates the device hiding logic to move the devnodes into `/dev/inputplumber/sources` when a device is hidden. It also includes a better fallback mechanism for detecting the `chmod` and `mv` binaries for the udev rules.

This change should be tested against many devices to ensure this logic doesn't cause any regressions.

Fixes #502

### DIFF
--- a/src/udev/mod.rs
+++ b/src/udev/mod.rs
@@ -6,7 +6,11 @@ pub mod device_test;
 
 pub mod device;
 
-use std::{error::Error, fs, path::Path};
+use std::{
+    error::Error,
+    fs,
+    path::{Path, PathBuf},
+};
 
 use tokio::process::Command;
 use udev::Enumerator;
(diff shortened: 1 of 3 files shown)


src/udev/mod.rs
```diff
@@ -30,11 +30,33 @@ pub async fn hide_device(path: &str) -> Result<(), Box<dyn Error>> {
         return Err("Unable to create match rule for device".into());
     };
 
+    // Create the directory to move devnodes to
+    tokio::fs::create_dir_all("/dev/inputplumber/sources").await?;
+
     // Find the chmod command to use for hiding
     let chmod_cmd = if Path::new("/bin/chmod").exists() {
-        "/bin/chmod"
+        "/bin/chmod".to_string()
```
pastaq: Why not always use which?

pastaq:
Closes #502

src/udev/mod.rs
ShadowApex: Slight speed optimization and also the rare case where `which` might not be installed.

ShadowApex:
> I'll test this when I can.
> 
> Do you remember what the downsides of moving these were & why we pivoted to udev only?

I believe we always used pure udev + chmod hiding with InputPlumber and only moved the device nodes back with HandyGCCS. If I remember correctly, there was some pain points with the nodes not being restored when HandyGCCS would stop.

pastaq (changes requested):
~~This PR breaks the Legion Go & Legion Go 2 when the controllers are used in detached mode. I suspect anything that has a mode shift (i.e. OneXPlayer keyboard mode) will run into similar issues. All the device nodes are still in `/dev/inputplumber/sources` after the mode change. I think it would make sense to make this an option in the device config that defaults to false so we only enable this on devices that really need it.

Edit:
After re-reading #502 I realized that wouldn't work for that scenario. A runtime option like we do for manage all devices perhaps? Or detect env var in the launch so it can be set in the systemd unit~~

Edit2:
Turns out this was a false positive. A recent firmware update broke the "dinput" mode by removing the endpoint we get our data from.

pastaq:
That is correct

loki666:
ok I'll test that

pastaq:
> ok I'll test that

It needs to be set to =1

PR #508, opened 2026-01-21
fix(HardwareSupport): Ensure default capabilities & misc fixes

pastaq:
- Adds (temporary) target device translations for screenshot capability for all target gamepads. Each device uses a different combination, and some devices (steam deck, horipad) need to queue events to avoid issues with input collision. These will be fairly difficult to manage with profiles. This makes them unsuitable for the default profile.
- Adds (temporary) target device translations for CenterPad -> RightPad and RightPad -> CenterPad for all target gamepads with touch capabilities. Ensures that any source -> target combination has a working default behavior. Fixes multiple reports of "Dualsense Touchpad doesn't work in InputPlumber" from devices with right pads. As some devices need CenterPad -> RightPad, some need RightPad -> CenterPad, and some need no translations, depending on the target, they are not suitable for the defualt profile.
- Fixed touchpad translation for Dualsense source device. It was using normalize_signed when it should have been unsigned.
- Reduce poll rate of dualsense target to match the real hardware (250Hz)
- Add IMU timestamp to dualsense target.
- Fix logic for touch events in Dualsense target. Previously it was assumed that 127 = touching and 128 = released, but it is actually a touch counter that ranged 0-127, and the "off" state for each touch is
  \+ 128 of the current counter.
- Fix max value for X and Y of the Dualsense (off by 1).
- Switch dualsense to use the new pitch/roll/yaw names for gyro events.
- Implement new gyro/accel events for Steam Deck, Dualsense, and horipad targets. We should review any remaining source devices for the deprecated Gamepad:Gyrosope|Accelerometer events and update them in another PR to remove the Composite Device direct translations. (This one already got out of hand a bit)
- Switch Legion Go 2 driver to use QuickAccess2 instead of Screenshot for one of the new buttons. This will still be Screenshot when using the default capability map.
- Change a few debug statements to trace.
- Clean up some target device match trees with catch all's to improve readability and reduce nesting in some cases.


src/input/target/dualsense.rs
```diff
@@ -972,6 +998,9 @@ impl TargetInputDevice for DualSenseDevice {
             Capability::Touchpad(Touchpad::CenterPad(Touch::Button(TouchButton::Press))),
             Capability::Touchpad(Touchpad::CenterPad(Touch::Button(TouchButton::Touch))),
             Capability::Touchpad(Touchpad::CenterPad(Touch::Motion)),
+            Capability::Touchpad(Touchpad::RightPad(Touch::Button(TouchButton::Press))),
+            Capability::Touchpad(Touchpad::RightPad(Touch::Button(TouchButton::Touch))),
+            Capability::Touchpad(Touchpad::RightPad(Touch::Motion)),
```
ShadowApex, line 1003: I hate this, lol

src/input/target/mod.rs
```diff
@@ -849,17 +846,12 @@ impl TargetDevice {
                 let driver = TargetDriver::new_with_options(id, device, dbus, options);
                 Ok(Self::Touchscreen(driver))
             }
-            "xb360" | "gamepad" => {
-                let device = XBox360Controller::new()?;
-                let driver = TargetDriver::new(id, device, dbus);
-                Ok(Self::XBox360(driver))
-            }
```
ShadowApex, line 856: I don't agree with the removal of this device. Users should be able to have the choice to choose this target device and there may be rare edge cases where Xbox 360 gamepad emulation is required.

src/input/target/steam_deck.rs
```diff
@@ -586,8 +582,34 @@ impl SteamDeckDevice {
...
+                            let guide =
+                                ScheduledNativeEvent::new(guide, Duration::from_millis(240));
+                            let bumpr =
+                                ScheduledNativeEvent::new(bumpr, Duration::from_millis(160));
+                            (guide, bumpr)
+                        };
+
+                        self.queued_events.push(guide);
+                        self.queued_events.push(bumpr);
+                    }
```
ShadowApex, line 612: Again, I'm really not a fan of including Steam-specific shortcuts as part of the target device implementation.

src/input/target/mod.rs
pastaq: Maybe we can refactor the target device to take an argument of the type and it can configure the capabilities and pid based on that argument then? These are all nearly identical so we can lower the maintenance burden if we do that and still support older controllers.

src/input/target/dualsense.rs
pastaq: I understand. The issue is that we have people complaining that the touchpad doesn't work for the DS5 and DS5 Edge controller. All of 0 handhelds have a "Center" pad so the events will never make it to the target device. This can be solved with a profile, but we can't easily make a default profile that makes sense for all devices. I.E. if you have a dualsense source you need it to route to center pad for the dualsense target  and if you have a s team deck you need right pad to route to right pad on the deck target. The barrier for entry is also very high for adding/creating a profile on a lot of systems as they don't come packaged with OpenGamepadUI to make profiles, and even then you will need to know you need to do this for some devices and not others. In all its is a crapshoot right now. What my intention is for this PR is to improve the default experience as much as possible as these issues are difficult to solve in the short term.

src/input/target/steam_deck.rs
pastaq: What is the alternative that will fix the screenshot button? The combo is different depending on the target device, so this can't be done in the default profile. Shipping something that works now and can be improved later but isn't idiomatically perfect is sometimes better than a solution that is philosophically pure.

src/input/target/dualsense.rs
ShadowApex: I think this is acceptable as a short-term solution, but I just wanted to note that we should be looking for a more long-term solution for this problem.

src/input/target/mod.rs
ShadowApex: I'm fine with that, but that should be included as part of a separate PR.

src/input/target/steam_deck.rs
ShadowApex: Right, again I'm just noting that I think this is fine as a short-term solution, but we should try to make an effort towards a long term solution that doesn't hard-code Steam-specific shortcuts into target device implementations.

src/input/target/mod.rs
pastaq: Okay. I'll add it back and apply the same fixes for relevant events.

src/input/target/dualsense.rs
pastaq: I agree, which is why I left all the TODO's.

PR #510, opened 2026-01-31
feat(Hardware Support): Add tap-to-click and L4/R4 for GPD Win Mini

### LINKED ISSUE #157: L4/R4 and trackpad for GPD Win Mini
I couldn't get L4/R4 on the GPD Win Mini to respond as paddles on "Test Device Inputs" in Steam, and I couldn't map them to anything with a yaml config.. They also don't show in the InputPlumber logs when pressed. I can see the GPD Win Mini recognized from the InputPlumber logs.

I verified that `usb-0000:63:00.3-3/input1` sends KEY_SYSRQ and KEY_PAUSE for the physical L4/R4 buttons. They act as a key repeat instead of a single press, hold, release though.

I also noticed that the trackpad and physical keyboard aren't in the composite device config. The trackpad doesn't have a physical click, and gamescope-session doesn't support tap-to-click, so it would be perfect if inputplumber could map or handle a tap as a physical click.

the physical touchpad on the Win Mini is

```
NAME="HTIX5288:00 093A:0255 Touchpad"
PHYS="i2c-HTIX5288:00"
```

and the physical keyboard is

```
NAME="AT Translated Set 2 keyboard"
PHYS="isa0060/serio0/input0"
```

Ultimately my goal would be to create a config that would map trackpad with simulated click, L4/R4, buttons, and gyro all to a simulated DS5 or Steam Deck controller, so I could use it all with Steam Input.

Happy to help collect data or try things.

justinweiss:
Adds a tap-to-click similar to the Legion Go and the OrangePi NEO. 

Also debounces the macro keys and remaps them to L4/R4 internally. Then uses the capability map to remap R4 to Quick Access, since the Win Mini doesn't have a dedicated Quick Access button. L4 is free for Steam Input or other mapping.

Fixes #157.

### DIFF
--- a/src/drivers/gpd_win_mini/touchpad_driver.rs
+++ b/src/drivers/gpd_win_mini/touchpad_driver.rs
@@ -49,6 +49,8 @@ pub struct TouchpadDriver {
     is_touching: bool,
     /// Timestamp of the last touch event.
     last_touch: Instant,
+    /// Timestamp of the last release event.
+    last_release: Instant,
     /// Whether or not a touch event was started that hasn't been cleared.
     touch_started: bool,
     /// State for the touchpad device
@@ -74,6 +76,7 @@ impl TouchpadDriver {
             is_clicked: false,
             is_touching: false,
             last_touch: Instant::now(),
+            last_release: Instant::now(),
             touch_started: false,
             touchpad_state: None,
         })
@@ -166,6 +169,13 @@ impl TouchpadDriver {
                 self.first_touch_x = state.touch_x0;
                 self.first_touch_y = state.touch_y0;
                 log::trace!("Started TOUCH event");
+
+                // If this happened soon after a click, drag
+                if self.last_release.elapsed() <= CLICK_DELAY {
+                    log::trace!("Started DRAG event");
+                    let mut click_events = self.start_click();
+                    events.append(&mut click_events);
+                }
             }
         // Handle tap-to-click
         } else if !self.is_touching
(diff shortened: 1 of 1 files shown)


justinweiss:
> I recall we only need a 20ms release check to determine if it should be an up or down event. 

Yep, it should just need a simple debounce.

> That might be how steam is interpreting the generic target right now. If you change the vchi deck target to use the physical deck device PID it might have better results.

I tried the "Steam Deck Controller" target / PID with the same results:

```
[2026-02-01T21:14:21Z DEBUG inputplumber::input::target::steam_deck] Found Steam Deck target config: Valve Corporation Steam Controller PID: 4613
```

Do you know if it still works with either deck target for OrangePi / Legion Go? I've tried some tests to replicate the order and types of those events that should be sent those devices with no change in results. So I haven't been able to figure out why those would work and this one wouldn't.

The Steam controller tester shows the clicks, they just have no impact in the game mode UI or in-game.

pastaq:
> OK! This was a misunderstanding on my part 🙃 . After digging out my Deck and trying it out, I found out the mouse cursor isn't supposed to appear at all in game mode on the Deck / Steam Controller inputs, and that the default is to map right pad to joystick. Seeing the mouse cursor was confusing me into thinking it was supposed to click.
> 
> Blocking the event* for the touchpad in the device yaml makes it work just like the Deck -- no cursor in game mode, no clicks in game mode, controller tester shows correct behavior, remapping right trackpad click to mouse click works fine in-game. I'll add that to this PR soon.

Ah, yeah, that quirk has caused a lot of confusion. If you hold the guide button the trackpad switches to mouse mode

justinweiss:
@pastaq I made two other small tweaks here. One to ignore some taps that were causing accidental clicks, and another to support drag.

pastaq:
Cool. We'll include this in the next version. I'm waiting on a review for the mouse events PR and some info from Valve.

PR #514, opened 2026-02-03
feat: add mouse wheel support

pastaq:
replaces #445

I pulled in @LukeShortCloud commit and made some changes. He had already made the InputValue a vec2 so there was nothing to be done there. I added the InputValue -> REL_WHEEL and REL_HWHEEL to evdev.rs and moved all capability -> string into the standard location, and added  the new mouse wheel to the string list.

Here is a full diff of the changes from Luke's patch to my version.

```
< From b03631e9312fe1f602768ef80414cbeee805b949 Mon Sep 17 00:00:00 2001
---
> From 3abbad3eede76e2fa2dfd6e8ba5a1e5a373b3af9 Mon Sep 17 00:00:00 2001
4c4
< Subject: [PATCH] feat: add mouse wheel support
---
> Subject: [PATCH 1/2] feat: add mouse wheel support
6a7,9
>
> (cherry picked from commit b03631e9312fe1f602768ef80414cbeee805b949)
> Signed-off-by: Derek J. Clark <derekjohn.clark@gmail.com>
10,11c13,14
<  src/dbus/interface/composite_device.rs        |  2 +
<  src/input/capability.rs                       | 41 +++++++++++++++++
---
>  src/dbus/interface/composite_device.rs        | 20 +--------
>  src/input/capability.rs                       | 42 ++++++++++++++++++
13c16
<  src/input/event/evdev.rs                      | 14 +++++-
---
>  src/input/event/evdev.rs                      | 16 ++++++-
16c19
<  8 files changed, 124 insertions(+), 6 deletions(-)
---
>  8 files changed, 127 insertions(+), 24 deletions(-)
50c53
< index be15acd..d0ac977 100644
---
> index e4d0435..057dc33 100644
53,54c56
< @@ -236,6 +236,7 @@ pub struct DialCapability {
<  pub struct MouseCapability {
---
> @@ -275,6 +275,7 @@ pub struct MouseCapability {
55a58
>      #[serde(skip_serializing_if = "Option::is_none")]
61c64
< @@ -245,6 +246,13 @@ pub struct MouseMotionCapability {
---
> @@ -286,6 +287,13 @@ pub struct MouseMotionCapability {
76c79
< index e9995e3..c79accd 100644
---
> index 55730bd..e3db7ac 100644
79,94c82,115
< @@ -352,6 +352,7 @@ impl CompositeDeviceInterface {
<                  Capability::Mouse(mouse) => match mouse {
<                      Mouse::Motion => "Mouse:Motion".to_string(),
[...]

PR #518, opened 2026-02-11
feat(Hardware Support): Razer Tartarus Pro

dripsnek:
This PR implements support for the [Razer Tartarus Pro](https://www.razer.com/gaming-keypads/razer-tartarus-pro) gaming keypad. To date I've not found anything which allows usage of the analog mode specific to this model under Linux. This aims to bridge that gap.

The need for a driver exists because analog mode generates vendor defined HID reports that contain the positional value of each key instead of events. These must be processed and mapped to input events and this seemed like a natural fit for InputPlumber to handle.

I think this device is the first of its kind for InputPlumber and assuming the project wishes to extend support to these peripherals, how to represent them from a profile perspective in my mind is the major piece of work to go.

The driver has two modes, basic and analog. Basic mode simply uses the Tartarus Pro in its power-up configuration as a regular keyboard/mouse hybrid and is configured as any other device. Analog mode requires commanding the Tartarus Pro. This can occur manually before starting InputPlumber or if using OpenRazer >=3.12.2 you can use `driver_mode = True` in your config file to the same effect. `False` will leave the device in basic mode.

This implementation takes full control of the hidraw (x3) and evdev nodes (x4) but passes through the last endpoint allowing lighting control software e.g. OpenRGB or Polychromatic to manage the RGB functions whilst InputPlumber is running.

The analog event handler is implemented using simple linear regressions on a small buffer* to drive a per-key state machine and supports the following:
- User defined actuation points between 1.5mm** and 3.6mm depths
- Dual-function keys
- Retrigger*** with shared or separate upward and downward thresholds
- Wooting-style continuous retrigger***

_* The buffer depth of 5 was based on protocol analysis of fast key transitions from the device_
_** Quantization and implementation quirks label the minimum at 1.4mm but the Tartarus Pro marketing spec stands._
_*** AKA Rapid Trigger from other vendors_

Analog mode usage:

Analog mode parameters currently reside at `50-razer_tartarus_pro.yaml` in the analogkeys mapping. There are 5 arrays of 20 values, each value left-to-right represents the numbered key on the Tartarus Pro in order, putting aside 0 indexing. All 20 keys have individual analog settings and these will apply to any loaded key binding profiles. Values are sanitised with a specific exception in place for zeroes. Analog parameters cannot be changed at runtime, a restart of InputPlumber is required to take effect.
| Config | Definition | Limits | Notes |
| :---     | :---    | :---     | :--- |
| primary_actuation     | Absolute depth value for actuation of first key bind | 1.4 to 3.6mm in 0.1mm steps   | Cannot be disabled. A value of 0 will map to effectively 1.4mm. Key bindings are per basic mode, can be rebound using a profile   |
| secondary_actuation     | Set to enable dual-function. Absolute depth value for actuation of second key bind | 1.4 to 3.6mm in 0.1mm steps     | Default is disabled. Must be greater than primary_actuation to enable. Can be explicitly disabled by setting to 0, otherwise any invalid setting will be interpreted as disabled. Prevents usage of retrigger though a similar effect exists for the second actuation point due to how events are processed. Secondary bindings are mapped to fixed "phantom" keys which can be further customized via profile, see `secondary_rebind.yaml` | 
| retrigger_reset_threshold | Set to enable retrigger logic. Value reflects travel distance in mm to reset a key |  0 to 2.1mm in 0.1mm steps |  Default is disabled. Automatically disabled when secondary_actuation is active. Can be explicitly disabled by setting to 0mm.
| retrigger_trigger_threshold | Complements retrigger_reset_threshold. Travel distance in mm to trigger a key after being reset | 0 to 2.1mm in 0.1mm steps | If this value is 0 then retrigger_reset_threshold will be used for this purpose. This variable exists in case separate reset and trigger thresholds are desired
| continuous_retrigger | Complements retrigger_reset_threshold. During retrigger ignore the primary_actuation value, continue until key reaches the top of travel | String (Y/N) | Default is N. Has no effect if retrigger_reset_threshold is disabled.

Limits to this PR:
- No plans to include other types of analog events to this driver considering the recent issues surrounding Snap Tap.
- No Hypershift or equivalent mode, I think that's more of a wider InputPlumber capability rather than something specific to a device driver.
- The profile LEDs on the side of the Tartarus should eventually be supported by OpenRazer and changing their state is better managed by a profile front-end application rather than InputPlumber directly.
- RGB support is already implemented by Polychromatic and OpenRGB.

Other topics influencing implementation, not necessarily impacting the path out of WIP:
- Device details:
Like a lot of products in this class it is possible to acquire the serial number, firmware revision and other such data from the device registers. This could be used to uniquely assign profiles to a specific device or implement quirks for a given firmware. Should this be incorporated?
- Multi-instance support:
The Tartarus Pro is a left-handed device but someone, somewhere will somehow find a way to use 2 of them. Should we embrace this or defend against it?

Assuming this type of device can be included, I propose the following to navigate out of WIP:
- [x] Decide to [keep basic mode] ~~or refactor to only manage analog mode~~
- [ ] Decide to keep num_enum crate or refactor impacted code (event.rs / driver.rs - handle_basic)
- [x] If keeping basic mode, ~#514~ merged to support scroll wheel remapping
- [x] Incorporate device profile support and remove hard coded settings
- [x] ~~Return device to basic mode either on demand or if InputPlumber closes~~ InputPlumber assumes device state is managed externally, no implementation required.
- [ ] Write up device usage .md file at nominated location
- [x] Wire up analog event handler to the driver event vector
- [ ] Test what happens when two devices are connected
- [ ] Housekeeping prior to merge

Blockers:
#582 - As this impacts external applications managing device state


src/drivers/razer_tartarus_pro/driver.rs
```diff
@@ -0,0 +1,463 @@
...
+        let api = hidapi::HidApi::new()?;
+        let device = api.open_path(&cs_path)?;
+        let info = device.get_device_info()?;
+        let mut razer_message_id = 0 as u8;
+        if info.vendor_id() != VID || info.product_id() != PID {
+            return Err(format!("Device '{hidrawpath}' is not a Razer Tartarus Pro").into());
+        }
+        if info.interface_number() == RAZER_FEATURE_ENDPOINT {
+            // Check if we can get serial number
+            let mut packet = Self::razer_report(0x0, 0x82, 0x16, &[0x0], &mut razer_message_id);
```
pastaq: In general, we try to stay away from sending input reports back to the device aside from rumble events in DInput. I would much rather see these sorts of things added to a kernel driver, exposed over sysfs, and controlled by udev, either through rules or the udev crate.

src/drivers/razer_tartarus_pro/driver.rs
```diff
@@ -0,0 +1,463 @@
...
+                "Tartarus Serial Number: {:?}",
+                String::from_utf8(output).unwrap().trim_end_matches("\0")
+            );
+            // Enable analog mode
+            packet = Self::razer_report(0x0, 0x4, 0x2, &[0x3, 0x0], &mut razer_message_id);
+            log::info!("Enabling Tartarus Pro analog mode");
+            let _ = Self::transaction(packet, &device).unwrap();
+        }
+
+        let mut zeroes = VecDeque::with_capacity(REGRESSION_WINDOW);
```
pastaq, line 122: Can you add a comment explaining the purpose of this array?

src/drivers/razer_tartarus_pro/driver.rs
```diff
@@ -0,0 +1,463 @@
...
+        Ok(events)
+    }
+
+    // The hidapi crate enforces the Windows-ism of needing a dud zeroth byte for sending feature
+    // reports to devices with no report ID specified. For unused report ID This means our window
+    // is [1..] and we ignore [0], though leave it 0. The XOR is always the penultimate byte with
+    // the ultimate byte being 0. Byte [1] is a status byte which for PC->DEV is 0. Replies
+    // always start with 0x2. Bytes [3..5] are used in other Razer devices but not the Tartarus
+    // so we leave them 0.
+    fn razer_report(
```
pastaq: I'd prefer we use a PackedStruct instead of manually decoding the reports.

See https://github.com/ShadowBlip/InputPlumber/blob/main/src/drivers/legos/hid_report.rs for an example.

src/drivers/razer_tartarus_pro/driver.rs
```diff
@@ -0,0 +1,463 @@
...
+                });
+            }
+        }
+
+        // Save state for next time
+        self.key_state = pad_state;
+        Ok(events)
+    }
+
+    fn handle_analog(&mut self, keys: &[u8]) -> Result<Vec<Event>, Box<dyn Error + Send + Sync>> {
```
pastaq, line 353: In general this entire function is difficult to follow. It is very long, the variable names aren't obvious for what they are attempting to accomplish, and the comments refer to actions that aren't well defined. If you could take another look at this to make it less ambiguous I'd be able to provide more constructive feedback.

Please add a plain English explanation above the function using documentation markers `///`

src/drivers/razer_tartarus_pro/driver.rs
```diff
@@ -0,0 +1,463 @@
...
+                    }
+                    self.key_actions[index].keydown_up_n = false;
+                }
+            }
+        }
+
+        Ok(Vec::new())
+    }
+
+    fn linear_regression(&self, data: &[f64]) -> Option<f64> {
```
pastaq, line 604: Please add a plain English explanation above the function using documentation markers `///`

src/drivers/razer_tartarus_pro/driver.rs
```diff
@@ -0,0 +1,463 @@
...
+            let _ = handle.send_feature_report(&packet);
+            let mut feature_buffer: [u8; RAZER_HID_ARRAY_SIZE] = [0; RAZER_HID_ARRAY_SIZE];
+            let _ = handle.get_feature_report(&mut feature_buffer);
+            return Ok(Self::razer_decapsulate(&packet, &feature_buffer)?);
+        }
+        log::error!("Tartarus attempting to transact on incorrect endpoint");
+        return Ok(Vec::new());
+    }
+
+    fn handle_input_report(
```
pastaq, line 187: Please add a plain English explanation above the function using documentation markers `///`

src/drivers/razer_tartarus_pro/driver.rs
```diff
@@ -0,0 +1,463 @@
...
+        is_valid = (response[89] == xorval) && is_valid;
+        if !is_valid {
+            log::error!("Invalid Tartarus transaction received");
+            Ok(Vec::new())
+        } else {
+            Ok(response[9..(9 + response[6] as usize)].to_vec())
+        }
+    }
+
+    fn transaction(
```
pastaq: Please add a plain English explanation above the function using documentation markers `///`

src/drivers/razer_tartarus_pro/driver.rs
```diff
@@ -0,0 +1,463 @@
...
+        message[89] = xorval as u8;
+
+        // Increment next transaction id
+        *id = id.wrapping_add(1);
+        return message;
+    }
+
+    // This validates any received reports against the feature request and provides the payload
+    // as a vector for further processing.
+    fn razer_decapsulate(
```
pastaq: Please add a plain English explanation above the function using documentation markers `///`

src/drivers/razer_tartarus_pro/driver.rs
pastaq: Please add a plain English explanation above the function using documentation markers `///`

src/drivers/razer_tartarus_pro/driver.rs
```diff
@@ -0,0 +1,463 @@
...
+            2 => self.handle_basic(buf, KeyCodes::MClick, false),
+            _ => Ok(events),
+        }
+    }
+
+    // The basic case for all 3 endpoints is essentially identical. The first byte is unique for
+    // each; For endpoints 1.1 and 1.3 this is where code 0x04 is interpreted and discarded for
+    // endpoint 1.2 as it was a report ID which has served its purpose if we got this far.
+    // Given the conversion to variant space it is open to misinterpretation if left there.
+    fn handle_basic(
```
pastaq, line 231: Please add a plain English explanation above the function using documentation markers `///`

src/input/source/hidraw/razer_tartarus_pro.rs
```diff
@@ -0,0 +1,205 @@
...
+        razer_tartarus_pro::event::KeyCodes::Aux => NativeEvent::new(
+            Capability::Keyboard(Keyboard::KeyLeftAlt),
+            InputValue::Bool(event.pressed),
+        ),
+        razer_tartarus_pro::event::KeyCodes::MClick => NativeEvent::new(
+            Capability::Mouse(Mouse::Button(MouseButton::Middle)),
+            InputValue::Bool(event.pressed),
+        ),
+        razer_tartarus_pro::event::KeyCodes::ScrollDown => {
+            NativeEvent::new(Capability::Mouse(Mouse::Wheel), InputValue::Float(-1.0))
```
pastaq, line 247: Is this always -1 from the hardware? Many scroll wheels can have larger values depending on if they are high resolution or moved quickly.

src/input/source/hidraw/razer_tartarus_pro.rs
```diff
@@ -0,0 +1,205 @@
...
+fn translate_events(events: Vec<razer_tartarus_pro::event::Event>) -> Vec<NativeEvent> {
+    events.into_iter().map(translate_event).collect()
+}
+
+fn translate_event(event: razer_tartarus_pro::event::Event) -> NativeEvent {
+    match event.key {
+        razer_tartarus_pro::event::KeyCodes::Blank => {
+            NativeEvent::new(Capability::NotImplemented, InputValue::None)
+        }
+        razer_tartarus_pro::event::KeyCodes::ScrollUp => {
```
pastaq: Please order these events logically to make finding a given event easier in the match list. I.e. scroll wheels together, either numerically or alphabetically for the keys, etc.

src/input/source/hidraw/razer_tartarus_pro.rs
```diff
@@ -0,0 +1,205 @@
...
+            InputValue::Bool(event.pressed),
+        ),
+        razer_tartarus_pro::event::KeyCodes::ScrollDown => {
+            NativeEvent::new(Capability::Mouse(Mouse::Wheel), InputValue::Float(-1.0))
+        }
+    }
+}
+
+/// List of all capabilities that the Razer Tartarus Pro implements
+pub const CAPABILITIES: &[Capability] = &[
```
pastaq, line 259: Sort these as well please.

pastaq (changes requested):
Sorry for taking so long to review. Overall I am impressed with the direction this is going. I do have a few concerns about a couple of the approaches, but we can discuss them below in each of the issues I raised.

src/drivers/razer_tartarus_pro/driver.rs
dripsnek: OpenRazer just added support for the Tartarus Pro so there is no need to have a custom control pane now. Code has (pending push) been adjusted to use those sysfs nodes, but that does introduce a dependency on the razerkbd module for analog mode and potentially profile filtering to specific boards.

src/drivers/razer_tartarus_pro/driver.rs
dripsnek: Commentary in pending push.

src/drivers/razer_tartarus_pro/driver.rs
dripsnek: Section removed as control related code offloaded to razerkbd module

src/drivers/razer_tartarus_pro/driver.rs
dripsnek: Refactored in next push

src/drivers/razer_tartarus_pro/driver.rs
dripsnek: Section removed as control related code offloaded to razerkbd module

src/drivers/razer_tartarus_pro/driver.rs
dripsnek: Added

src/drivers/razer_tartarus_pro/driver.rs
dripsnek: Added

src/input/source/hidraw/razer_tartarus_pro.rs
dripsnek: USB packet captures and evtest at the raw device show it only ever returning 1 and -1, tried scrolling at multiple speeds.

src/input/source/hidraw/razer_tartarus_pro.rs
dripsnek: Reordered

src/input/source/hidraw/razer_tartarus_pro.rs
dripsnek: Reordered

src/drivers/razer_tartarus_pro/driver.rs
dripsnek: Added

src/drivers/razer_tartarus_pro/driver.rs
dripsnek: Section removed as control related code offloaded to razerkbd module

src/drivers/razer_tartarus_pro/driver.rs
dripsnek: Section removed as control related code offloaded to razerkbd module

rootfs/usr/share/inputplumber/profiles/secondary_rebind.yaml
pastaq, line 1: This file seems odd to me. The mappings are all 1-1. If the device is already populating native events (used for detection) with those values they will route to any target device that supports them. Can you explain what you were trying to accomplish here?

src/drivers/razer_tartarus_pro/driver.rs
```diff
@@ -0,0 +1,695 @@
...
+use crate::config::SourceDevice;
+use crate::udev::device::UdevDevice;
+use hidapi::HidDevice;
+use std::collections::VecDeque;
+use std::{error::Error, ffi::CString, fs, io::Write, path::Path};
+
+pub const VID: u16 = 0x1532;
+pub const PID: u16 = 0x0244;
+
+const ENDPOINT_1: i32 = 0;
```
pastaq, line 11: Do these need to be i32 for a specific reason? u8 seems more appropriate at a glance.

pastaq:
This it's starting to look pretty close to being ready. I'll take another look at it soon (the comments from today were pending for a while and I didn't realize I hadn't posted them yet)

Prior to moving this from WIP to ready, can you squash the commits into a single commit that follows our semantic versioning format? Something like:
```
feat (Hardware Support): Add Driver and Configuration Support for Razer Tararus Pro

- explanation of changes...
```

src/drivers/razer_tartarus_pro/driver.rs
dripsnek: It compares against `hidapi::DeviceInfo::interface_number ` which demands i32. Could always cast it but that may be complicating it. Originally in the match (line 203) these were direct values, but naming them makes it easier to read IMO. Happy to alter as required.

rootfs/usr/share/inputplumber/profiles/secondary_rebind.yaml
dripsnek: This file is more of a template to aid users in customizing keys bound to the secondary actuation should they use it as function of analog mode. 

How this is intended to work is that during analog mode there are no HID scancodes to speak of, we have to create them in the driver based on what keydepth is seen during sampling and send that to the composite device. As per the analog mode description this driver supports primary and secondary actuation levels. The key associated with primary actuation takes its value from the scancode that is normally emitted from basic mode but we still need a unique value signify a secondary actuation, the Tartarus itself doesn't have much to offer on that front.

Take for example Tartarus key '01'; In basic mode this is mapped to ANSI numeric row 1 and that carries over to the primary actuation. When representing the second actuation, I picked the "unused" 'b' character as that's sequentially the first gap in the USB scancode list that the Tartarus emits, 'a' being mapped to Tartarus key '12' in firmware. Codes that are and aren't emitted from firmware are documented in `razer_tartarus_pro/event.rs`

Following on, 'b' in effect is used to signal to the composite device that this is "the secondary actuation of key '01'", not a literal 'b'. Though for the purposes of presenting an output having 'b' pulling double duty as the default value suffices. The profile exists to allow remapping as 'b' from the source (the output of the analog mode handler) has a fixed definition to the composite device which is compiled in, but may not be the key the user actually wants emitted when they press it. Changing the target representation provides this customization. This logic extends to the other keys, 'g' for key '02' and so on.

This seemed the most straight forward way to implement this functionality given the existing capability types, albeit with a very convenient amount of characters to map on the Tartarus. The profile being a 1-1 mapping is a consequence of really being a template, if the users are happy with the default mapping they can go without.

rootfs/usr/share/inputplumber/devices/50-razer_tartarus_pro.yaml
```diff
@@ -0,0 +1,67 @@
...
+    unique: false
+    hidraw:
+      vendor_id: 0x1532
+      product_id: 0x244
+      interface_num: 1
+    config:
+      analogkeys:
+        primary_actuation:
+          #       01   02   03   04   05   06   07   08   09   10   11   12   13   14   15   16   17   18   19   20
+          keys: [2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0]
```
pastaq, line 44: I'm not a huge fan of this array to be honest. I think it is going to be too difficult to manually edit for anyone who didn't write the code. IMU configs are already difficult enough and that's just 9 values and they are shared among many devices. 

Perhaps you could have default values that are overridden by a config. 
Then add these fields as optional field, each as an indexed named entry that matches the capabilities for each key & activation level.

Maybe there is a better way than that as well. I'd like to get @ShadowApex opinion on it.

rootfs/usr/share/inputplumber/profiles/secondary_rebind.yaml
pastaq: IMO this sounds like it should be set in the driver and then modified by the user using a profile. Each button should have two unique capabilities, and these values should be set in the default profile. If a user wants to change them it becomes much more simple and there are multiple userspace front ends to change profiles on the fly. It's much more difficult from a UX perspective to change the capability map, and it isn't as intuitive to change "KeyB" as it is to remap TK0Secondary (or some similar convention).

src/config/mod.rs
```diff
@@ -319,6 +321,56 @@ pub struct LedConfig {
...
+    pub retrigger_reset_threshold: Option<RetriggerResetThreshold>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub retrigger_trigger_threshold: Option<RetriggerTriggerThreshold>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub continuous_retrigger: Option<ContinuousRetrigger>,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
+#[serde(rename_all = "snake_case")]
+#[allow(clippy::upper_case_acronyms)]
```
pastaq: These clippy notes seem like they don't apply?

src/drivers/razer_tartarus_pro/driver.rs
```diff
@@ -0,0 +1,623 @@
...
+    /// Tracks keystroke direction of previous report
+    key_is_upstroke: bool,
+    /// Flag to prevent early actuation of primary function during secondary function reset
+    backoff_first_func: bool,
+}
+
+/// Translates an absolute key actuation point in mm with sanitised precision to a u8 unit value
+fn convert_actuation_to_unit(value: f64) -> u8 {
+    if !(KEY_TOP_MM..=KEY_BOTTOM_MM).contains(&value) {
+        // 0 is allowed, but other values outside the key travel range are flagged as errors.
```
pastaq: This doesn't make sense to me. Only a key travel range of 0 is valid?

src/drivers/razer_tartarus_pro/driver.rs
```diff
@@ -0,0 +1,623 @@
...
+                    return self.handle_basic(buf, KeyCodes::PhantomAux, false);
+                }
+                Ok(Vec::new())
+            }
+            ENDPOINT_2 => {
+                if bytes_read == LONG_DATA_PAYLOAD {
+                    return match buf[0] {
+                        // Only manage report IDs 1 & 6. Other report types are defined in firmware
+                        // but don't appear to be used.
+                        0x1 => self.handle_basic(buf, KeyCodes::PhantomBlank, true),
```
pastaq: Can we make 0x01 and 0x06 named const or enum values as other drivers use?

pastaq (changes requested):
Left a few more comments. Hopefully @ShadowApex has some time to review this as well soon as there are many decisions that need to be made on his this is presented to the user for configuration

src/drivers/razer_tartarus_pro/driver.rs
dripsnek: The accepted range is 1.4 to 3.6mm inclusive, though where '0' is encountered this is assigned immediately as 1.4mm maps to 0 units anyway. Any other value outside the range is reported as an error with the assignment set to 0. It probably means my idea was over-complicated so I'll remove that ambiguity and also get 0 reported as an error in the next push.

rootfs/usr/share/inputplumber/devices/50-razer_tartarus_pro.yaml
dripsnek: I'm open to options, this first iteration seemed the most straight forward to me at the time as I couldn't find a way to get a profile setting to work its way through to a backing source device, it all seemed to terminate at the composite device. I'll revisit this later on in the week.

PR #527, opened 2026-02-16
fix(Target): use sync channel for config transmission

ShadowApex:
This change fixes deadlock issues the can be caused while the target waits for a response from a task that will never run.


pastaq (approved):
Works on my LGO2

PR #541, opened 2026-03-04
fix(NixOS): check nix paths for binaries in udev hiding rules

ShadowApex:
Since adding the devnode moving logic, inputplumber would fail in finding the correct binary paths for `chmod` and `mv`. This change fixes that by checking for the binaries in NixOS paths. Otherwise it will try to use the `which` built-in shell command to discover the path.

``` 
WARN  inputplumber::input::composite_device] Failed to hide device '/dev/input/event4': Os { code: 2, kind: NotFound }
```

### DIFF
--- a/src/udev/mod.rs
+++ b/src/udev/mod.rs
@@ -50,8 +50,14 @@ pub async fn hide_device(path: &str, flags: &[HideFlag]) -> Result<(), Box<dyn E
             "/bin/chmod".to_string()
         } else if Path::new("/usr/bin/chmod").exists() {
             "/usr/bin/chmod".to_string()
+        } else if Path::new("/run/current-system/sw/bin/chmod").exists() {
+            "/run/current-system/sw/bin/chmod".to_string()
         } else {
-            let output = Command::new("which").arg("chmod").output().await?;
+            let output = Command::new("sh")
+                .arg("-c")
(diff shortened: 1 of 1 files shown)


ShadowApex:
@Pandapip1 

> Wouldn't the better approach be for InputPlumber itself to traverse through the path everywhere instead of using some hardcoded paths and if that fails shelling out? This should be relatively trivial to implement ourselves, or there's [this crate](https://docs.rs/pathsearch/latest/pathsearch/) if you're okay with adding another dependency.

You're absolutely right, that is a much better solution :smile: . Since this is relatively trivial to implement (parse `PATH` and check if binary exists at each path) I would much rather just implement this logic ourselves instead of adding a new dependency. If you would like to add this an open a PR for it, we will accept it. Otherwise I will try and add this myself in the next few days. Thanks!

> Also, completely separately: why is the chmod command needed in the udev rule if we're setting MODE="000" anyway? I'm sure there's a good reason, but I'd appreciate a comment explaining that.
> 
> Thank you both on your work on InputPlumber and making sure it works on NixOS!

Thanks for your comments! In our testing, we found that just setting `MODE="000"` was not sufficient for actually setting the permissions on the device for reasons we're still not fully sure about. We suspect that other rules are overriding the mode.

PR #544, opened 2026-03-05
feat(target): add 8BitDo Ultimate 2 Wireless controller emulation

honjow:
Adds a `8bitdo-u2` target device that emulates the 8BitDo Ultimate 2 Wireless in DInput mode (VID=0x2DC8, PID=0x6012) via UHID.

Steam uses [SDL_hidapi_8bitdo.c](https://github.com/libsdl-org/SDL/blob/main/src/joystick/hidapi/SDL_hidapi_8bitdo.c) to identify and parse this device by VID/PID, reading the 34-byte input report at fixed offsets. The HID report structure and IMU coordinate mapping are derived from that source.

Supported features:
- Buttons: face, shoulder, triggers, d-pad, L4/R4/PL/PR back buttons
- Gyroscope and accelerometer (axes and signs verified on hardware)
- Rumble via HID output report ID 0x05

To support rumble routing, this also adds `OutputEvent::Rumble` and a `process_rumble` handler in the evdev gamepad source that keeps a single persistent `FF_RUMBLE` effect alive and updates its magnitude in-place.

A udev rule is included to grant hidraw access so Steam can open the virtual device without root.

Tested on MSI Claw 8 and GPD Win5 .

image

### DIFF
--- a/src/input/output_event/mod.rs
+++ b/src/input/output_event/mod.rs
@@ -17,7 +17,7 @@ pub enum OutputEvent {
     DualSense(SetStatePackedOutputData),
     SteamDeckHaptics(PackedHapticReport),
     SteamDeckRumble(PackedRumbleReport),
-    Rumble {
+    GenericRumble {
         weak_magnitude: u16,
         strong_magnitude: u16,
     },
@@ -75,7 +75,7 @@ impl OutputEvent {
                 }
             }
             OutputEvent::SteamDeckRumble(_) => vec![OutputCapability::ForceFeedback],
-            OutputEvent::Rumble { .. } => vec![OutputCapability::ForceFeedback],
+            OutputEvent::GenericRumble { .. } => vec![OutputCapability::ForceFeedback],
         }
     }
 
@@ -90,7 +90,7 @@ impl OutputEvent {
             OutputEvent::DualSense(report) => report.use_rumble_not_haptics,
             OutputEvent::SteamDeckHaptics(_) => true,
             OutputEvent::SteamDeckRumble(_) => true,
-            OutputEvent::Rumble { .. } => true,
+            OutputEvent::GenericRumble { .. } => true,
         }
     }
 }
--- a/src/input/source/hidraw/dualsense.rs
+++ b/src/input/source/hidraw/dualsense.rs
@@ -186,7 +186,15 @@ impl SourceOutputDevice for DualSenseController {
                 }
                 Ok(())
             }
-            OutputEvent::Rumble { .. } => Ok(()),
+            OutputEvent::GenericRumble {
+                weak_magnitude,
+                strong_magnitude,
+            } => {
+                let left_speed = (strong_magnitude / 256) as u8;
+                let right_speed = (weak_magnitude / 256) as u8;
+                self.driver.rumble(left_speed, right_speed)?;
+                Ok(())
+            }
         }
     }
 
--- a/src/input/source/hidraw/legos_xinput.rs
+++ b/src/input/source/hidraw/legos_xinput.rs
@@ -185,7 +185,14 @@ impl SourceOutputDevice for LegionSXInputController {
                 let r_speed = (report.right_speed.to_primitive() / 256) as u8;
                 self.driver.haptic_rumble(l_speed, r_speed)?;
             }
-            OutputEvent::Rumble { .. } => (),
+            OutputEvent::GenericRumble {
+                weak_magnitude,
+                strong_magnitude,
+            } => {
+                let left_speed = (strong_magnitude / 256) as u8;
+                let right_speed = (weak_magnitude / 256) as u8;
+                self.driver.haptic_rumble(left_speed, right_speed)?;
+            }
         }
 
         Ok(())
--- a/src/input/source/hidraw/steam_deck.rs
+++ b/src/input/source/hidraw/steam_deck.rs
@@ -237,7 +237,13 @@ impl SourceOutputDevice for DeckController {
                 let report = packed_rumble_report.pack().map_err(|e| e.to_string())?;
                 self.driver.write(&report)?;
             }
-            OutputEvent::Rumble { .. } => (),
+            OutputEvent::GenericRumble {
+                weak_magnitude,
+                strong_magnitude,
+            } => {
+                self.driver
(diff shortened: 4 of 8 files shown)


src/input/output_event/mod.rs
```diff
@@ -85,6 +90,7 @@ impl OutputEvent {
             OutputEvent::DualSense(report) => report.use_rumble_not_haptics,
             OutputEvent::SteamDeckHaptics(_) => true,
             OutputEvent::SteamDeckRumble(_) => true,
+            OutputEvent::Rumble { .. } => true,
```
pastaq: I'm not familiar with this syntax, what does `{ .. }` do?

src/input/output_event/mod.rs
```diff
@@ -17,6 +17,10 @@ pub enum OutputEvent {
     DualSense(SetStatePackedOutputData),
     SteamDeckHaptics(PackedHapticReport),
     SteamDeckRumble(PackedRumbleReport),
+    Rumble {
```
pastaq: Is there a reason UinputOutputEvent wasn't usable here?

src/input/output_event/mod.rs
honjow: > I'm not familiar with this syntax, what does `{ .. }` do?

`{ .. }` is a struct pattern wildcard in Rust. Since `OutputEvent::Rumble` is a struct variant with named fields, matching it requires { } syntax. The .. inside means "ignore all fields" — we don't need the actual magnitude values here, just need to match the variant to return its capability.

src/input/output_event/mod.rs
honjow: > Is there a reason UinputOutputEvent wasn't usable here?

`UinputOutputEvent::FFUpload` requires a full `FFEffectData` and a `Sender<Option<i16>>` channel for effect ID negotiation — none of which applies here. A UHID device already has the final magnitudes from the HID report directly, so wrapping them in `FFUpload` would just mean constructing dummy data to satisfy the type. `OutputEvent::Rumble` keeps the intent clear, following the same pattern as `SteamDeckRumble`.

src/input/target/ultimate2_wireless.rs
```diff
@@ -0,0 +1,445 @@
...
+
+        if report_id == REPORT_ID_RUMBLE && data.len() >= 3 {
+            let strong_magnitude = (data[1] as u16) << 8;
+            let weak_magnitude = (data[2] as u16) << 8;
+            log::trace!("Rumble: strong={strong_magnitude} weak={weak_magnitude}");
+            return Ok(vec![OutputEvent::Rumble {
+                weak_magnitude,
+                strong_magnitude,
+            }]);
+        }
```
ShadowApex: nit: instead of manually parsing the bytes, it would be nice if we could decode the output report using packed_struct

honjow: > nit: instead of manually parsing the bytes, it would be nice if we could decode the output report using packed_struct

Done, added `PackedRumbleOutputReport` in `hid_report.rs` to decode the output report using packed_struct instead of manual byte indexing.

pastaq: @honjow can I ask how you decoded the output report? I need to implement that for a few other source devices and could use some tips.

honjow: > @honjow can I ask how you decoded the output report? I need to implement that for a few other source devices and could use some tips.

The report layout was derived from SDL's [`SDL_hidapi_8bitdo.c`](https://github.com/libsdl-org/SDL/blob/main/src/joystick/hidapi/SDL_hidapi_8bitdo.c), specifically `HIDAPI_Driver8BitDo_RumbleJoystick`:
```c
Uint8 rumble_packet[5] = { 0x05, 0x00, 0x00, 0x00, 0x00 };
rumble_packet[1] = low_frequency_rumble >> 8;
rumble_packet[2] = high_frequency_rumble >> 8;

File: src/input/source/hidraw/dualsense.rs [hunk 4]
```diff
@@ -186,6 +186,7 @@ impl SourceOutputDevice for DualSenseController {
                 }
                 Ok(())
             }
+            OutputEvent::Rumble { .. } => Ok(()),
```
ShadowApex: The DualSense controller supports force feedback, so we should handle translating this rumble event into DualSense force feedback.

src/input/source/hidraw/legos_xinput.rs
```diff
@@ -185,6 +185,7 @@ impl SourceOutputDevice for LegionSXInputController {
                 let r_speed = (report.right_speed.to_primitive() / 256) as u8;
                 self.driver.haptic_rumble(l_speed, r_speed)?;
             }
+            OutputEvent::Rumble { .. } => (),
```
ShadowApex: This device also supports force feedback, so we should handle translating this rumble event.

src/input/source/hidraw/steam_deck.rs
```diff
@@ -237,6 +237,7 @@ impl SourceOutputDevice for DeckController {
                 let report = packed_rumble_report.pack().map_err(|e| e.to_string())?;
                 self.driver.write(&report)?;
             }
+            OutputEvent::Rumble { .. } => (),
```
ShadowApex: The Steam Deck controller supports haptic feedback, so we should translate this rumble event into haptic feedback.

src/input/output_event/mod.rs
```diff
@@ -17,6 +17,10 @@ pub enum OutputEvent {
...
+        weak_magnitude: u16,
+        strong_magnitude: u16,
+    },
```
ShadowApex: I think I would prefer this be named `GenericRumble` or even `Ultimate2Rumble` to align with the current pattern of output events.

ShadowApex (changes requested):
Some small comments, but otherwise looks good! Thanks for this!

src/input/output_event/mod.rs
honjow: Makes sense, will rename it

honjow: Done in the latest commit — renamed `Rumble` to `GenericRumble`, and also implemented rumble routing for DualSense, Legion Go, and Steam Deck source devices.

ShadowApex (approved):

PR #545, opened 2026-03-07
Feat: Support QA button for switch pro controller and pass through LT/RT correctly

Ericky14:
The switch pro controller LT/RT weren't working in Pass intercept mode due to a missing mapping, and QA button as well needed to be mapped

There is also a minor fix regarding an error loading the controller's LEDs which contain invalid characters that can't be used in dbus path

### DIFF
--- a/src/input/event/evdev.rs
+++ b/src/input/event/evdev.rs
@@ -159,6 +159,12 @@ impl EvdevEvent {
                 KeyCode::BTN_EAST => Capability::Gamepad(Gamepad::Button(GamepadButton::East)),
                 KeyCode::BTN_TL => Capability::Gamepad(Gamepad::Button(GamepadButton::LeftBumper)),
                 KeyCode::BTN_TR => Capability::Gamepad(Gamepad::Button(GamepadButton::RightBumper)),
+                KeyCode::BTN_TL2 => {
+                    Capability::Gamepad(Gamepad::Button(GamepadButton::LeftTrigger))
+                }
+                KeyCode::BTN_TR2 => {
+                    Capability::Gamepad(Gamepad::Button(GamepadButton::RightTrigger))
+                }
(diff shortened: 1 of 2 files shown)


src/input/target/xpad.rs
```diff
@@ -78,6 +78,8 @@ impl XBoxController {
         keys.insert(KeyCode::BTN_WEST);
         keys.insert(KeyCode::BTN_TL);
         keys.insert(KeyCode::BTN_TR);
+        keys.insert(KeyCode::BTN_TL2);
+        keys.insert(KeyCode::BTN_TR2);
```
ShadowApex: Is there a reason these keycodes are being added to all xbox target devices? This may break SDL mappings that expect Xbox button ordering.

Ericky14: Ah I see, I didn't know that was how it worked, I just added button mappings to where it was missing, thanks for explaining it :)

For the future, it might be helpful to explain things a bit more in comments instead of asking questions, not all contributors are working on this project 100% of the time and familiar with the codebase

ShadowApex: I typically ask to make sure there isn't a reason for the change that I might have missed and to try to get an explanation for specific changes.

Ericky14: Yea I was looking for why the trigger buttons weren't working and playing around with it, didnt realize there was a TL2/TR2, makes no sense why theres a separate enum value for same trigger buttons, controller mappings are weird
But I guess its to differentiate digital vs analog triggers? Not 100% sure yet

pastaq: > Yea I was looking for why the trigger buttons weren't working and playing around with it, didnt realize there was a TL2/TR2, makes no sense why theres a separate enum value for same trigger buttons, controller mappings are weird But I guess its to differentiate digital vs analog triggers? Not 100% sure yet

Correct, though many controllers do produce both analog and digital events as games may only support one of the two methods.

PR #546, opened 2026-03-07
feat: use vendor_id and product_id to match switch pro controllers via bluetooth as well

Ericky14:
I found that when connected via bluetooth, my switch pro controller is listed as "Pro Controller" instead of the full name, thus matching a generic device profile instead of the specific one.

This change works, but I would like to have feedback if this is the best way to go about it.. we could also match on name "*Pro Controller", but I don't know if there are other controllers which would conflict in terms of naming


Ericky14:
> Only one small change needed for this. Just need to use the `fix:` prefix in your commit message and squash these into a single commit. Thanks!

Done

ShadowApex (approved):

PR #547, opened 2026-03-07
feat: add support for output LED events and simple LEDs

Ericky14:
This PR is adding:
- simple LED driver when it only requires brightness
- supporting common output LED events so apps can set them
- allow adding virtual udev LED devices, when connecting switch pro controller over bluetooth, the LEDs become virtual devices

This PR is best tested with commits from both of these for Switch Pro Controllers:
https://github.com/ShadowBlip/InputPlumber/pull/545
https://github.com/ShadowBlip/InputPlumber/pull/546

Additonal commits added:
- Refactored multicolor implementation to be called "rgb"
- Parse led color support from kernel sys name

### DIFF
--- a/src/input/source/led/rgb.rs
+++ b/src/input/source/led/rgb.rs
@@ -286,12 +286,6 @@ impl SourceOutputDevice for LedRgb {
                 g,
                 b,
                 brightness,
-            }
-            | OutputEvent::LedMultiColor {
-                r,
-                g,
-                b,
-                brightness,
             } => {
                 self.write_color(r, g, b)?;
                 self.write_brightness(brightness)?;
--- a/src/input/source/hidraw/legos_xinput.rs
+++ b/src/input/source/hidraw/legos_xinput.rs
@@ -186,7 +186,6 @@ impl SourceOutputDevice for LegionSXInputController {
                 self.driver.haptic_rumble(l_speed, r_speed)?;
             }
             OutputEvent::LedRgb { .. } => (),
-            OutputEvent::LedMultiColor { .. } => (),
             OutputEvent::LedSingleColor { .. } => (),
         }
 
--- a/src/input/source/hidraw/steam_deck.rs
+++ b/src/input/source/hidraw/steam_deck.rs
@@ -238,7 +238,6 @@ impl SourceOutputDevice for DeckController {
                 self.driver.write(&report)?;
             }
             OutputEvent::LedRgb { .. } => (),
-            OutputEvent::LedMultiColor { .. } => (),
             OutputEvent::LedSingleColor { .. } => (),
         }
 
--- a/src/input/source/hidraw/xpad_uhid.rs
+++ b/src/input/source/hidraw/xpad_uhid.rs
@@ -160,7 +160,6 @@ impl SourceOutputDevice for XpadUhid {
             OutputEvent::SteamDeckHaptics(_packed_haptic_report) => Ok(()),
             OutputEvent::SteamDeckRumble(_packed_rumble_report) => Ok(()),
             OutputEvent::LedRgb { .. } => Ok(()),
-            OutputEvent::LedMultiColor { .. } => Ok(()),
             OutputEvent::LedSingleColor { .. } => Ok(()),
         }
     }
--- a/src/input/source/led/multicolor.rs
+++ b/src/input/source/led/multicolor.rs
@@ -1,9 +1,0 @@
-//! Multicolor LED source device.
-//!
-//! Multicolor LEDs use the kernel `multi_intensity` sysfs interface with
-//! separate color channels (e.g. `multi_index` contains `red green blue`).
-//! They share the same driver implementation as RGB LEDs since both use
-//! the same sysfs interface — the [`LedRgb`](super::rgb::LedRgb) driver
-//! handles both encodings via the `multi_index_map`.
-
-pub use super::rgb::LedRgb as LedMultiColor;
--- a/src/input/source/led/single.rs
+++ b/src/input/source/led/single.rs
@@ -102,8 +102,7 @@ impl SourceOutputDevice for LedSingleColor {
             OutputEvent::LedSingleColor { brightness } => {
                 self.write_brightness(brightness)?;
             }
-            OutputEvent::LedRgb { brightness, .. }
-            | OutputEvent::LedMultiColor { brightness, .. } => {
+            OutputEvent::LedRgb { brightness, .. } => {
                 self.write_brightness(brightness)?;
             }
             _ => {}


src/input/manager.rs
```diff
@@ -1361,14 +1361,6 @@ impl Manager {
 
             "leds" => {
                 log::debug!("LED device added: {} ({})", device.name(), device.sysname());
-
```
pastaq, line 1364: We don't currently support virtual LED targets, so this shouldn't have a negative effect on anything currently, but it does limit future additions unless we can figure out how to identify our own virtual LEDs. (As an example, aggregate the RGB of multiple devices so they can be set as a group)

rootfs/usr/share/inputplumber/devices/60-switch_pro.yaml
```diff
@@ -26,6 +26,26 @@ source_devices:
   #- group: imu
   #  evdev:
   #    name: Nintendo Co., Ltd. Pro Controller (IMU)
+  - group: led
+    udev:
+      sys_name: "*057E*2009*:green:player-1"
```
pastaq: We should be able to pattern match these and allow for multiple sources with a single entry

src/input/composite_device/mod.rs
```diff
@@ -901,6 +906,43 @@ impl CompositeDevice {
         Ok(())
     }
 
+    /// Set the active player LED on the composite device. Turns on the LED
+    /// source device whose ID contains `player-{player}` and turns off all
+    /// other LED source devices. Pass `0` to turn all LEDs off.
```
pastaq, line 912: I recall seeing this in kernel before but I don't remember if it was Nintendo, xpad, or what. Can you list/link the kernel references you used for this interface so I can make a judgement call on if this is too opinionated an implementation?

src/input/composite_device/mod.rs
```diff
@@ -901,6 +906,43 @@ impl CompositeDevice {
...
+            if !source_id.starts_with("leds://") {
+                continue;
+            }
+
+            let brightness = match &target {
+                Some(t) if source_id.contains(t.as_str()) => 255,
+                _ => 0,
+            };
+
+            let event = OutputEvent::LED {
```
pastaq: I'm concerned about this not having any sort of color/type detection. Admittedly the existing implementation is also incomplete, but if we're touching it we should fix it instead of going deeper into the hole on this. The kernel tags LEDs with their color. We should detect if the led is a single color, multiple colors, or RGB, and have led output types for each.

src/input/source/led/simple.rs
```diff
@@ -0,0 +1,118 @@
...
+    Read(std::io::Error),
+
+    #[error("Error writing value: {0}")]
+    Write(std::io::Error),
+
+    #[error("Path not found: {0}")]
+    Path(PathBuf),
+}
+
+/// Simple brightness-only LED source device. Used for LEDs that only
```
pastaq: As stated above, I'd rather call this single or solid color and make a new output type for it

src/input/source/led.rs
```diff
@@ -85,16 +101,29 @@ impl LedDevice {
...
+    /// Return the driver type for the given device info.
+    /// Uses `LedMultiColor` when `multi_intensity` sysfs exists,
+    /// otherwise falls back to the simple brightness-only driver.
     fn get_driver_type(device: &UdevDevice) -> DriverType {
         let device_name = device.sysname();
         let name = device_name.as_str();
         log::debug!("Finding driver for LED interface: {name}");
 
-        DriverType::LedMultiColor
+        let multi_intensity = PathBuf::from(device.syspath().as_str()).join("multi_intensity");
```
pastaq: Again, if we're touching this we should fix it by checking the udev properties for the color type, rename the existing multicolor to RGB and match it on the RGB color ID, and add a new multicolor that matches on the multicolor color ID. For single color LEDs we should match on any color ID that identifies it as a single color.

pastaq (changes requested):
Patch 3 can be squashed into patch 2.

src/input/composite_device/mod.rs
Ericky14: Yea
Driver: https://github.com/torvalds/linux/blob/master/drivers/hid/hid-nintendo.c#L546
Common naming: https://github.com/torvalds/linux/blob/master/include/dt-bindings/leds/common.h#L70
Basic naming docs: https://github.com/torvalds/linux/blob/master/Documentation/leds/leds-class.rst#led-device-naming

src/input/source/led.rs
Ericky14: There won't be much of a distinction between multicolor and RGB for now code-wise, correct? I don't have a multicolor LED to test with, I'll do the renaming though

src/input/source/led/simple.rs
Ericky14: I'll rename it to single

src/input/composite_device/mod.rs
Ericky14: Adding detection of color based on sys name

Ericky14:
Tested LedRgb path in the Ayaneo device with "dbus-send --system --print-reply --type=method_call --dest=org.shadowblip.InputPlumber /org/shadowblip/InputPlumber/CompositeDevice1 org.shadowblip.Input.CompositeDevice.SetLedColor byte:44 byte:44 byte:0 byte:255", it was working great
Using this driver https://github.com/ShadowBlip/ayaneo-platform

src/input/manager.rs
Ericky14: I implemented this method of differentiating between inputplumber's own virtual devices and other virtual devices, since inputplumber LEDs don't go through uhid, we can just check syspath
Let me know if this is good
https://github.com/ShadowBlip/InputPlumber/pull/547/changes/a60f0585ed2b39e572e9d09e5d79a335253bf82a

src/input/source/led.rs
pastaq: We can just rename it for now. Multicolor would be a hashmap if color string to u8, determined by the number of colors it supports. For now we can just abort if we match multicolor.

Ericky14: Removed multicolor, left just the renaming to rgb

src/dbus/interface/composite_device.rs
```diff
@@ -343,6 +344,59 @@ impl CompositeDeviceInterface {
...
+        )
+        .await?;
+        self.composite_device
+            .set_player_led(player)
+            .await
+            .map_err(|e| fdo::Error::Failed(e.to_string()))?;
+
+        Ok(())
+    }
+
```
ShadowApex, line 399: Not all composite devices will have source devices that implement LEDs. We should instead include these methods as part of either the `SourceLedInterface` or a new interface for LED control. The current implementation also assumes the presence of only a single LED on the source device(s) outside the limited use of `set_player_led`. Ideally this interface should allow a user application to control the behavior of all individual LEDs.

src/input/composite_device/command.rs
```diff
@@ -54,6 +54,7 @@ pub enum CompositeCommand {
     #[allow(dead_code)]
     UpdateSourceCapabilities(String, HashSet<Capability>),
     UpdateTargetCapabilities(String, HashSet<Capability>),
+    SetPlayerLed(u8),
```
ShadowApex, line 57: LED writes are output events, so instead of adding a dedicated command for setting LEDs, it should instead be handled how output events are handled.

src/input/composite_device/mod.rs
```diff
@@ -323,6 +324,11 @@ impl CompositeDevice {
                             log::error!("Failed to process output event: {:?}", e);
                         }
                     }
+                    CompositeCommand::SetPlayerLed(player) => {
+                        if let Err(e) = self.set_player_led(player).await {
+                            log::error!("Failed to set player LED: {:?}", e);
+                        }
+                    }
```
ShadowApex, line 331: Also here, instead of adding a new command for LED handling, this should be processed like any other output event (i.e. in `process_output_event`)

src/input/composite_device/mod.rs
```diff
@@ -901,6 +907,67 @@ impl CompositeDevice {
...
+        };
+
+        for (source_id, source) in self.source_devices.iter() {
+            // Only target LED sources
+            if !source_id.starts_with("leds://") {
+                continue;
+            }
+
+            let brightness = match &target {
+                Some(t) if source_id.contains(t.as_str()) => 255,
```
ShadowApex, line 927: This assumes that max brightness for the LED is `255`. This value should be normalized so it can be calculated based on the source device's max value.

src/input/source/led.rs
```diff
@@ -13,92 +15,191 @@ use crate::{
...
+        assert_eq!(parse_led_color_tag("ally:rgb:joystick_rings"), "rgb");
+        assert_eq!(parse_led_color_tag("go_s:rgb:joystick_rings"), "rgb");
+        assert_eq!(parse_led_color_tag("zotac:rgb:spectra_zone_0"), "rgb");
+    }
+
+    #[test]
+    fn test_parse_color_tag_single_segment() {
+        assert_eq!(parse_led_color_tag("no_colons_here"), "");
+    }
+}
```
ShadowApex, line 205: To maintain consistency with the rest of the project, unit tests should be moved to their own `*_test.rs` file.

src/input/manager.rs
```diff
@@ -1364,8 +1364,18 @@ impl Manager {
...
+                    // Allow these through, but filter out other virtual LEDs
+                    // (e.g. future InputPlumber virtual LED targets).
+                    if device.syspath().contains("/uhid/") {
+                        log::debug!(
+                            "{dev_name} ({dev_sysname}) is a virtual LED for a Bluetooth HID device. Treating as real - {dev_path}"
+                        );
+                    } else {
+                        log::debug!("{dev_name} ({dev_sysname}) is virtual, skipping consideration for {dev_path}");
+                        notify_device_added = false;
+                    }
```
ShadowApex, line 1378: This is not correct. Checking for `uhid` in the path will not distinguish between virtual hid devices created from the bluetooth service from virtual LED devices that inputplumber creates that also use `uhid`. You will need to explicitly find the hid device that this LED is associated with and check to see if the device is a paired device in bluez.

src/dbus/interface/composite_device.rs
```diff
@@ -343,6 +344,59 @@ impl CompositeDeviceInterface {
...
+            .process_output_event(event)
+            .await
+            .map_err(|e| fdo::Error::Failed(e.to_string()))?;
+
+        Ok(())
+    }
+
+    /// Set the active player LED. Turns on the LED matching `player-{player}`
+    /// and turns off all other LED source devices. Pass `0` to turn all LEDs off.
+    async fn set_player_led(
```
ShadowApex, line 380: @pastaq What are your thoughts on if we should include and maintain this feature in inputplumber versus giving this responsibility to the application/user? The app/user could instead just turn on/off LEDs of their choosing.

src/input/composite_device/mod.rs
```diff
@@ -2218,3 +2285,26 @@ impl CompositeDevice {
...
+        "yellow" => (255, 255, 0),
+        "purple" | "violet" => (128, 0, 255),
+        "cyan" => (0, 255, 255),
+        "pink" => (255, 105, 180),
+        "lime" => (191, 255, 0),
+        "infrared" => (255, 0, 0), // Nearest visible approximation
+        "rgb" => (255, 255, 255),
+        _ => (255, 255, 255),
+    }
+}
```
ShadowApex, line 2310: These kind of kernel-level implementation details should not be part of the composite device. It is the source device's responsibility for translating output events into the correct form that the LED kernel interface expects.

src/input/source/led.rs
```diff
@@ -13,92 +15,191 @@ use crate::{
...
 };
 use std::error::Error;
-/// List of available drivers
-enum DriverType {
-    LedMultiColor,
+
+/// List of available LED drivers, determined by the LED's kernel color ID.
+pub enum DriverType {
+    /// RGB LED (sysfs color tag `rgb`). Uses `multi_intensity` interface.
+    Rgb,
```
ShadowApex, line 22: This should still be `LedMultiColor` or `MutliColor` instead of `Rgb`. A multicolor LED may or may not be RGB.

src/input/source/led.rs
```diff
@@ -13,92 +15,191 @@ use crate::{
...
-        DriverType::LedMultiColor
+        // Parse the color tag from the sysfs name (second-to-last colon segment)
+        let color_tag = parse_led_color_tag(&sysname);
+        log::debug!("LED color ID: '{color_tag}'");
+
+        let driver_type = DriverType::from_color_tag(color_tag);
+        if driver_type.is_none() {
+            log::warn!("LED '{sysname}' is not yet supported, skipping");
+        }
+        driver_type
```
ShadowApex, line 149: Are you sure if using the sysname is stable enough to assume the driver type here? It might be more robust to just check for the presence of `multi_index` to determine if it is a multicolor LED.

src/input/source/led/rgb.rs
```diff
@@ -69,13 +69,15 @@ impl TryFrom<String> for ColorType {
...
         }
     }
 }
 
-/// MultiColorChassis source device implementation
-pub struct LedMultiColor {
+/// RGB/multicolor LED source device implementation. Handles LEDs that use
+/// the kernel `multi_intensity` sysfs interface, including both RGB LEDs
+/// (packed u32 color) and multicolor LEDs (separate color channels).
+pub struct LedRgb {
```
ShadowApex, line 80: This should continue to be `LedMultiColor` instead of `LedRgb`. A multicolor LED may or may not be an RGB LED.

src/input/source/hidraw/dualsense.rs
```diff
@@ -186,6 +186,8 @@ impl SourceOutputDevice for DualSenseController {
                 }
                 Ok(())
             }
+            OutputEvent::LedRgb { .. } => Ok(()),
+            OutputEvent::LedSingleColor { .. } => Ok(()),
```
ShadowApex, line 190: DualSense controllers have LEDs, so these should be implemented in the source device implementation (or a TODO added so it can be implemented later).

ShadowApex (changes requested):
I left some comments below on this implementation, but despite these comments, I don't think this is a good overall approach for how we should handle LEDs in InputPlumber. Implementing single color LED source devices is a good change, but if we want to provide an interface for LED control over DBus, there should be a specific `DbusLed` output event that different source devices can implement. Then we can expose an LED color dbus interface either on the individual source devices and/or the composite device.

The dbus interface itself needs to be able to account for multiple LEDs, either through an interface that allows you to specify the color for each individual LED or by creating a DBus object for each LED with controls for that LED.

src/input/composite_device/mod.rs
Ericky14: I had no idea max brightness went over 255

src/input/source/led/rgb.rs
Ericky14: Isn't a multicolor LED fundamentally different than an rgb LED? Not sure if there will be any differences in implementation

src/input/composite_device/mod.rs
Ericky14: I guess this one is up to opinion, if input plumber is handling LEDs, it should take care of the heavy lifting for consuming apps, at a higher level users shouldn't have to dig into the kernel code to see what possible values there are for example IMO

src/input/composite_device/mod.rs
ShadowApex: The kernel interface provides a `max_brightness` attribute that defines the maximum brightness value. E.g.

```bash
$ cat /sys/class/leds/chromeos:multicolor:charging/max_brightness
100
```

src/input/source/led/rgb.rs
ShadowApex: An RGB LED is a subset of a multicolor LED. All RGB LEDs are mutlicolor LEDs, but not all multicolor LEDs are RGB LEDs. An RGB LED is the most common class of multicolor LEDs, but there are other multicolor LEDs. E.g. amber, red, white.

src/input/composite_device/mod.rs
ShadowApex: I think you're misunderstanding my comment. InputPlumber itself already provides a high-level API for users to use. What I'm saying here is that this code should not be in the composite device. It should instead exist inside the source device implementation code (i.e. `input/source/led/*.rs`) that controls LEDs via the kernel interface. Using the kernel interface to control LEDs is not the only way we can control LEDs. For example, in the DualSense source device implementation, we do _not_ use the kernel LED interface and instead control the gamepad's LEDs using HID packets.

src/input/manager.rs
Ericky14: Do you have docs or can you point me to the code that suggests inputplumber has `uhid` in its path?
I believe that is set by the kernel and not something you can fake

Input plumber virtual LEDs would go into /sys/devices/virtual probably like the other current virtual devices

ShadowApex: https://github.com/ShadowBlip/InputPlumber/blob/main/src/input/target/dualsense.rs#L147

src/input/source/led.rs
Ericky14: I believe so since its defined at the kernel level, but I agree your suggestion is more robust

src/input/source/led.rs
Ericky14: @pastaq What do you think? Should we keep name Rgb or MultiColor
I guess the suggestion here is to rename back to MultiColor and have a field is_rgb?

src/input/composite_device/mod.rs
Ericky14: Got it

ShadowApex:
> Initially my goal was to only implement the player LEDs, but then it became a bigger change regarding rgb/multicolor leds as well, but if we want to do this the right way then.. the remaining changes would be:
> 
>     * Create `DbusLed`  output event for source devices to implement
> 
>     * Expose an LED color control interface for both source and composite devices
> 
>     * In addition to above, potentially add one LED dbus object per LED device so consumer can control individual LEDs
> 
> 
> If this is the agreed architecture I can refactor to that, let me know @ShadowApex

That is the general sentiment, but I would first propose how you would like to approach the dbus interface to get agreement before spending time writing the implementation.

ShadowApex:
> This is what I would propose:
> 
> * Per LED dbus objects (already exists)
> 
> ```
> /org/shadowblip/InputPlumber/devices/source/input48:rgb:indicator
> /org/shadowblip/InputPlumber/devices/source/input48:white:player-1
> /org/shadowblip/InputPlumber/devices/source/input48:white:player-2
> ```
> 
> * Interface `org.shadowblip.Input.LED`
> 
> 
> ```
> Properties (read-only):
>   Name         : String    — kernel sysfs name
>   Type           : String    — "multicolor" | "single_color"
>   ColorTag   : String     — kernel color tag
>   Function   : String     — kernel function tag
> 
> Properties (read-write):
>   Brightness   : u8        — 0-max-brightness
>   Color             : (u8,u8,u8) — (r,g,b), only useful for multicolor
> 
> Methods:
>   SetColor(r: u8, g: u8, b: u8, brightness: u8)
>   SetBrightness(brightness: u8)
> ```
> 
> * New composite device interface change:
> 
> 
> ```
> Properties:
>   LEDs : Vec<ObjectPath>
> ```
> 
> * Generic LED Output Event, requiring specific devices like DualSense to produce a generic output event instead of `OutputEvent::DualSense(SetStatePackedOutputData)` which is too specific.
> 
> 
> ```
> pub enum OutputEvent {
>     Led(LedOutputEvent),
> }
> 
> pub struct LedOutputEvent {
>     pub target: Option<LedTarget>,
>     pub color: Option<(u8, u8, u8)>,
>     pub brightness: Option<u8>,
> }
> 
> pub enum LedTarget {
>     Function(String),
>     Type(LedType),  // MultiColor or SingleColor
>     All,
> }
> ```
> 
> Solution for virtual device detection:
> 
> * Short circuit virtual check by first checking if HID_ID is 0005 to match bluetooth only (Unless there can be physical virtual devices?)
> 
> * Use `HID_UNIQ` tree property to match against bluez bluetooth device (cache bluetooth device list so we don't query it each time)

Is this your proposal or an AI-generated proposal?

The interface listed here is effectively a 1-to-1 wrapper around the sysfs interface and is limited to only RGB LEDs. What about source devices that do not use the kernel LED interface?

This change should also not remove all the existing output events. There needs to be logic in place for converting different types of output events into different forms. It is up to the source device implementations to be able to convert the different LED output events into the correct form.

You also probably don't need to cache the bluetooth listings.

Ericky14:
> > This is what I would propose:
> > 
> > * Per LED dbus objects (already exists)
> > 
> > ```
> > /org/shadowblip/InputPlumber/devices/source/input48:rgb:indicator
> > /org/shadowblip/InputPlumber/devices/source/input48:white:player-1
> > /org/shadowblip/InputPlumber/devices/source/input48:white:player-2
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * Interface `org.shadowblip.Input.LED`
> > 
> > ```
> > Properties (read-only):
> >   Name         : String    — kernel sysfs name
> >   Type           : String    — "multicolor" | "single_color"
> >   ColorTag   : String     — kernel color tag
> >   Function   : String     — kernel function tag
> > 
> > Properties (read-write):
> >   Brightness   : u8        — 0-max-brightness
> >   Color             : (u8,u8,u8) — (r,g,b), only useful for multicolor
> > 
> > Methods:
> >   SetColor(r: u8, g: u8, b: u8, brightness: u8)
> >   SetBrightness(brightness: u8)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * New composite device interface change:
> > 
> > ```
> > Properties:
> >   LEDs : Vec<ObjectPath>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * Generic LED Output Event, requiring specific devices like DualSense to produce a generic output event instead of `OutputEvent::DualSense(SetStatePackedOutputData)` which is too specific.
> > 
> > ```
> > pub enum OutputEvent {
> >     Led(LedOutputEvent),
> > }
> > 
> > pub struct LedOutputEvent {
> >     pub target: Option<LedTarget>,
> >     pub color: Option<(u8, u8, u8)>,
> >     pub brightness: Option<u8>,
> > }
> > 
> > pub enum LedTarget {
> >     Function(String),
> >     Type(LedType),  // MultiColor or SingleColor
> >     All,
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Solution for virtual device detection:
> > 
> > * Short circuit virtual check by first checking if HID_ID is 0005 to match bluetooth only (Unless there can be physical virtual devices?)
> > * Use `HID_UNIQ` tree property to match against bluez bluetooth device (cache bluetooth device list so we don't query it each time)
> 
> Is this your proposal or an AI-generated proposal?
> 
> The interface listed here is effectively a 1-to-1 wrapper around the sysfs interface and is limited to only RGB LEDs. What about source devices that do not use the kernel LED interface?
> 
> This change should also not remove all the existing output events. There needs to be logic in place for converting different types of output events into different forms. It is up to the source device implementations to be able to convert the different LED output events into the correct form.
> 
> You also probably don't need to cache the bluetooth listings.

Apparentally no one can write anything without them being AI generated :sweat: 
Was the initial LED code in input plumber AI generated as well? Since it seems it wasn't implemented perfectly and needs a refactor.

Alright I'll let you handle it, you can close the PR. I don't want to use my personal time to discuss whether something is AI generated or not instead of solving the problem. Crazy how politics got into engineering.

But to answer your question, no it's not, I typed that after doing some research and thinking about it on the side while I am slammed with work, working overtime (already past 7pm here).

Also maybe you didn't understand my proposal, but it works just fine :man_shrugging:
I don't see why you can't just merge the events and implement a helper to map dual sense output event report to a generic led output event, having specific code all over the place is how your codebase becomes like gamescope, workarounds/hacks everywhere

PR #554, opened 2026-03-16
Add support for GPD Win Mini 2024 (8840U / G1617-01)

tctinh:
NOTE: This PR add additional support for 2024 model while keep the same compability with older model which Inputplumber already supported. Some GPD Win Mini 2024 can also name G1617-01 duplicated with 2023/2022 version

Description of Changes:
This PR introduces full hardware support for the 2024 iteration of the GPD Win Mini (8840U), fixing issues where the controller was not properly grabbed and the L4/R4 back buttons were entirely unresponsive due to changed hardware hex codes.

1. Rust Driver Updates (src/drivers/gpd_win_mini/macro_keyboard_driver.rs)

The Bug: The 2024 model (G1617-02) uses completely different hex codes for the L4 and R4 macro keys than the 2023 model.

The Fix: * Refactored the single L4_CODE (0x46) and R4_CODE (0x48) constants into arrays to support both generations.

Added the discovered 2024 hardware codes: L4 = 0x6f, R4 = 0x70.

Created a report_has_any_key helper function to check the report against the new code arrays.

Updated the press state checks so the software debounce logic now successfully catches and holds the macro pulses for both the 2023 and 2024 models.

2. YAML Profile Updates (rootfs/usr/share/inputplumber/devices/50-gpd_winmini.yaml)

DMI Matching: Added G1617-02 (the 2024 model) to the product_name match list.

PCI Routing: Updated the phys_path definitions for the Xbox pad and Windows Mouse to use an array {usb-0000:63...,usb-0000:c3...}. The 2024 AMD motherboard routes through c3 instead of the 2023's 63, which previously caused InputPlumber to drop the physical controller and crash when steamos-manager requested the ds5-edge profile.

3. Notes for the Maintainers (Optional / Discussion):

Touchpad Handling: We found that completely omitting the group: mouse blocks from the YAML allows Linux to handle the touchpad natively. When InputPlumber grabs it and virtualizes it into the ds5-edge profile, Steam's default behavior hijacks the PS5 touchpad and turns it into a Right Joystick, breaking native desktop navigation.

R4 Capability Map: We locally overrode gpd_type3.yaml to change R4 from QuickAccess to RightPaddle1. The default QuickAccess mapping prevents users from mapping R4 as a standard gameplay paddle in Steam's Controller Settings.

### DIFF
--- a/rootfs/usr/share/inputplumber/devices/50-gpd_winmini.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-gpd_winmini.yaml
@@ -25,7 +25,7 @@ source_devices:
   - group: gamepad
     evdev:
       name: "Microsoft X-Box 360 pad"
-      phys_path: usb-0000:63:00.3-5/input0
+      phys_path: "{usb-0000:63:00.3-5/input0,usb-0000:c3:00.3-5/input0}"
       handler: event*
   - group: keyboard
     hidraw:
--- a/rootfs/usr/share/inputplumber/devices/50-gpd_winmini.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-gpd_winmini.yaml
@@ -11,49 +11,44 @@ name: GPD WinMini
 # Only allow a single source device per composite device of this type.
 single_source: false
 
-# Matches: We explicitly define both the 2023 (7840U) and 2024 (8840U) 
-# DMI product names so this profile universally applies to both generations.
+# Only use this profile if *any* of the given matches matches. If this list is
+# empty, then the source devices will *always* be checked.
+# /sys/class/dmi/id/product_name
 matches:
   - dmi_data:
       product_name: G1617-01
       sys_vendor: GPD
-  - dmi_data:
-      product_name: G1617-02
-      sys_vendor: GPD
 
-# Source Devices: The raw hardware inputs InputPlumber will grab and manage.
+# One or more source devices to combine into a single virtual device. The events
+# from these devices will be watched and translated according to the key map.
 source_devices:
   - group: gamepad
     evdev:
       name: "Microsoft X-Box 360 pad"
-      # We use an array {63,c3} to support the different internal USB PCI bus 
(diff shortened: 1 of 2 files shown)


tctinh:
Hi @justinweiss thanks for your review
After recheck my GPD 2024 model is G1617-01 not G1617-02 that is mistaken on my part
On mouse block part, I think better we should keep your mapping leave it to local user to change if they found the touchpad mapped wrong the reason it's funky on my side because I mapped the gamepad to DS5-Edge instead of Steam Deck Controller which wrongly map the touchpad into joysticks
about L4/R4 issue let me do more testing, currently wiped the window boot so cannot access software to remap the L4/R4 behavior, do you know anyway to do the mapping on linux perhap install HHD map there then uninstall it.

I think the best way forward for this PR should be keep the default config of gpd win mini as it is right now and add       input0,usb-0000:c3:00.3-5/input0 as on phys path, I will test and make change to this PR

justinweiss:
I think you can use pyWinControls to remap from Linux, but I haven't tried it myself: https://github.com/pelrun/pyWinControls

pastaq (approved):

ShadowApex (approved):

Ryklyyn:
Even though GPD Win Mini 2024 edition had G1617-01 name listed in the Bios page, driver wise it identifies itself as G1617-02, current yaml have the product_name as G1617-01 resulting controller non-detection outside of Steam until manual correction has been made.

G1617-02 is the correct code name for the 2024 device.

pastaq:
> Even though GPD Win Mini 2024 edition had G1617-01 name listed in the Bios page, driver wise it identifies itself as G1617-02, current yaml have the product_name as G1617-01 resulting controller non-detection outside of Steam until manual correction has been made.
> 
> G1617-02 is the correct code name for the 2024 device.

Please make a new issue on this instead of posting on an old PR.

When you do, include a copy of `udevadm info -a /sys/class/dmi/id`

PR #558, opened 2026-03-18
feat(gpd_win5): add HID driver for back buttons on new firmware

honjow:
GPD Win 5 new controler firmware (since 1.09) changed the back buttons (mode switch, L4, R4) from
keyboard events (F13/F14/F15) to a vendor HID report.
Device: VID 0x2f24, PID 0x0137, Interface 0, Usage Page 0xFF00
Report format (12 bytes):
- BUF[8]: mode switch (0x68 = pressed)
- BUF[9]: left back (0x69 = pressed)  
- BUF[10]: right back (0x6a = pressed)
Tested on actual hardware with updated firmware.

### DIFF
--- a/src/drivers/gpd_win5/driver.rs
+++ b/src/drivers/gpd_win5/driver.rs
@@ -19,9 +19,6 @@ const HID_TIMEOUT: i32 = 10;
 
 pub struct GpdWin5ButtonDriver {
     device: HidDevice,
-    mode_pressed: bool,
-    l4_pressed: bool,
-    r4_pressed: bool,
     state: Option<GpdWin5ButtonReport>,
 }
 
@@ -40,9 +37,6 @@ impl GpdWin5ButtonDriver {
 
         Ok(Self {
             device,
-            mode_pressed: false,
-            l4_pressed: false,
-            r4_pressed: false,
             state: None,
         })
     }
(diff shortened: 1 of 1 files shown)


src/drivers/gpd_win5/driver.rs
```diff
@@ -0,0 +1,124 @@
...
+pub const PID: u16 = 0x0137;
+pub const IID: i32 = 0x00;
+
+const HID_TIMEOUT: i32 = 10;
+
+pub struct GpdWin5ButtonDriver {
+    device: HidDevice,
+    mode_pressed: bool,
+    l4_pressed: bool,
+    r4_pressed: bool,
```
pastaq: The same data is stored in the state, so these bool's are superfluous.

src/drivers/gpd_win5/driver.rs
```diff
@@ -0,0 +1,124 @@
...
+        let report = GpdWin5ButtonReport::unpack(&buf)?;
+        let old_state = self.state;
+        self.state = Some(report);
+
+        // Skip the first report (no previous state to compare)
+        if old_state.is_none() {
+            return Ok(Vec::new());
+        }
+
+        Ok(self.translate_events(&report))
```
pastaq: You are storing report as self.state. Send in old_state instead since you can access this globally.

src/drivers/gpd_win5/driver.rs
```diff
@@ -0,0 +1,124 @@
...
+
+        // Right back button (BUF[10])
+        let r4_now = report.right_back != 0;
+        if r4_now != self.r4_pressed {
+            log::trace!("Right back button: {}", if r4_now { "pressed" } else { "released" });
+            events.push(Event::GamepadButton(GamepadButtonEvent::R4(
+                BinaryInput { pressed: r4_now },
+            )));
+            self.r4_pressed = r4_now;
+        }
```
pastaq: for these, you can use the pattern if self.state.mode_switch != old_state.mode_switch ...

honjow: > for these, you can use the pattern if self.state.mode_switch != old_state.mode_switch ...

Thanks, I've made the changes as suggested

pastaq (approved):

pastaq:
It seems better & a move towards where we originally wanted HID to be. I think it will get there.

PR #563, opened 2026-03-24
feat(CapabilityMap): add generic hidraw button driver

honjow:
## Summary

Add a generic hidraw button driver that uses capability map YAML configs to define button mappings, following the same pattern as existing evdev capability maps. This removes the need to write per-device Rust driver modules for simple HID button devices.

Includes GPD Win 5 HID button support (new firmware) as the first use case. **Supersedes #558**, which implements the same functionality with a dedicated driver.

## Approach

The existing evdev source already uses a declarative approach — devices like MSI Claw and GPD define button mappings in YAML capability maps without any device-specific Rust code. This PR extends that pattern to hidraw by:

1. Extending `HidrawConfig` with flexible detection modes (non-zero byte, exact value match, specific bit check)
2. Adding `HidrawEventTranslator` to translate raw HID reports based on the capability map
3. Falling back to the generic driver automatically when `DriverType::Unknown` but a hidraw capability map is present

Adding support for a new simple HID button device now only requires a YAML capability map file — no Rust code or recompilation needed.

## Test plan
- [x] Verify GPD Win 5 back buttons (mode switch, L4, R4) work with new firmware
- [ ] Verify existing hidraw devices (DualSense, Legion Go, etc.) are unaffected

### DIFF
--- a/rootfs/usr/share/inputplumber/capability_maps/gpd_v2_hid1.yaml
+++ b/rootfs/usr/share/inputplumber/capability_maps/gpd_v2_hid1.yaml
@@ -1,7 +1,14 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ShadowBlip/InputPlumber/main/rootfs/usr/share/inputplumber/schema/capability_map_v2.json
+# Schema version number
 version: 2
+
+# The type of configuration schema
 kind: CapabilityMap
-name: GPD Win 5 HID Buttons
-id: gpd_win5_hid
+
+# Name for the device event map
+name: GPD HID Type 1
+
+id: gpd_v2_hid1
 
 # GPD Win 5 vendor HID report (VID 0x2f24, PID 0x0137, Usage Page 0xFF00)
 # Idle:  01 a5 00 5a ff 00 01 09 00 00 00 00
--- a/src/config/capability_map/hidraw.rs
+++ b/src/config/capability_map/hidraw.rs
@@ -6,8 +6,13 @@ use serde::{Deserialize, Serialize};
 #[derive(Debug, Deserialize, Serialize, Clone, JsonSchema, PartialEq)]
 #[serde(rename_all = "snake_case")]
 pub struct HidrawConfig {
-    pub report_id: u32,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub report_id: Option<u8>,
     pub input_type: String,
     pub byte_start: u64,
-    pub bit_offset: u8,
+    /// Bit position within the byte (LSB=0).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub bit_offset: Option<u8>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub value: Option<u8>,
 }
--- a/src/input/event/hidraw/translator.rs
+++ b/src/input/event/hidraw/translator.rs
@@ -1,0 +1,132 @@
+use crate::{
+    config::capability_map::CapabilityMapConfigV2,
+    input::{
+        capability::Capability,
+        event::{native::NativeEvent, value::InputValue},
+    },
+};
+
+#[derive(Debug, Clone)]
+struct HidrawButtonMapping {
+    report_id: Option<u8>,
+    byte_index: usize,
+    detection: DetectionMode,
+    capability: Capability,
+}
+
+#[derive(Debug, Clone)]
+enum DetectionMode {
+    NonZero,
+    Value(u8),
+    /// Bit position (LSB=0)
+    Bit(u8),
+}
+
+/// Translates raw HID reports into [NativeEvent]s using a capability map.
+#[derive(Debug)]
+pub struct HidrawEventTranslator {
+    mappings: Vec<HidrawButtonMapping>,
+    state: Vec<bool>,
+}
+
+impl HidrawEventTranslator {
+    /// Create a new translator from a V2 capability map.
+    pub fn new(capability_map: &CapabilityMapConfigV2) -> Self {
+        let mut mappings = Vec::new();
+
+        for mapping in capability_map.mapping.iter() {
+            for source in mapping.source_events.iter() {
+                let Some(hidraw) = source.hidraw.as_ref() else {
+                    continue;
+                };
+
+                if hidraw.input_type != "button" {
+                    log::warn!(
+                        "Unsupported hidraw input_type '{}' in mapping '{}', skipping",
+                        hidraw.input_type,
+                        mapping.name,
+                    );
+                    continue;
+                }
+
+                let cap: Capability = mapping.target_event.clone().into();
+                if cap == Capability::NotImplemented {
+                    log::warn!(
+                        "Unresolved target capability in mapping '{}', skipping",
+                        mapping.name,
+                    );
+                    continue;
+                }
+
+                let detection = if let Some(value) = hidraw.value {
+                    DetectionMode::Value(value)
+                } else if let Some(bit) = hidraw.bit_offset {
+                    DetectionMode::Bit(bit)
+                } else {
+                    DetectionMode::NonZero
+                };
+
+                mappings.push(HidrawButtonMapping {
+                    report_id: hidraw.report_id,
+                    byte_index: hidraw.byte_start as usize,
+                    detection,
+                    capability: cap,
+                });
+            }
+        }
+
+        let state = vec![false; mappings.len()];
+        Self { mappings, state }
+    }
+
+    pub fn has_mappings(&self) -> bool {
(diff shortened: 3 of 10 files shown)


rootfs/usr/share/inputplumber/capability_maps/gpd_v2_hid1.yaml
pastaq, line 1: This file name doesn't match our current convention. gpd_type5 would be the next logical pattern. TBS, perhaps it would be good to have more descriptive names that include the family, capability map version, and subsystem.

i.e. gpd_type1 would become v1_gpd_evdev1 and this could be v2_gpd_hid1

@ShadowApex thoughts?

rootfs/usr/share/inputplumber/capability_maps/gpd_v2_hid1.yaml
```diff
@@ -0,0 +1,37 @@
+version: 2
+kind: CapabilityMap
+name: GPD Win 5 HID Buttons
+id: gpd_win5_hid
+
+# GPD Win 5 vendor HID report (VID 0x2f24, PID 0x0137, Usage Page 0xFF00)
+# Idle:  01 a5 00 5a ff 00 01 09 00 00 00 00
```
pastaq, line 14: Do we know what these first 8 bytes are?

src/input/event/hidraw/translator.rs
```diff
@@ -0,0 +1,132 @@
...
+    state: Vec<bool>,
+}
+
+impl HidrawEventTranslator {
+    /// Create a new translator from a V2 capability map.
+    pub fn new(capability_map: &CapabilityMapConfigV2) -> Self {
+        let mut mappings = Vec::new();
+
+        for mapping in capability_map.mapping.iter() {
+            for source in mapping.source_events.iter() {
```
pastaq, line 38: Can we call this source_event? Its a little ambiguous since we also have source_device elsewhere.

src/input/event/hidraw/translator.rs
```diff
@@ -0,0 +1,132 @@
...
+
+        for (idx, mapping) in self.mappings.iter().enumerate() {
+            if let Some(expected_id) = mapping.report_id {
+                if report.first().copied() != Some(expected_id) {
+                    continue;
+                }
+            }
+
+            if mapping.byte_index >= report.len() {
+                continue;
```
pastaq, line 108: This is be very spammy, but we should add a warn here so people know why their mapping is failing.

src/input/event/hidraw/translator.rs
```diff
@@ -0,0 +1,132 @@
...
+                    capability: cap,
+                });
+            }
+        }
+
+        let state = vec![false; mappings.len()];
+        Self { mappings, state }
+    }
+
+    pub fn has_mappings(&self) -> bool {
```
pastaq: I don't know why we need both this and has_hid_mappings? It already filters entries to self.mappings based on hidraw events, so its more performant to just check if self.mappings is empty in both cases since the work is done. In evdev we use has_translation. Perhaps we should rename that one to has_evdev_translation to disambiguate, and call this one has_hid_translation and also remove the below one?

src/input/source/hidraw.rs
```diff
@@ -269,7 +287,25 @@ impl HidRawDevice {
         let driver_type = HidRawDevice::get_driver_type(&device_info, is_blocked);
 
         match driver_type {
-            DriverType::Unknown => Err("No driver for hidraw interface found".into()),
+            DriverType::Unknown => {
+                if let Some(cap_map) = Self::load_hidraw_capability_map(&conf) {
```
pastaq: Instead of checking this and running an iter() again when there are mappings, we could just load the GenericHidrawButtons and have it return an error if there are no translations in the map.

pastaq (changes requested):
Much needed feature, thanks for working on this. I want to discuss a few things, so before you change anything below lets get to a consensus.

rootfs/usr/share/inputplumber/capability_maps/gpd_v2_hid1.yaml
honjow: Cannot be determined; GPD has not provided any relevant documentation for it

src/input/event/hidraw/translator.rs
honjow: Of course

src/input/event/hidraw/translator.rs
honjow: Makes sense, will add a warn log here.

src/input/event/hidraw/translator.rs
honjow: Yeah, no reason to keep both. Will rename to has_hid_translation to match the evdev pattern and drop the static one.

src/input/source/hidraw.rs
honjow: Right, that simplifies things. Will just pass the capability map to GenericHidrawButtons and let it handle the validation.

rootfs/usr/share/inputplumber/capability_maps/gpd_v2_hid1.yaml
ShadowApex: For sorting, I would prefer something like `gpd_v1_evdev1`, `gpd_v2_hid1`, etc. But yes, I agree with having more descriptive filenames.

honjow: So I'll rename it to `gpd_v2_hid1`? @pastaq

honjow: Rename committed

pastaq (commented):
Thanks for making those changes. I'll test it this week and verify there are no strange effects in other hid devices.

pastaq (approved):
Tested on Legion Go S and Legion Go 2 and experienced 0 conflicts.

src/config/capability_map/hidraw.rs
```diff
@@ -6,8 +6,13 @@ use serde::{Deserialize, Serialize};
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub report_id: Option<u8>,
     pub input_type: String,
     pub byte_start: u64,
-    pub bit_offset: u8,
+    /// Bit position within the byte (LSB=0).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub bit_offset: Option<u8>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub value: Option<u8>,
```
ShadowApex, line 17: What is the purpose of this field in decoding hidraw data?

src/input/event/hidraw/translator.rs
```diff
@@ -0,0 +1,129 @@
...
+                };
+
+                if hidraw.input_type != "button" {
+                    log::warn!(
+                        "Unsupported hidraw input_type '{}' in mapping '{}', skipping",
+                        hidraw.input_type,
+                        mapping.name,
+                    );
+                    continue;
+                }
```
ShadowApex, line 50: Since this implementation only supports simple button decoding, it would be nice to add a `TODO` somewhere so we can ensure that we finish implementing decoding other types of input (e.g. axis info, gyro, etc.)

src/input/source/hidraw/generic_buttons.rs
```diff
@@ -0,0 +1,80 @@
...
+        source::{InputError, SourceInputDevice, SourceOutputDevice},
+    },
+    udev::device::UdevDevice,
+};
+
+const HID_TIMEOUT: i32 = 10;
+const READ_BUF_SIZE: usize = 64;
+
+/// Generic hidraw button source device driven by a capability map.
+pub struct GenericHidrawButtons {
```
ShadowApex, line 19: Since this will eventually also support non-button input, we should probably call this simply `GenericHidraw`.

src/input/source/hidraw.rs
```diff
@@ -269,7 +286,21 @@ impl HidRawDevice {
...
+                        SourceDriver::new(composite_device, device, device_info.into(), conf);
+                    return Ok(Self::GenericButtons(source_device));
+                }
+
+                let vid = device_info.id_vendor();
+                let pid = device_info.id_product();
+                Err(format!(
+                    "No driver for hidraw interface found. VID: {vid:#06x}, PID: {pid:#06x}"
+                )
+                .into())
```
ShadowApex, line 302: It would be good to further specify in the error here that no driver _or_ capability map was found for this device.

ShadowApex (changes requested):
Thank you for this change! This is something we have wanted for a while. I left a few short comments, but we should be able to accept this with minimal changes. Thanks again!

PR #567, opened 2026-03-27
feat(Hardware Support): add OXP HID driver for OneXPlayer X1 and APEX

honjow:
Add a vendor HID driver for OneXPlayer devices (VID `1a86`, PID `fe00`, interface 2) to support back paddles (M1/M2), Keyboard, and Guide buttons through the proprietary report mode.

The HID protocol reverse engineering was largely based on work by @srsholmes in [srsholmes/onexplayer-apex-bazzite-fixes#8](https://github.com/srsholmes/onexplayer-apex-bazzite-fixes/pull/8), where the B4 key mapping, B2 report mode, and OXP key encoding were captured and documented via HID sniffing on Windows. Thanks to srsholmes for the thorough protocol analysis.

## How it works

The OXP MCU exposes a HID interface that can be configured and polled over hidraw:

- **B4** configures button-to-keycode mappings (M1→F14, M2→F13). Note that our input parser reads the button's own identifier from the report, not the configured keycode — the B4 mapping only ensures that these buttons produce HID events independently of the Xbox gamepad.
- **B2** ENABLE→DISABLE cycle activates the vendor report mode. This appears to be required only on the APEX; on X1 Mini the report mode works without it, but the cycle is harmless there.
- **B3** sets vibration intensity to max (5). The MCU does not persist this across reboots, so it needs to be sent on every initialization. This appears to be necessary only on X1 Mini — the APEX works without it, but the command is harmless on devices that don't require it.

## Suspend/resume handling

On X1 Mini, after s2idle resume the MCU resets its volatile settings (B3, B4) and emits a `B8` status packet with `byte3=0xFE` roughly 6 seconds after wake. The driver detects this and schedules a full re-initialization on the next poll cycle. This was observed consistently on X1 Mini (CH32 MCU) but may not occur on all models — the detection is safe: if the device never sends `B8 0xFE`, nothing happens, and the re-init sequence is idempotent. Whether this should remain default behavior or be behind a per-device flag is open for discussion.

## Device support

- **X1 series** (X1 A / X1 i / X1 mini): device config, IMU mount matrix correction, external/docked controller support, HID Keyboard→QuickAccess mapping. The IMU mount matrix was only verified on X1 Mini — it may need adjustment for other X1 variants.
- **APEX**: device config with back paddle swap (physical left/right are reversed compared to X1), and special button mappings (Turbo, KB, Orange).

## Testing status

Tested on X1 Mini and APEX — basic functionality (back paddles, special buttons, vibration) works correctly on both. APEX gyroscope/IMU configuration has not been verified yet and may need a follow-up.


src/drivers/oxp_hid/driver.rs
```diff
@@ -0,0 +1,308 @@
...
+//   B8 [... fe ... 00 ...]   ← byte3=0xFE, only appears during MCU init
+//   B8 [... fd 04 ... 04 ...]
+// The 0xFE packet signals that the MCU has completed its own re-initialization
+// and has reset volatile settings (B3 vibration intensity, B4 button mappings).
+// Without re-initialization, vibration stops working after resume.
+//
+// This detection is safe for devices that do not emit B8 0xFE (e.g. Apex):
+// the condition simply never triggers, and re-initialization is idempotent.
+const CMD_MCU_STATUS: u8 = 0xB8;
+const MCU_INIT_COMPLETE: u8 = 0xFE;
```
pastaq: These will be covered by a kernel driver.

src/drivers/oxp_hid/driver.rs
```diff
@@ -0,0 +1,308 @@
...
+        log::debug!("OXP HID: B3 vibration sent ({w5}B) — intensity=5");
+        std::thread::sleep(std::time::Duration::from_millis(50));
+        self.drain_responses("B3", &mut drain_buf)?;
+
+        log::info!(
+            "OXP HID: initialization complete — B4({w1}+{w2}B) B2({w3}+{w4}B) B3({w5}B)"
+        );
+        self.initialized = true;
+        Ok(())
+    }
```
pastaq: Same for these.

src/drivers/oxp_hid/driver.rs
```diff
@@ -0,0 +1,308 @@
...
+
+    /// Poll the device and read input reports
+    pub fn poll(&mut self) -> Result<Vec<Event>, Box<dyn Error + Send + Sync>> {
+        if !self.initialized {
+            self.initialize()?;
```
pastaq: These can be removed as well.

src/drivers/oxp_hid/driver.rs
```diff
@@ -0,0 +1,308 @@
...
+                );
+                self.initialized = false;
+            } else {
+                log::debug!(
+                    "OXP HID: non-B2 packet CID=0x{cid:02X}: [{}]",
+                    hex_prefix(&buf, 16)
+                );
+            }
+            return Ok(Vec::new());
+        }
```
pastaq: This check can just return here instead.

src/drivers/oxp_hid/driver.rs
```diff
@@ -0,0 +1,308 @@
...
+
+        let pkt_type = buf[3];
+        let flag = buf[5];
+        let btn = buf[6];
+        let func_code = buf[7];
+        let pressed = buf[12] == 1;
```
pastaq: This can all be handled automatically if you use PackedStruct. For the values with known ranges (i.e. button) you can assign the bytes to an enum. For simple boolean logic, you can assign the byte to bool and it will handle this conversion.

src/drivers/oxp_hid/driver.rs
```diff
@@ -0,0 +1,308 @@
...
+
+        let event = match btn {
+            BTN_M1 => ButtonEvent::M1(BinaryInput { pressed }),
+            BTN_M2 => ButtonEvent::M2(BinaryInput { pressed }),
+            BTN_KEYBOARD => ButtonEvent::Keyboard(BinaryInput { pressed }),
+            BTN_GUIDE => ButtonEvent::Guide(BinaryInput { pressed }),
+            0x00 => return Ok(Vec::new()),
+            _ => {
```
pastaq: Combine this with the 0x00 case and drop the unnecessary warnings

src/drivers/oxp_hid/driver.rs
```diff
@@ -0,0 +1,308 @@
...
+            }
+        };
+
+        // Debounce
+        if let Some(prev) = self.btn_state.get_mut(btn as usize) {
+            if *prev == pressed {
+                return Ok(Vec::new());
+            }
+            *prev = pressed;
+        }
```
pastaq: You can do this first and avoid generating events to then throw away.

pastaq (changes requested):
The extra RE has been useful for adding some of the missing bits to the WIP kernel driver, so thank you for those efforts. The driver will be added to the OGC kernel and the Valve kernel, so all major gaming distros will be able to handle the HID writes without the userspace implementation.

src/drivers/oxp_hid/driver.rs
honjow: Changed to PackedStruct method

src/drivers/oxp_hid/driver.rs
honjow: Modified

honjow:
Already processed and submitted according to the modification suggestions

srsholmes:
Are all requested changes done on this now? Can we merge it ? It would be great to give input plumber a go on bazzite with the other kernal fixes. 

I dont mind building from source, but it would be good to know that this is not going to change in the meantime.

pastaq:
> Are all requested changes done on this now? Can we merge it ? It would be great to give input plumber a go on bazzite with the other kernal fixes.
> 
> I dont mind building from source, but it would be good to know that this is not going to change in the meantime.

I have some feedback form the latest version of the WIP kernel driver to incorporate. Once that is done I'll submit the v2 patch and PR it to the OGC kernel. This implementation relies on that driver so I want to wait for it to be merged before this is merged.

src/drivers/oxp_hid/driver.rs
```diff
@@ -0,0 +1,133 @@
...
+        .join(" ")
+}
+
+impl Driver {
+    pub fn new(path: String) -> Result<Self, Box<dyn Error + Send + Sync>> {
+        let cs_path = CString::new(path.clone())?;
+        let api = hidapi::HidApi::new()?;
+        let device = api.open_path(&cs_path)?;
+        let info = device.get_device_info()?;
+        log::info!(
```
pastaq: Please make this debug level

src/drivers/oxp_hid/hid_report.rs
```diff
@@ -0,0 +1,62 @@
...
+    L3 = 0x0b,
+    R3 = 0x0c,
+    DUp = 0x0d,
+    DDown = 0x0e,
+    DLeft = 0x0f,
+    DRight = 0x10,
+    Guide = 0x21,
+    M1 = 0x22,
+    M2 = 0x23,
+    Keyboard = 0x24,
```
pastaq: This is essentially identical to this: https://github.com/ShadowBlip/InputPlumber/blob/2cefc80f9b39a6ee91f69a6ba4f86f4611861612/src/drivers/oxp_tty/serial_report.rs#L27-L51

The only differences being I used modern terms for start/select and triggers. We should de-duplicate this. I think the simplest way would be to move it to drivers/oxp_tty/mod.rs and make it public,  t hen you can import it here.

pastaq (changes requested):
I noticed a couple more minor things after another review. I think the driver is in a good state & once these are done it should be gtg from my perspective.

src/drivers/oxp_hid/driver.rs
honjow: Modified

src/drivers/oxp_hid/hid_report.rs
honjow: Okay, modified and committed

pastaq (approved):

ShadowApex (approved):

PR #534, opened 2026-02-23
feat(CapabilityMap): add config for inverting axis values

ShadowApex:
This change adds a new `inverse` option for axes, which can allow a user to invert the axis value of a matching direction.

E.g.

```yaml
mapping:
  - name: Invert left stick vertical
    source_events:
      - gamepad:
          axis:
            name: LeftStick
    target_event:
      gamepad:
        axis:
          name: LeftStick
          direction: vertical
          invert: true
```

Fixes #532

Issue #532, opened 2026-02-21
(feat): capabilty map axis flip


loki666:
some devices have their joystick mounted the wrong way, and so report flipped axis
add a capability map flag to let InputPlumber correct that

something like

```
  - name: Right Trigger
    source_events:
      - evdev:
          event_type: ABS
          event_code: ABS_RZ
          value_type: trigger
          inverse: true
    target_event:
      gamepad:
        trigger:
          name: RightTrigger
```
and
```
  - name: Right Stick
    source_events:
      - evdev:
          event_type: ABS
          event_code: ABS_RX
          value_type: joystick_x
          inverse: true
      - evdev:
          event_type: ABS
          event_code: ABS_RY
          value_type: joystick_y
          inverse: true
    target_event:
      gamepad:
        axis:
          name: RightStick
```

PR #569, opened 2026-03-28
fix(CapabilityMap): use min/max range for unsigned axis normalization

honjow:
`normalize_unsigned_value` was computing `value / max`, which only works when min is 0. For axes where min is non-zero — such as ADC-based triggers found on some ARM handhelds (e.g. Gameforce Ace, which uses the `adc-joystick` driver and reports triggers as `ABS_HAT2X`/`ABS_HAT2Y` with a range of 890 to 1530) — the values are never correctly scaled to 0.0 to 1.0.

Changed to `(value - min) / (max - min)` with a `.clamp(0.0, 1.0)`, and updated `denormalize_unsigned_value` accordingly.

Also adds an `inverted` field to `EvdevConfig` (source side). This complements the existing `invert` option on `AxisCapability` (target side): the existing one handles `Vector2` joystick axes after translation, but there's no way to invert a trigger (`Float` value) at the source. ADC triggers on some hardware report values in reverse — high at rest, low when pressed — and without this option there's currently no way to handle it in a capability map.

```yaml
- evdev:
    event_type: ABS
    event_code: ABS_HAT2X
    value_type: trigger
    inverted: true
```

No behavior change for devices where min is already 0.

### DIFF
--- a/rootfs/usr/share/inputplumber/schema/capability_map_v2.json
+++ b/rootfs/usr/share/inputplumber/schema/capability_map_v2.json
@@ -268,11 +268,6 @@
           "maximum": 65535,
           "minimum": 0
         },
-        "invert": {
-          "description": "If true, invert the axis value. For triggers (0.0 to 1.0), this applies 1.0 - value. For joystick axes (-1.0 to 1.0), this negates the value.",
-          "type": "boolean",
-          "default": false
-        },
         "value_type": {
           "$ref": "#/$defs/ValueType"
         }
--- a/rootfs/usr/share/inputplumber/schema/capability_map_v2.json
+++ b/rootfs/usr/share/inputplumber/schema/capability_map_v2.json
@@ -268,8 +268,8 @@
           "maximum": 65535,
           "minimum": 0
         },
-        "inverted": {
-          "description": "If true, invert the axis value. For triggers (0.0~1.0), this applies 1.0 - value. For joystick axes (-1.0~1.0), this negates the value.",
+        "invert": {
+          "description": "If true, invert the axis value. For triggers (0.0 to 1.0), this applies 1.0 - value. For joystick axes (-1.0 to 1.0), this negates the value.",
           "type": "boolean",
           "default": false
         },
--- a/rootfs/usr/share/inputplumber/schema/capability_map_v2.json
+++ b/rootfs/usr/share/inputplumber/schema/capability_map_v2.json
@@ -268,6 +268,11 @@
           "maximum": 65535,
           "minimum": 0
         },
+        "inverted": {
+          "description": "If true, invert the axis value. For triggers (0.0~1.0), this applies 1.0 - value. For joystick axes (-1.0~1.0), this negates the value.",
+          "type": "boolean",
+          "default": false
+        },
         "value_type": {
           "$ref": "#/$defs/ValueType"
         }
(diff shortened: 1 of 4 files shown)


src/config/capability_map/evdev.rs
```diff
@@ -10,6 +10,10 @@ pub struct EvdevConfig {
     pub event_value: Option<u16>,
     pub value_type: ValueType,
     pub axis_direction: Option<AxisDirection>,
+    /// If true, invert the axis value. For triggers (0.0~1.0), this applies
+    /// `1.0 - value`. For joystick axes (-1.0~1.0), this negates the value.
+    #[serde(default)]
+    pub inverted: bool,
```
ShadowApex: Invert is already supported: https://github.com/ShadowBlip/InputPlumber/pull/534

rootfs/usr/share/inputplumber/schema/capability_map_v2.json
```diff
@@ -268,6 +268,11 @@
           "maximum": 65535,
           "minimum": 0
         },
+        "inverted": {
```
pastaq: Let's call this invert as well to keep terms consistent

pastaq (changes requested):
Just a small bit, everything else looks good.

rootfs/usr/share/inputplumber/schema/capability_map_v2.json
honjow: > Let's call this invert as well to keep terms consistent

Changes pushed

rootfs/usr/share/inputplumber/schema/capability_map_v2.json
```diff
@@ -268,6 +268,11 @@
           "maximum": 65535,
           "minimum": 0
         },
+        "invert": {
```
pastaq: I don't think the inversion stuff is particularly necessary. It appears to be redundant with #534. It is also out of scope for fixing the bug with axes not starting at 0. If you think there is a valid case for it please submit that as a separate PR.

pastaq (changes requested):

rootfs/usr/share/inputplumber/schema/capability_map_v2.json
honjow: Okay, I'll remove it

sydarn: Would inverting triggers in a similar manner as in #534 be acceptable?

sydarn: After thinking about this, the issue in #532 seems identical to the one I had with triggers, just with stick-axes. #534 only seems to solve the issue for stick-axes, whereas I believe honjow suggestions would solve both cases iiuc. I would need need to test this with loki666 to be sure though.

pastaq: In either case it's a separate issue so a separate PR is necessary

ShadowApex (approved):

src/config/capability_map/evdev.rs
sydarn: I can't get that to work with linux' adc-joystick driver, whereas I was able to get the implementation from honjow to work.

pastaq (approved):

PR #571, opened 2026-03-29
feat(Touchscreen): edge-swipe gesture detection for handheld

honjow:
## Summary

Adds single-finger edge-swipe gesture detection to the touchscreen source device, allowing touchscreen edges to be mapped to gamepad buttons via device profiles — enabling gestures like swiping from the left edge to open the Steam menu, or from the right edge to open the Quick Access Menu, without requiring changes to the compositor.

Tested and verified on **GPD Win5** and **ONEXPLAYER X1 Mini**.

---

## How gesture detection works

Touch coordinates are normalized to `[0.0, 1.0]`. A gesture is recognized when the touch starts within 3% of a screen edge, travels inward by at least 12%, and completes within 400 ms.

```
+-----------------------------------+
|###################################|
|#                                 #|
|#   +--------------------------+  #|
|#   |                          |  #|
|#   |     normal touch area    |  #|
|#   |                          |  #|
|#   +--------------------------+  #|
|#                                 #|
|###################################|
+-----------------------------------+
  ^                               ^
  # = edge zone (3% of each side)
```

Gesture names describe the **direction of finger movement** (e.g. swiping inward from the left edge moves rightward → `SwipeRight`). Each direction carries a `Top`/`Bottom` zone qualifier split at 33% along the perpendicular axis.

**Default profile mappings:**

| Gesture             | Action        |
|---------------------|---------------|
| `SwipeRight:Bottom` | `Guide`       |
| `SwipeLeft:Bottom`  | `QuickAccess` |

---

## Two operating modes: passthrough vs. grab

**Passthrough (`grab: false`, default):** Raw touch events are forwarded normally alongside any gesture events. Suitable for desktop use; avoid in game compositors as gestures will also register as taps.

**Grab (`grab: true`):** The touchscreen is exclusively captured via `EVIOCGRAB`. Touches starting within 1% of an edge are held back until the gesture outcome is known — confirmed gestures suppress the touch, unconfirmed ones replay it as a normal tap. Multi-finger touch immediately cancels gesture detection and passes all events through. If `EVIOCGRAB` fails, the source falls back to passthrough with a warning.

Grab mode is recommended for devices running a game compositor such as Gamescope.

---

## Why MIN_FRAME_TIME was moved to TargetDriver

Gesture detection emits press and release nearly simultaneously. Without a minimum press duration, both events fall within a single HID poll interval and the compositor sees a report where the button was never pressed.

`steam_deck` already handled this with an 80 ms minimum, but only for itself. This PR moves the enforcement into `TargetDriver` so every target benefits automatically, and removes the now-redundant `steam_deck` implementation.


rootfs/usr/share/inputplumber/schema/composite_device_v1.json
```diff
@@ -243,6 +243,10 @@
       "type": "object",
       "additionalProperties": false,
       "properties": {
+        "grab": {
+          "description": "If true, the touchscreen will be exclusively grabbed so raw touch events are not passed to other consumers. Defaults to false.",
+          "type": "boolean"
+        },
```
ShadowApex, line 249: This is not necessary as this is already managed using the `passthrough` config option.

src/config/mod.rs
```diff
@@ -303,6 +303,12 @@ pub struct TouchscreenConfig {
     /// instead of the size advertised by the device itself.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub override_source_size: Option<bool>,
+    /// If true, the device will be grabbed exclusively. Touch events will be
+    /// fully managed by InputPlumber; a virtual touchscreen target device will
+    /// be created and edge-swipe gestures will suppress raw touch delivery.
+    /// Defaults to false (pass-through mode).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub grab: Option<bool>,
```
ShadowApex, line 311: Also here, this is already implemented by `passthrough`.

src/input/composite_device/targets.rs
```diff
@@ -51,6 +54,9 @@ pub struct CompositeDeviceTargets {
     /// Map of DBusDevice DBus paths to their respective transmitter channel.
     /// E.g. {"/org/shadowblip/InputPlumber/devices/target/dbus0": <Sender>}
     target_dbus_devices: HashMap<String, TargetDeviceClient>,
+    /// True if this composite device has a grabbed touchscreen source.
+    /// When true, the touchscreen target will always be kept in the target list.
+    has_touchscreen_source: bool,
```
ShadowApex, line 59: We should not always keep a touchscreen target in the target list. This should be controllable by the user. There may be cases where the user does _not_ want to emulate a touchscreen (e.g. by converting the touchscreen inputs into touchpad inputs).

src/input/composite_device/targets.rs
```diff
@@ -124,9 +141,20 @@ impl CompositeDeviceTargets {
...
+
+        // If the composite device has a grabbed touchscreen source, always keep
+        // touchscreen in the target list regardless of what the caller requested.
+        if self.has_touchscreen_source {
+            let touchscreen_id: TargetDeviceTypeId = TOUCHSCREEN_TARGET_ID.try_into().unwrap();
+            if !device_types.contains(&touchscreen_id) {
+                log::debug!("[{dbus_path}] Auto-adding touchscreen target (grabbed touchscreen source detected)");
+                device_types.push(touchscreen_id);
+            }
+        }
```
ShadowApex, line 156: Also here, we should not assume to always emulate a virtual touchscreen. Users may not want to use their touchscreen as a touchscreen.

src/input/source/evdev/touchscreen.rs
```diff
@@ -40,11 +53,82 @@ impl From<&str> for Orientation {
...
+    Idle,
+    /// Touch started; watching for a gesture.
+    /// In grab mode, `suppressing` indicates whether slot-0 events are being
+    /// held back because the touch started in an edge zone.
+    Tracking { suppressing: bool },
+    /// A gesture was recognized; suppress remaining touch events until lift.
+    Triggered,
+    /// A second finger arrived; gesture detection disabled for this touch.
+    Invalidated,
+}
```
ShadowApex, line 78: I'm not sure if we actually want gesture detection to take place in the source device implementation. We may want to instead handle this in the composite device.

@pastaq thoughts?

ShadowApex (changes requested):
This feature is definitely something we would like to include, but we may need to consider exactly how we want to implement this.

src/input/source/evdev/touchscreen.rs
pastaq: I agree, but I'm having difficulty imagining how that would work exactly. They don't cleanly map to native events, and we don't want to lose the rest of the touchscreen.

Perhaps we should allow for a config file or dbus command to assign a capability_map to specific regions by start/end for x and y by % of the screen height/width, then the driver can translate those regions to a specific native event. This would be more flexible than just using gestures as it would allow for a fully virtual gamepad or touchpad on the screen itself. An overlay program could then assign those regions to match the location for its glyphs over the top of a game.

PR #574, opened 2026-04-03
fix(iio): improve AccelGyro3d driver for HID Sensor Hub devices

honjow:
## Summary

Fix the `AccelGyro3d` IIO driver for HID Sensor Hub devices (Intel ISH, AMD SFH) where accelerometer and gyroscope are exposed as separate IIO devices (`accel_3d`, `gyro_3d`).

### Handle separate accel/gyro IIO devices

The previous code assumed both accel and gyro channels were always present in one device, emitting zero-valued events for the missing sensor. This caused alternating valid/zero data on HID Sensor Hub hardware (e.g. MSI Claw).

- Detect capabilities dynamically via `has_accel()` / `has_gyro()`
- Return `None` from poll when no channels are found
- Keep the IIO `Device` alive in `Driver` to prevent `Channel` pointer invalidation

### Correct scaling factors

The previous scaling factors (`* 10.0` for accel, `* (180/π) * 1500.0` for gyro) were empirically tuned on the Legion Go — the original commit explicitly states they were derived "from testing" to make motion "feel like natural 1:1". These values were likely calibrated against the `amd_sfh` kernel driver's imprecise output (the driver has known integer division errors that reduce accel values by 10× and gyro by 100×), rather than against correct SI-unit data.

This PR replaces them with physically derived values based on the Steam Deck UHID protocol constants:

- Accel: `1 / 0.0006125` (m/s² → UHID LSB)
- Gyro: `(180/π) / 0.0625` (rad/s → °/s → UHID LSB)

These factors assume the IIO subsystem reports correct SI-unit values. Hardware-specific sensitivity is handled by the per-device `scale` sysfs attribute. Verified on MSI Claw A1M (Intel ISH) and MSI Claw A8 BZ2EM (AMD SFH with kernel precision patch) — both produce correct and consistent output.

**Impact on Legion Go:** When the `hid-lenovo-go` kernel driver is loaded, InputPlumber already disables the internal IMU and uses the controller's built-in gyroscope instead (`get_default_event_filter`), so this change has no effect. On kernels without `hid-lenovo-go` where the internal IMU is active, the new factors may produce different output if the `amd_sfh` precision bug is present — but the old values were not physically correct either.

### Initialize sampling rate on startup

Some HID Sensor Hub devices default to very low rates (e.g. 10 Hz on MSI Claw), which is insufficient for gaming input. The driver now initializes the rate with priority: YAML config > hardware max available > 200 Hz default. Supports both per-channel and device-level attributes with fallback.

## Test plan

- [x] MSI Claw A1M (Intel ISH, separate `accel_3d` + `gyro_3d`)
- [x] MSI Claw A8 BZ2EM (AMD SFH). Note: stock `amd_sfh` kernel driver has precision issues requiring a kernel patch, independent of InputPlumber.
- [x] BMI260 device — no regression in shared Driver code
- [ ] Verify no regression on Legion Go / ROG Ally

### DIFF
--- a/src/drivers/iio_imu/driver.rs
+++ b/src/drivers/iio_imu/driver.rs
@@ -91,11 +91,17 @@ impl Driver {
             }
         }
 
-        if !accel.is_empty() {
-            set_sample_rate(&device, &accel, ChannelType::Accel, sample_rate);
-        }
-        if !gyro.is_empty() {
-            set_sample_rate(&device, &gyro, ChannelType::AnglVel, sample_rate);
+        // Request a higher sampling rate
+        for (channels, ch_type) in [(&accel, ChannelType::Accel), (&gyro, ChannelType::AnglVel)] {
+            if channels.is_empty() {
+                continue;
+            }
+            if let Err(err) =
+                set_sample_rate_or_default(&device, channels, ch_type, sample_rate)
+            {
+                log::warn!("Failed to set sample rate: {err}, falling back to max available");
+                set_sample_rate_max(&device, channels, ch_type);
+            }
         }
 
         Ok(Self {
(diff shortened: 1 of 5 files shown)


src/drivers/iio_imu/driver.rs
```diff
@@ -103,6 +103,14 @@ impl Driver {
     //a standalone crate. When this driver is eventually separated, refactor the Event type to
     //follow the pattern DeviceEvent(Event, Value) and create a match table for
     //Capability->Event/Event->Capability in the SourceDriver implementation.
+    pub fn has_accel(&self) -> bool {
```
pastaq: These functions seem fairly redundant to me. Please remove them and just do the same check they are already doing

pastaq (changes requested):

src/input/source/iio/accel_gyro_3d.rs
```diff
@@ -111,25 +111,22 @@ fn translate_events(events: Vec<iio_imu::event::Event>) -> Vec<NativeEvent> {
 fn translate_event(event: iio_imu::event::Event) -> NativeEvent {
     match event {
         iio_imu::event::Event::Accelerometer(data) => {
+            let factor = 1.0 / 0.0006125; // m/s² → UHID LSB
```
pastaq: This never changes. Instead of recalculating it every time please declare this as a const f64 1632.65

src/input/source/iio/accel_gyro_3d.rs
```diff
@@ -111,25 +111,22 @@ fn translate_events(events: Vec<iio_imu::event::Event>) -> Vec<NativeEvent> {
...
             };
             NativeEvent::new(cap, value)
         }
         iio_imu::event::Event::Gyro(data) => {
-            // Translate gyro values into the expected units of degrees per sec
-            // We apply a 500x scale so the motion feels like natural 1:1 motion.
-            // Adjusting the scale is not possible on the accel_gyro_3d IMU.
-            // From testing this is the highest scale we can apply before noise
-            // is amplified to the point the gyro cannot calibrate.
+            let factor = (180.0 / PI) / 0.0625; // rad/s → UHID LSB
```
pastaq: Same for this, const f64 916.73

pastaq (changes requested):
Any links to substantiate the SI units claim would be good in the commit description.

src/drivers/iio_imu/driver.rs
```diff
@@ -88,6 +91,13 @@ impl Driver {
             }
         }
 
+        if !accel.is_empty() {
+            set_sample_rate(&device, &accel, ChannelType::Accel, sample_rate);
```
pastaq: I don't really like this function as is.

- The function doesn't check if the sample rate provided is valid.
- If no sample rate is provided, it either uses the max (unintuitive to the function name), or uses a default which isn't checked as available. This will provide inconsistent behavior that is undefined.

The function name doesn't provide any of this context, so depending on how it's used you have two paths to force undefined behavior and one that provides a random result.

I would have two functions:
 - One called set_sample_rate_or_default() that takes the rate as an option. This function should throw a warning under two circumstances. 1.) The requested rate isn't available. 2.)  The default isn't available. If the first failcondition is hit, or if no rate is provided, it should fall back to the default.
 -  One called set_sample_rate_max that you fall back to if the first one warns. This should produce an error if it fails, but the driver should keep working.

src/drivers/iio_imu/driver.rs
```diff
@@ -88,6 +91,13 @@ impl Driver {
...
+        }
+        if !gyro.is_empty() {
+            set_sample_rate(&device, &gyro, ChannelType::AnglVel, sample_rate);
```
pastaq: Same here

src/drivers/iio_imu/driver.rs
```diff
@@ -358,3 +368,96 @@ fn is_driver_loaded(driver_name: &str) -> io::Result<bool> {
...
+
+fn set_sample_rate(
+    device: &Device,
+    channels: &HashMap<String, Channel>,
+    channel_type: ChannelType,
+    target_rate: Option<f64>,
+) {
+    let rate = if let Some(r) = target_rate {
+        log::debug!("Using configured sample rate: {r} Hz");
+        r
```
pastaq: Please don't represent variables with single letters outside of using Err(e) which is universally understandable in context. Explicit variable names are preferred and shadowing won't happen in rust with the conplilers scope enforcement.

src/drivers/iio_imu/driver.rs
```diff
@@ -358,3 +368,96 @@ fn is_driver_loaded(driver_name: &str) -> io::Result<bool> {
...
+        match channel.attr_write_float("sampling_frequency", rate) {
+            Ok(_) => {
+                let actual = channel
+                    .attr_read_float("sampling_frequency")
+                    .unwrap_or(rate);
+                log::debug!("Set sampling_frequency to {actual} Hz via channel {id}");
+                return;
+            }
+            Err(e) => {
+                log::debug!(
```
pastaq: I don't like hiding errors that obscure information like this behind a debug command as this has an appreciable effect on the behavior of the device. After you refactor the previous function, ensure that both failing to set the config/default rate and if the fallback to max fails that they at least produce a warning.

src/drivers/iio_imu/driver.rs
```diff
@@ -358,3 +368,96 @@ fn is_driver_loaded(driver_name: &str) -> io::Result<bool> {
...
+    // device-level fallback
+    let attr = match channel_type {
+        ChannelType::Accel => "in_accel_sampling_frequency",
+        ChannelType::AnglVel => "in_anglvel_sampling_frequency",
+        _ => return,
+    };
+
+    match device.attr_write_float(attr, rate) {
+        Ok(_) => {
+            let actual = device.attr_read_float(attr).unwrap_or(rate);
```
pastaq: This also obscures an error reading from the sysfs. Please handle the error oth a warning and then report that we're assuming a successful write to the rate.

pastaq (changes requested):

src/drivers/iio_imu/driver.rs
pastaq: I edited this to be more clear. Please adjust. 

Also, your changes were very quick which seems tool assisted. Please disclose any use of AI tooling.

honjow:
Heads up per the AI policy — the last two commits (027034f, 97c5f5d) used Claude to help with the refactoring. Marked with Co-developed-by in the commit messages.

What I did: after getting the review feedback, I described the requested changes to Claude and had it generate the refactored code. Specifically:

- For 027034f: I asked it to remove the `has_accel()`/`has_gyro()` methods and revert to the static CAPABILITIES array, and to move the scaling factor expressions to const. The prompt was basically "the reviewer says these wrapper methods are redundant, remove them" and "move these to const f64".

- For 97c5f5d: I described the reviewer's concern about `set_sample_rate` — that it doesn't validate the rate and hides errors behind debug logging — and asked it to split the function into a validated path and a max-fallback path with proper warn-level logging.

I reviewed the generated output, checked that it compiles, and verified the logic matches what was asked for in the review. Haven't re-deployed to hardware yet since these are refactors of the same behavior, but can do if needed.

matte-schwartz:
this PR significantly improves the gyro experience on all the ROG Ally models I've tried, as the default polling rate of 25 Hz puts bmi323 into a lower-power mode where reporting tends to be more inconsistent and has a directional drift.

ShadowApex (approved):

pastaq (approved):

pastaq:
bah, I meant to squash this one first. My bad.

PR #576, opened 2026-04-13
fix(udev): run udev rule on bind event

matte-schwartz:
When hid-generic binds first and the specific HID driver takes over later, only a "bind" event is generated, not "add" or "change". This means our udev rule will not be triggered during bootup as it should be, and os_mode is not set to linux on Go S.


pastaq (approved):
To recap what we discussed in DM, I'd like to explore this bind action for the Asus driver as well since the symptoms seems to match.

PR #590, opened 2026-05-21
feat(Hardware Support): Qualcomm SSC sensors through FastRPC

giodotblue:
This PR adds support for the FastRPC "subsystem" and the Qualcomm Sensor Core

When combined with the correct sensor data and [hexagonrpc](https://github.com/linux-msm/hexagonrpc), this can provide accelerometer & gyroscope support to Snapdragon handhelds that use the SSC for that

**Big changes, in no order:**

- libloading has been added as a dependency
- libssc, glib and gobject are runtime dependencies (they don't need to be on the build machine or any targets that don't need it)
- FastRPC has been added as an input source / "subsystem"
- ssc has been added as a driver

**I'm opening this as a draft as I have a couple concerns:**

- FastRPC devices show up in the "misc" subsystem - UdevDevice now checks for devices with the misc subsystem and fastrpc prefix, then fakes the subsystem
- I'd like to see what other people think of the [dylib loading at runtime](https://github.com/gio3k/InputPlumber/blob/1b1459db0ad7c2543a23aff763ec0de0eeeb7f8b/src/drivers/ssc/bindings.rs), it's a bit weird code-wise but I'm not sure there's a better way here

**Also, minor naming concern:** 

Maybe the ssc driver should be split into two, something like:

`drivers/ssc/runtime.rs, drivers/ssc/bindings.rs`
`drivers/ssc_imu/driver.rs, drivers/ssc_imu/event.rs`

This is because SSC can provide other sensor data in the future aside from gyro / accel. 

Where would I put the `drivers/ssc` folder in that case - it seems out of place to have it there if it's just the core SSC stuff?

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -440,6 +440,16 @@ pub struct MountMatrix {
     pub z: [f64; 3],
 }
 
+impl Default for MountMatrix {
+    fn default() -> Self {
+        MountMatrix {
+            x: [1.0, 0.0, 0.0],
+            y: [0.0, 1.0, 0.0],
+            z: [0.0, 0.0, 1.0],
+        }
+    }
+}
+
 #[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
 #[serde(rename_all = "snake_case")]
 pub struct EventsConfig {
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -231,6 +231,9 @@ pub struct SourceDevice {
     /// Devices that match the given evdev properties will be captured by InputPlumber
     #[serde(skip_serializing_if = "Option::is_none")]
     pub evdev: Option<Evdev>,
+    /// Devices that match the given fastrpc properties will be captured by InputPlumber
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub fastrpc: Option<FastRpc>,
     /// Devices that match the given hidraw properties will be captured by InputPlumber
     #[serde(skip_serializing_if = "Option::is_none")]
     pub hidraw: Option<Hidraw>,
@@ -401,6 +404,15 @@ pub struct IIO {
     )]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub mount_matrix: Option<MountMatrix>,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
+#[serde(rename_all = "snake_case")]
+pub struct FastRpc {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub id: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub name: Option<String>,
 }
 
 #[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
@@ -586,6 +598,12 @@ impl CompositeDeviceConfig {
             "tty" => {
                 let tty_config = config.tty.as_ref()?;
                 if self.has_matching_tty(udevice, tty_config) {
+                    return Some(config.clone());
+                }
+            }
+            "fastrpc" => {
+                let fastrpc_config = config.fastrpc.as_ref()?;
+                if self.has_matching_fastrpc(udevice, fastrpc_config) {
                     return Some(config.clone());
                 }
             }
@@ -843,6 +861,35 @@ impl CompositeDeviceConfig {
         if let Some(name) = tty_config.name.as_ref() {
             let dname = device.name();
             log::trace!("Checking name: {name} against {dname}");
+            if !glob_match(name.as_str(), dname.as_str()) {
+                return false;
+            }
+        }
+
+        true
+    }
+
+    /// Returns true if a given FastRPC device is within a list of fastrpc configs.
+    pub fn has_matching_fastrpc(&self, device: &UdevDevice, fastrpc_config: &FastRpc) -> bool {
+        log::trace!(
+            "Checking FastRPC config: {:?} against {:?}",
+            fastrpc_config,
+            device
+        );
+
+        if let Some(id) = fastrpc_config.id.as_ref() {
+            let dsyspath = device.syspath();
+            log::trace!("Checking id: {id} against {dsyspath}");
+            if !glob_match(id.as_str(), dsyspath.as_str()) {
+                return false;
+            }
+        }
+
+        if let Some(name) = fastrpc_config.name.as_ref() {
+            // FastRPC devices don't seem to have a name (or it's empty)
+            // The sysname works fine
+            let dname = device.sysname();
+            log::trace!("Checking sysname: {name} against {dname}");
             if !glob_match(name.as_str(), dname.as_str()) {
                 return false;
             }
--- a/src/drivers/ssc/bindings.rs
+++ b/src/drivers/ssc/bindings.rs
@@ -37,25 +37,13 @@ pub mod glib {
     pub type FnGMainContextIteration =
         unsafe extern "C" fn(context: *mut GMainContext, may_block: gboolean) -> gboolean;
 
-    /* Helpers */
-    /// Converts a GLib error ptr to a basic string (or None if the error is null)
-    pub fn glib_error(error: *mut GError) -> Option<String> {
-        if error.is_null() {
-            None
-        } else {
-            unsafe {
-                Some(format!(
-                    "GLib error: domain = {}, code = {}",
-                    (*error).domain,
-                    (*error).code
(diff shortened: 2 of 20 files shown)


src/config/mod.rs
```diff
@@ -231,6 +231,9 @@ pub struct SourceDevice {
     /// Devices that match the given evdev properties will be captured by InputPlumber
     #[serde(skip_serializing_if = "Option::is_none")]
     pub evdev: Option<Evdev>,
+    /// Devices that match the given fastrpc properties will be captured by InputPlumber
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub fastrpc: Option<FastRpc>,
```
pastaq, line 236: Does this new config interface add additional value? I would think that the udev config would be able to find these devices without the need for another protocol specific entry. evdev/hidraw/etc. are all legacy config options we'd ideally like to get away from.

src/drivers/ssc/bindings.rs
```diff
@@ -0,0 +1,277 @@
+pub mod glib {
```
pastaq: This file contains a lot of C binding abstractions, but there is very little context to the purpose of each function or variable. Please go through it and add amplifying information such that we don't need to find & read the ssc documentation to have an understanding of the purpose of each. It doesn't need to be an exhaustive reproduction of the documentation, just enough to explain as an overview.

src/drivers/ssc/bindings.rs
```diff
@@ -0,0 +1,277 @@
...
+        pub ref_count: u32,
+        _truncated_record_marker: std::ffi::c_void,
+    }
+    pub type GClosureNotify = Option<unsafe extern "C" fn(gpointer, *mut GClosure)>;
+    pub type GCallback = Option<unsafe extern "C" fn()>;
+
+    /* Function ptrs */
+    /// g_main_context_iteration
+    pub type FnGMainContextIteration =
+        unsafe extern "C" fn(context: *mut GMainContext, may_block: gboolean) -> gboolean;
```
pastaq: Additionally, please add `// SAFETY:` comment blocks to all unsafe calls explaining how the function can be safely used, per the rust guidelines.

src/drivers/ssc/driver.rs
```diff
@@ -0,0 +1,139 @@
...
+        Ok(HashSet::new())
+    }
+
+    /// Poll the device for data
+    pub fn poll(&mut self) -> Result<Vec<Event>, Box<dyn Error + Send + Sync>> {
+        // libssc uses GLib and relies on the GLib main loop
+        // Instead of creating a main loop and main context then managing it alongside InputPlumber's loop etc,
+        // we can just perform an iteration of the GLib context every poll
+        // Not sure how this will work if other parts of InputPlumber start using the GLib main loop too for some reason
+        unsafe { (self.runtime.libglib.g_main_context_iteration)(std::ptr::null_mut(), GFALSE) };
```
pastaq: `// SAFETY:` comment needed.

src/drivers/ssc/driver.rs
```diff
@@ -0,0 +1,139 @@
...
+
+                Err(TryRecvError::Empty) => break,
+                Err(error) => return Err(format!("Error when handling SSC events: {error}").into()),
+            }
+        }
+
+        Ok(events)
+    }
+
+    /// Rotate the given axis data according to the mount matrix. This is used
```
pastaq, line 114: Any time we start duplicating code, it is a good idea to consider if a higher level utility file is beneficial. This could indicate we need a public IMU Utils file that each driver can import.

src/input/source/fastrpc/ssc_imu.rs
```diff
@@ -0,0 +1,108 @@
...
+
+impl SourceOutputDevice for SscImu {}
+
+impl Debug for SscImu {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("SscImu").finish()
+    }
+}
+
+// NOTE: Mark this struct as thread-safe as it will only ever be called from
```
pastaq: Use // SAFETY: instead of // NOTE:

src/udev/device.rs
```diff
@@ -436,15 +432,21 @@ impl UdevDevice {
             Err(_) => return Default::default(),
         };
 
+        let subsystem = match device.subsystem().map(|s| s.to_string_lossy().to_string()) {
+            // FastRPC has special behaviour, check the comment in "From<::udev::Device> for UdevDevice"
```
pastaq, line 436: Please don't refer to other comments/files. Each section should be understandable within its own context.

src/udev/device.rs
```diff
@@ -695,11 +700,17 @@ impl From<::udev::Device> for UdevDevice {
             .unwrap_or(Path::new(""))
             .to_string_lossy()
             .to_string();
-        let subsystem = device
-            .subsystem()
-            .unwrap_or(OsStr::new(""))
-            .to_string_lossy()
-            .to_string();
+
```
pastaq, line 703: I think it would be better to fully support the "misc" subsystem, then determine a device is a fastrpc when searching for a driver. That will be more scalable in the future if we add additional "misc" devices as we won't need to add additional cases here or fake the subsystem.

src/input/source/fastrpc.rs
pastaq, line 1: In the same vein as the comment below, this file should be for "misc" subsystem, then it should identify a device as a fastrpc IMU here and load the driver. This will allow us to add more "misc" devices in the future if we need to without refactoring udev/manager again.

pastaq (changes requested):
Looking good so far. Understand this is a WIP so it isn't in a finished state. I'm asking for a bit of cleanup with the code comments, and one significant change to device discovery. I'm also not convinced we need a fastrpc specific config option.

src/input/source/fastrpc/ssc_imu.rs
giodotblue: Sure, iirc this line was copied from bmi_imu so it might need to be changed there too

src/udev/device.rs
giodotblue: Yeah sorry this is bad I was planning to put this comment in the PR description but I forgot

src/drivers/ssc/bindings.rs
giodotblue: I reworked it and added comments around the externals. Let me know what you think

src/udev/device.rs
giodotblue: Will fix this when I create the misc subsystem

src/drivers/ssc/driver.rs
giodotblue: I agree, but iio_imu uses a different mount_matrix type than ssc does (the iio_imu one is deprecated) 
https://github.com/ShadowBlip/InputPlumber/blob/85f5fd5c52e414cc01f59f9754c0eb71a5a91b7e/src/config/mod.rs#L398-L403

iio_imu isn't currently reading the new config.imu.mount_matrix key either, so I was thinking of fixing that in a subsequent PR

If you'd like I could do that in this PR, I'm just worried about it getting too big

pastaq: That's probably best to wait and do in a separate PR then. I don't want to introduce regressions for older configs

giodotblue:
sorry, I didn't actually re-request the review, my bad!

when you get time, let me know if the bindings.rs changes are OK. I'll start working on changing the subsystem to misc

pastaq:
> sorry, I didn't actually re-request the review, my bad!
> 
> when you get time, let me know if the bindings.rs changes are OK. I'll start working on changing the subsystem to misc

I missed this comment. They are better. Are you still working on this? We're getting a lot more ARM devices supported so this would be a huge benefit.

giodotblue:
> I missed this comment. They are better. Are you still working on this? We're getting a lot more ARM devices supported so this would be a huge benefit.

Sure, will look into finishing this PR sometime this week

edit: sorry work got in the way, will do this as soon as I can

jenneron:
BTW We should switch odin 3 target to `hori-steam` or `8bitdo-u2` because current `xbox-elite` doesn't support gyroscope. The same applies to other qualcomm devices with gyroscope. Although this can be done in a separate PR

pastaq:
> BTW We should switch odin 3 target to `hori-steam` or `8bitdo-u2` because current `xbox-elite` doesn't support gyroscope. The same applies to other qualcomm devices with gyroscope. Although this can be done in a separate PR

Please keep this on topic. If you feel that is a problem please open an issue and we can discuss it.

PR #593, opened 2026-06-01
Implement Axis to Mouse Wheel mapping

TheBigViolinist:
Note: Sometimes when the axis dont move the wheel when moved directly to the maximum of an axis, i realy dont know why.... Il be glad if its possible to fix this... And i added on capatibility map the "wheelspeed" value for future updates (i dont know if it works)
hope that feature goes into the next release :)
edit: part of the code was LLM assisted, i dont know how to give input on the max border of axis y an x, didnt work either...

### DIFF
--- a/src/config/capability_map.rs
+++ b/src/config/capability_map.rs
@@ -302,6 +302,7 @@ pub struct MouseMotionCapability {
 #[serde(rename_all = "snake_case")]
 pub struct MouseWheelCapability {
     pub direction: Option<String>,
+    pub wheelspeed: Option<f64>, 
 }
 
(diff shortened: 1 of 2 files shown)


pastaq:
After discussing with @ShadowApex again yesterday, we're going to be updating our policy on AI usage to not allow it for the time being, and we'll also be removing any existing AI assisted contributions and redeveloping them ourselves. This policy will be revisited after the legal questions about FOSS licensing and LLM generated code are answered. We're primarily concerned about copyleft contamination braking the GPL license.

TheBigViolinist:
Will it release in the next update? :((

PR #601, opened 2026-06-07
fix(DualSense): only emit changed input reports for IMU

sunshineinabox:
DS5 emits at its standard poll rate of 250Hz, however for devices that don't have any IMU it continues to report the same data (zero). This should be a no-op for devices with actual IMU reported. Issue discovered on Rocknix Mangmi Air X build where the polling was causing high CPU usage in box86/64.

### DIFF
--- a/src/input/target/dualsense.rs
+++ b/src/input/target/dualsense.rs
@@ -124,7 +124,10 @@ pub struct DualSenseDevice {
     context: u8,
     hardware: DualSenseHardware,
     queued_events: Vec<ScheduledNativeEvent>,
+    /// Last report written, used to skip re-emitting identical reports.
     last_written: Option<Vec<u8>>,
+    /// Whether a consumer has the device open, writes are skipped while closed.
+    is_open: bool,
 }
 
 impl DualSenseDevice {
@@ -139,6 +142,7 @@ impl DualSenseDevice {
             hardware,
             queued_events: Vec::new(),
             last_written: None,
+            is_open: false,
         })
     }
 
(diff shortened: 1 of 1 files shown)


src/input/target/dualsense.rs
```diff
@@ -193,26 +195,21 @@ impl DualSenseDevice {
...
 
-                // Write the state to the virtual HID
-                if let Err(e) = self.device.write(&data) {
-                    let err = format!("Failed to write input data report: {:?}", e);
-                    return Err(err.into());
-                }
-            }
-            PackedInputDataReport::Bluetooth(state) => {
-                let data = state.pack()?;
+        if self.last_written.as_ref() == Some(&data) {
```
pastaq, line 214: Please add a comment why this is happening for future reference

src/input/target/dualsense.rs
```diff
@@ -137,6 +138,7 @@ impl DualSenseDevice {
             context: Default::default(),
             hardware,
             queued_events: Vec::new(),
+            last_written: None,
```
pastaq, line 144: This was an early driver, so we didn't have comments on the variables yet here. For at least the new addition please add a comment about what it stores and why.


If you're feeling froggy I'll also take additional comments for the existing ones, but adding that isn't a blocker and I'll do it if you don't.

src/input/target/dualsense.rs
```diff
@@ -1073,6 +1070,7 @@ impl TargetOutputDevice for DualSenseDevice {
             // send UHID_INPUT events to the kernel.
             uhid_virt::OutputEvent::Open => {
                 log::debug!("Open event received");
+                self.last_written = None;
```
pastaq: This will be none by default, so it's a no-op at initial opening. Then we carry that stale data in memory until the next time it is opened. Rather than do this here, does it make more sense to do it after close?

pastaq (changes requested):
Seems fine from a code perspective. Does this match the behavior of the real hardware?

sunshineinabox:
I don't have a real controller to actually test this on, if there is proper motion control this will likely be no behavior change as without fuzz it would constantly report anyway. This is effectively only useful for devices that are emulating the controller without motion sensors. If someone with real DS5 could look into real hardware behavior I would appreciate it.

src/input/target/dualsense.rs
sunshineinabox: Done

src/input/target/dualsense.rs
sunshineinabox: Documented above now

src/input/target/dualsense.rs
sunshineinabox: Went with doing it around close. Only issue is resetting on close alone doesn't really work since ``write_state`` still runs every poll even with nothing attached, so it just repopulates ``last_written`` before the next open and the reconnecting consumer can get its first report skipped as unchanged. So I also skip ``write_state`` while closed and reset ``last_written`` on close.

pastaq (approved):

Issue #584, opened 2026-05-04
Legion Go 2: scroll wheel goes one way only.


Patola:
package version: inputplumber 0.76.1-1.1 being used on CachyOS handheld.

The scroll wheel works but it goes only one way. Both scrolling up and scrolling down seemingly give the same result on applications and games, seemingly only the scroll down signal is sent either way.

I am not sure how to demonstrate that with text-based utilities and/or logs though. I can record games and applications if that is the case.

The best I could come up with was using libinput:

```bash
❯ sudo libinput debug-events
-event3   DEVICE_ADDED                 Video Bus                         seat0 default group1  cap:k
-event2   DEVICE_ADDED                   Legion Controller for Windows  Keyboard seat0 default group2  cap:k
-event4   DEVICE_ADDED                   Legion Controller for Windows  Mouse seat0 default group2  cap:p left scroll-nat scroll-button
-event5   DEVICE_ADDED                   Legion Controller for Windows  Touchpad seat0 default group2  cap:t  size 28x28mm ntouches 1 calib
-event10  DEVICE_ADDED                 Ideapad extra buttons             seat0 default group3  cap:k
-event8   DEVICE_ADDED                 GXTP6933:00 27C6:0133             seat0 default group4  cap:t  size 190x119mm ntouches 10 calib
-event0   DEVICE_ADDED                 Power Button                      seat0 default group5  cap:k
-event1   DEVICE_ADDED                 AT Translated Set 2 keyboard      seat0 default group6  cap:k
-event16  DEVICE_ADDED                 InputPlumber Mouse                seat0 default group7  cap:p left scroll-nat scroll-button
-event17  DEVICE_ADDED                 InputPlumber Keyboard             seat0 default group8  cap:k
-event18  DEVICE_ADDED                 BX17 (AVRCP)                      seat0 default group9  cap:k
-event16  POINTER_BUTTON               +0.000s  BTN_MIDDLE (274) pressed, seat count: 1
 event16  POINTER_BUTTON               +0.192s  BTN_MIDDLE (274) released, seat count: 0
 event16  POINTER_SCROLL_WHEEL         +1.128s  vert 15.00/120.0* horiz 0.00/0.0 (wheel)
 event16  POINTER_SCROLL_WHEEL      2  +1.152s  vert 1920.00/15360.0* horiz 0.00/0.0 (wheel)
[...]

PR #602, opened 2026-06-07
fix(Hardware Support): Fix Legion Go 2 mouse wheel #584

Patola:
Fix issue #584 
Co-developed-by: Claude Sonnet 4.6

The following part is AI-generated, the Claude Code explanation:

> The Legion Go 2 uses a different encoding for the scroll wheel (mouse_z) compared to the original Legion Go 1:
> 
> - Legion Go 1: mouse_z centered at 128 (neutral=128, deltas around it)
> - Legion Go 2: mouse_z is a 7-bit signed value pulsed from 0
>     scroll up   = 0x01 (+1 in 7-bit two's complement)
>     scroll down = 0x7f (-1 in 7-bit two's complement, bit 6 = sign bit)
>   The device resets mouse_z back to 0x00 after each pulse.
> 
> The old code used wrapping_sub(128) which was correct for Gen 1 but produced wrong sign and wrong magnitude for Gen 2:
> - Scroll up   (0x01) → wrapping_sub → -127 → kernel REL_WHEEL -127
> - Scroll down (0x7f) → wrapping_sub → match arm → -1  → kernel REL_WHEEL -1
> - Reset to 0         → wrapping_sub → -128 → spurious extra scroll event
> 
> All kernel values were negative, so content always scrolled one direction. The reset event also doubled every scroll event in libinput.
> 
> Fix: interpret mouse_z as a 7-bit signed integer by sign-extending bit 6 to bit 7, skip the 0x00 reset pulse, and use the value directly:
> - 0x01 → +1 → REL_WHEEL +1 → traditional scroll up   ✓
> - 0x7f → -1 → REL_WHEEL -1 → traditional scroll down ✓
> 
> Tested on Legion Go 2 (product 83N0/83N1) in CachyOS desktop mode and in gaming mode with The Planet Crafter and The Elder Scrolls IV: Oblivion Remastered.

There is additional data that Claude Code made on the PR to the main repository (which I denied) which expands on the explanation and test cases:

> 
> 
> ## Test plan
> 
> - [x] Scroll up moves content up, scroll down moves content down
> - [x] Each physical click produces exactly one libinput scroll event (no spurious duplicate)
> - [x] Tested on Legion Go 2 (DMI product `83N0`) running CachyOS
> - [x] Verified in desktop mode and in gaming mode with The Planet Crafter and The Elder Scrolls IV: Oblivion Remastered (mouse/keyboard emulation profile)
> - [ ] Regression test on original Legion Go 1 (Gen 1 uses 0x80 neutral — the new code path only activates for non-zero mouse_z, so Gen 1 values near 0x80 like 0x7f/0x81 will now be sign-extended differently; Gen 1 owners should verify)
> 
> ## Reviewer notes
> 
> The Gen 1 and Gen 2 share the same `lego` driver. The Gen 1 encoding (centered at 128) would now be interpreted differently by this code — a Gen 1 scroll-down of 0x7f (127, = -1 after wrapping_sub) would now sign-extend to -1, which happens to be the same result, so typical ±1 click values are unaffected. Larger multi-click deltas may differ. If a Gen 1 owner can test, that would be ideal.

So that summarizes well with what we went through. Initially for this troubleshooting I have used the same command that I have used in my bug report, which is `sudo libinput debug-events` which is quite verbose. Please let me know whether you want the complete history of the troubleshooting and fix session, I am not pasting it here because it seems excessive and not particularly informative.

### DIFF
--- a/src/drivers/lego/driver.rs
+++ b/src/drivers/lego/driver.rs
@@ -407,15 +407,17 @@ impl Driver {
                     value: state.a_trigger_r,
                 })));
             }
-            if state.mouse_z != old_state.mouse_z {
-                log::debug!("Raw mouse value: {:b}, {}", state.mouse_z, !state.mouse_z);
-                let value = state.mouse_z.wrapping_sub(128) as i8;
-                let value = match value {
-                    64..=127 => value - 127 - 1,
-                    v => v,
+            // mouse_z is a 7-bit signed delta (bit 6 = sign bit) pulsed from 0:
+            //   scroll up   = 0x01 (+1 in 7-bit)
+            //   scroll down = 0x7f (-1 in 7-bit)
+            // Positive value → REL_WHEEL positive → traditional scroll up.
+            if state.mouse_z != old_state.mouse_z && state.mouse_z != 0 {
+                let value = if state.mouse_z & 0x40 != 0 {
+                    (state.mouse_z | 0x80) as i8 // sign-extend bit 6 → bit 7
+                } else {
+                    state.mouse_z as i8
                 };
-
-                log::debug!("Normalized mouse value: {value}");
+                log::debug!("Mouse wheel value: {value}");
                 events.push(Event::Trigger(TriggerEvent::MouseWheel(MouseWheelInput {
                     value,
                 })));


pastaq:
After discussing with @ShadowApex again yesterday, we're going to be updating our policy on AI usage to not allow it for the time being, and we'll also be removing any existing AI assisted contributions and redeveloping them ourselves. This policy will be revisited after the legal questions about FOSS licensing and LLM generated code are answered. We're primarily concerned about copyleft contamination braking the GPL license.

Patola:
I understand. Thanks for the consideration though. I'm happy with using my fork while a fix on the official repository is not done yet. Cheers!

PR #607, opened 2026-06-14
fix(ROG Ally): invert gyro yaw for deck-uhid target

lonsdaleite:
## Summary
Fix inverted gyro yaw (left/right rotation) on ASUS ROG Ally devices when using the `deck-uhid` target.
## Problem
On ROG Ally / Ally X / Xbox Ally, gyro yaw is reversed with `deck-uhid` compared to `ds5-edge`. The IMU source is fine; the mismatch is in how yaw is written to the Deck UHID report versus how hid-steam exposes it as ABS_RZ.
## Changes
- Add `invert_gyro_yaw` to `SteamDeckConfig`
- Apply yaw sign correction when writing gyro data in `steam_deck_uhid`
- Enable inversion only for Ally DMI board names: `RC71L`, `RC72LA`,  `RC73XA`, `RC73YA` (from the existing `50-rog_*.yaml` profiles). This avoids affecting other handhelds that also use a BMI323 IIO source (e.g. Ayaneo). Detection uses runtime DMI `board_name`, so it still applies when the composite profile name is changed for controller spoofing.
## Test plan
- [x] Tested manually on ASUS ROG Xbox Ally X with `deck-uhid` target
- [x] Verified left/right gyro rotation in `inputplumber device 0 test`
- [x] Compared behavior against `ds5-edge` target (yaw direction matches)

### DIFF
--- a/src/input/target/steam_deck.rs
+++ b/src/input/target/steam_deck.rs
@@ -56,6 +56,10 @@ pub struct SteamDeckConfig {
     pub vendor: String,
     pub name: String,
     pub product_id: ProductId,
+    /// Invert gyro yaw written to the Deck HID report. Enabled only on ASUS ROG
+    /// Ally hardware (see `GYRO_YAW_INVERT_BOARD_NAMES` in steam_deck_uhid.rs)
+    /// when using the deck-uhid target.
+    pub invert_gyro_yaw: bool,
 }
 
 impl Default for SteamDeckConfig {
@@ -64,6 +68,7 @@ impl Default for SteamDeckConfig {
             vendor: "InputPlumber".to_string(),
             name: "Generic Steam Controller".to_string(),
             product_id: ProductId::Generic,
+            invert_gyro_yaw: false,
         }
     }
 }
--- a/src/input/target/steam_deck_uhid.rs
+++ b/src/input/target/steam_deck_uhid.rs
@@ -16,6 +16,7 @@ use uhid_virt::{Bus, CreateParams, StreamError, UHIDDevice};
 
 use crate::{
     config::CompositeDeviceConfig,
+    dmi::get_dmi_data,
     drivers::steam_deck::{
         hid_report::{
             PackedHapticReport, PackedInputDataReport, PackedRumbleReport, ReportType,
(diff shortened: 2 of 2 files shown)


pastaq:
This makes no sense to me. There are three distinct possibilities for this:
- 1.) The ds5 yaw is wrong for all devices.
- 2.) The deck-uhid is wrong for all devices.
- 3.) The IMU matrix is wrong for the ROG Ally along with one of the previous two, leading to an inverted conclusion of which is wrong.

Given that SteamOS officially released on the Go S and was validated by Valve, and that it has been tested as 1:1 compatible with the deck as a source device, I think it's much more likely that 1 and 3 are more likely.

lonsdaleite:
I didn't update Ally's matrix because I checked ds5 emulation and it worked correctly. But I definitely didn't think that maybe there are two problems exist simultaneously (1 and 3 from your message). 
Unfortunately I don't have any another handheld, so I can't check it :(

lonsdaleite:
@pastaq Hello!

Tested on the same Ally X with external controllers routed through InputPlumber, judged in-game (Talos Principle 2, gyro on, default settings, unchanged between runs):

- DualSense → deck-uhid: ok
- DualShock 4 → deck-uhid: ok
- DualShock 4 → ds5: ok
- Ally internal gyro → ds5: ok
- Ally internal gyro → deck-uhid: inverted

So options 1 and 2 are out - neither target flips yaw for every device. Option 3 doesn't fit either: an inverted Ally source would show through both targets, but through ds5 it's fine. Only the IIO source + deck-uhid combination inverts, and no hidraw source reproduces it.

That combination is exactly what this PR patches, so it does fix the reported symptom. What I don't know is how wide the problem is: does it hit only the Ally, or every handheld whose gyro comes from a built-in sensor (the iio source type)? If it's the latter, the sign belongs in one place on that source path and would fix all of them at once, instead of a board-name list that has to grow with every new model. 

**UPD**
Checked it: the inversion follows the emulated product id, not the source or the target.
Same Ally, same gyro, same deck-uhid - under 0x12f0 Generic yaw is correct, under 0x12fd Ally (and 0x12fe/0x12ff Lenovo) it's inverted. So it's Steam's per-PID handling, not InputPlumber.

pastaq:
> Checked it: the inversion follows the emulated product id, not the source or the target.
Same Ally, same gyro, same deck-uhid - under 0x12f0 Generic yaw is correct, under 0x12fd Ally (and 0x12fe/0x12ff Lenovo) it's inverted. So it's Steam's per-PID handling, not InputPlumber.

Thanks, I'll report it to Valve. Closing.

PR #610, opened 2026-06-18
fix(Hardware Support): Fix Legion Go Regressions

### LINKED ISSUE #597: Legion GO gyro is not responding.
With this command I confirmed gyro hardware is working.

```
watch -n 0.2 'cat /sys/bus/iio/devices/iio:device2/in_anglvel_x_raw /sys/bus/iio/devices/iio:device2/in_anglvel_y_raw /sys/bus/iio/devices/iio:device2/in_anglvel_z_raw 2>/dev/null || echo "IIO raw files not exposed for direct polling"'
``` 

In this test gyro is not responding.

```
inputplumber device 0 test
```

Distro: bazzite-deck:unstable
Kernel: Linux 7.0.9-ogc3.2.fc44.x86_64

If you need any more data for this I can provide.

pastaq:
Fixes multiple regressions on the Legion Go 1 with original firmware by doing the following:
- Split original Go 1 firmware and new Go 2 firmware into separate drivers.
- Remove unsupported events (IMU events, M1, Alt-Tab, and Show Desktop buttons), from the Go 1 driver.
- Remove default event filter from the Go 1 driver.
- Remove Screen based IMU from the Go 2 configuration.

Fixes #597

### DIFF
--- a/src/drivers/lego/go1_driver.rs
+++ b/src/drivers/lego/go1_driver.rs
@@ -5,32 +5,22 @@ use hidapi::HidDevice;
 use packed_struct::PackedStruct;
 use tokio::time::Instant;
 
-use crate::drivers::lego::DEFAULT_EVENT_FILTER;
-use crate::input::capability::Source;
+use crate::input::capability::Capability;
 use crate::udev::device::UdevDevice;
-use crate::{
-    drivers::lego::{
-        event::{ImuAxisInput, TouchAxisInput, TouchButtonEvent},
-        hid_report::GamepadMode,
-        CLICK_DELAY, PAD_FORCE_NORMAL, RELEASE_DELAY, XINPUT_COMMAND_ID,
-    },
-    input::capability::Capability,
-};
 
 use super::{
     event::{
         AxisEvent, BinaryInput, Event, GamepadButtonEvent, JoyAxisInput, MouseWheelInput,
-        TriggerEvent, TriggerInput,
+        TouchAxisInput, TouchButtonEvent, TriggerEvent, TriggerInput,
     },
-    hid_report::XInputDataReport,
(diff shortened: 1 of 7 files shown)


src/drivers/lego/go1_driver.rs
```diff
@@ -0,0 +1,468 @@
...
+        let report_id = buf[0];
+        let command_id = buf[2];
+
+        // Configuration event responses happen on the same endpoint. If this data packet isn't
+        // specifically xinput data it can crash the driver, so block it.
+        if command_id != XINPUT_COMMAND_ID {
+            //log::trace!("Got event that isn't xinput data, skipping");
+            return Ok(vec![]);
+        }
+        let slice = &buf[..bytes_read];
```
ShadowApex, line 94: It's possible this could crash if `bytes_read` is ever larger than `XINPUT_PACKET_SIZE`. We can prevent this by comparing the bytes read before reading the slice from the buffer.

src/drivers/lego/go1_driver.rs
```diff
@@ -0,0 +1,468 @@
...
+    ) -> Result<Vec<Event>, Box<dyn Error + Send + Sync>> {
+        let input_report = XInputDataReport::unpack(&buf)?;
+
+        // Print input report for debugging
+        //log::debug!("--- Input report ---");
+        //log::debug!("{input_report}");
+        //log::debug!(" ---- End Report ----");
+
+        // Update the state
+        let old_dinput_state = self.update_xinput_state(input_report);
```
ShadowApex: Should this be `old_xinput_state`?

ShadowApex (changes requested):
A few comments, but otherwise looks good 👍

ShadowApex (approved):

PR #620, opened 2026-07-06
feat(Hardware Support): Add MSI Claw 8 EX AI+ profile

matte-schwartz:
Also fix all the Claws to use board_name, like we do everywhere else.


matte-schwartz:
well, it's nice that that test caught that issue. just force-pushed the fix.

ShadowApex (approved):
Looks good, thank you for this!

PR #624, opened 2026-07-11
fix(Hardware Support): Add 8bitdo Pro3

HikariKnight:
Based this on the 8bitdo Ultimate 2 file as @pastaq suggested.

Blocking the DInput evdev seems to be enough to get this controller working, might be able to reuse support for the 8bitdo ultimate 2 when that is merged, as the controllers are functionally very similar.

Without this, the controllers inputs are messed up
IMG_20260711_162627

If there is anything else i can provide to get proper support added, i am happy to provide it

### DIFF
--- a/rootfs/usr/share/inputplumber/devices/60-8bitdo_pro3.yaml
+++ b/rootfs/usr/share/inputplumber/devices/60-8bitdo_pro3.yaml
@@ -1,0 +1,55 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ShadowBlip/InputPlumber/main/rootfs/usr/share/inputplumber/schema/composite_device_v1.json
+# Schema version number
+version: 1
+
+# The type of configuration schema
+kind: CompositeDevice
+
+# Name of the composite device mapping
+name: 8BitDo Pro 3
+
+# Only use this profile if *any* of the given matches matches. If this list is
+# empty, then the source devices will *always* be checked.
+# /sys/class/dmi/id/product_name
+matches: []
+
+# Only allow a single source device per composite device of this type.
+single_source: false
+
+# Maximum number of source devices per CompositeDevice.
+maximum_sources: 2
+
+# One or more source devices to combine into a single virtual device. The events
+# from these devices will be watched and translated according to the key map.
+source_devices:
+  # Dinput/SteamInput Mode
+  - group: gamepad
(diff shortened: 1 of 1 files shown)


rootfs/usr/share/inputplumber/devices/60-8bitdo_pro3.yaml
```diff
@@ -0,0 +1,55 @@
...
+      name: "8BitDo Pro 3"
+      vendor_id: "2dc8"
+      product_id: "6009"
+    capability_map_id: dinput_generic
+  - group: gamepad
+    hidraw:
+      vendor_id: 0x2dc8
+      product_id: 0x6009
+      interface_num: 0
+  # XInput/Receiver Mode
```
pastaq, line 38: This section is carried over from the ultimate 2 and should be removed

pastaq (changes requested):
This is the opposite of what we want to do. By blocking the evdev without an hidraw driver we are losing all input for this controller through InputPlumber. It doesn't make sense to ship a config to bypass InputPlumber within InputPlumber, and it removes the ability for OpenGamepadUI to see the input and open the quick bar. 

The actual problem is that SDL is directly reading the hidraw as well, so blocking that instead will prevent the duplicate input issue and allow evdev events to propagate through InputPlumber as expected. Then, once a driver is written for it, we can flip the block to this pattern for the full functionality provided by the controller.

HikariKnight:
not blocking the evdev device and instead block hidraw resulted in
* AB being swapped,
* XY being swapped,
* L4 and R4 not working,
* Dpad not working,
* Still getting double inputs (except A does AB and B does BA and so forth),

so i guess just moving the config to bazzite would be the best solution for now and then when you're back from vacation i can try help with anything you need for getting a proper driver made.

evtest shows it as dinput (makes sense since the switch on the controller in order to use steam input native requires you to set it in dinput mode)

and in terms of losing the intercept for ogui, i think for the average user its more important they have a functional controller over (hopefully) temporary losing ogui interception of buttons.

especially for the popular 3rd party controllers like 8bitdo, gulikit, flydigi, etc.

But that config would make more sense to have on the distro side until a hidraw driver is made and i had misunderstood you in earlier discussions.

pastaq:
It might be as simple as adding the capability_map_id for dinput_generic to the controller

PR #625, opened 2026-07-11
fix(Hardware Support): Add HandHeld Legend GC Ultimate

HikariKnight:
blocking the DInput evdev makes the controller function as expected in steam.
This device is a native DInput/Steam input device when used in wireless mode, with duplex triggers and gyro
IMG_20260711_162741

Again if more info is wanted i will happily provide it, although this is an extremely niche controller

### DIFF
--- a/rootfs/usr/share/inputplumber/devices/60-hhl_gc_ultimate.yaml
+++ b/rootfs/usr/share/inputplumber/devices/60-hhl_gc_ultimate.yaml
@@ -1,0 +1,43 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ShadowBlip/InputPlumber/main/rootfs/usr/share/inputplumber/schema/composite_device_v1.json
+# Schema version number
+version: 1
+
+# The type of configuration schema
+kind: CompositeDevice
+
+# Name of the composite device mapping
+name: HHL GC Ultimate
+
+# Only use this profile if *any* of the given matches matches. If this list is
+# empty, then the source devices will *always* be checked.
+# /sys/class/dmi/id/product_name
+matches: []
+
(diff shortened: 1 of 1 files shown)


rootfs/usr/share/inputplumber/devices/60-hhl_gc_ultimate.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+# Maximum number of source devices per CompositeDevice.
+maximum_sources: 2
+
+# One or more source devices to combine into a single virtual device. The events
+# from these devices will be watched and translated according to the key map.
+source_devices:
+  # Dinput/SteamInput Mode
+  - group: gamepad
+    blocked: true
```
pastaq, line 27: Again, blocking this will bypass InputPlumber, which isn't something we're going to ship in InputPlumber.

rootfs/usr/share/inputplumber/devices/60-hhl_gc_ultimate.yaml
```diff
@@ -0,0 +1,43 @@
...
+    evdev:
+      name: "SInput Gamepad"
+      vendor_id: "2e8a"
+      product_id: "10dd"
+    capability_map_id: dinput_generic
+  - group: gamepad
+    hidraw:
```
pastaq, line 34: Without a driver this part of the config won't do anything. Block this instead and we can develop a source device driver to unlock all capabilities.

pastaq (changes requested):

HikariKnight:
see https://github.com/ShadowBlip/InputPlumber/pull/624#issuecomment-4951174289 as results are similar
changing to draft for now, maybe closing would be better?

Issue #552, opened 2026-03-12
Add support for OneXPlayer Apex


pastaq:
OneXPlayer released a new handheld, add support for it.

It appears to follow the same mapping as the F1 series.

[evtest-apex.txt](https://github.com/user-attachments/files/25943022/evtest-apex.txt)
[dmi-2.log.txt](https://github.com/user-attachments/files/25943056/dmi-2.log.txt)
[devices-1.log.txt](https://github.com/user-attachments/files/25943057/devices-1.log.txt)
[tty_devs.log](https://github.com/user-attachments/files/25943063/tty_devs.log)

PR #626, opened 2026-07-11
fix(Hardware Support): Add updated OXP Apex config

pastaq:
Adds new phys paths present on updated frimware for OneXPlayer Apex devices.


pastaq:
Closes #552

PR #627, opened 2026-07-11
feat(Hardware Support): Add Ayaneo Pocket S2

Drakulix:
Adds capability maps and device conf for the Ayaneo Pocket S2 (Pro).

I am not 100% sure that everything in here is best practice as I had to go by other device files mostly where documentation was slightly lacking, but it has been tested and works on postmarketOS. (See https://gitlab.postmarketos.org/postmarketOS/pmaports/-/merge_requests/8979 if interested.)

Couple of notes:
- The device is an ARM device with an Android Bootloader and as such does not have DMI strings. As such matching is based on the upstream device tree.
- The device exposes it's internal controller as two devices in "Ayaneo Mode". One is exposed a as a normal "controller", while the other handes all "special" buttons, which is likely done so that the ayaneo utility on android can grab the second controller, while the first is picked up by games.
  - (There is a way to communicate over a serial interface with the internal controller to expose the device as a single "Xbox" controller, but that disables a couple of buttons and does not really need a inputplumber profile.)
- As such I created a capability-file per device. I am not sure that is the best way to handle this or if the files could be combined somehow?
- I am mostly targetting steam with this so I am using "deck-uhid" as the default target device. I guess steamos-manager already puts the device into this target mode, so maybe something like "xbox-series" or "xbox-elite" would be a better default?

Thanks for this utility!

### DIFF
--- a/rootfs/usr/share/inputplumber/capability_maps/ayaneo_type9.yaml
+++ b/rootfs/usr/share/inputplumber/capability_maps/ayaneo_type9.yaml
@@ -1,0 +1,51 @@
+version: 2
+kind: CapabilityMap
+name: AYANEO Type 9
+id: aya9
+
+mapping:
+    - name: Ayaneo Button to QuickAccess
+      source_events:
+          - evdev:
+                event_type: KEY
+                event_code: BTN6
+                value_type: button
+      target_event:
+          gamepad:
+              button: QuickAccess
+    - name: Ayaneo Button2 to Screenshot
+      source_events:
+          - evdev:
+                event_type: KEY
+                event_code: BTN5
+                value_type: button
+      target_event:
+          gamepad:
+              button: Screenshot
+    - name: L3
+      source_events:
+          - evdev:
+                event_type: KEY
+                event_code: BTN3
+                value_type: button
+      target_event:
+          gamepad:
+              button: LeftPaddle1
+    - name: R3
+      source_events:
+          - evdev:
+                event_type: KEY
+                event_code: BTN4
+                value_type: button
+      target_event:
+          gamepad:
+              button: RightPaddle1
+    - name: Side button to Keyboard
+      source_events:
+          - evdev:
+                event_type: KEY
+                event_code: BTN7
+                value_type: button
+      target_event:
+          gamepad:
+              button: Keyboard
--- a/rootfs/usr/share/inputplumber/devices/50-ayaneo_pocket_s2.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-ayaneo_pocket_s2.yaml
@@ -1,0 +1,43 @@
+version: 1
+kind: CompositeDevice
+name: AYANEO Pocket S2 (Pro)
+
+maximum_sources: 0
+options:
+    auto_manage: true
+
+matches:
+    - udev:
+          sys_path: /sys/firmware/devicetree/base
+          attributes:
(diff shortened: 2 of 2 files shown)


rootfs/usr/share/inputplumber/capability_maps/ayaneo_type9.yaml
pastaq, line 1: This looks very similar to the default dinput capability map. Can you check if all these are mapped the same on that map and if so switch to it

rootfs/usr/share/inputplumber/capability_maps/ayaneo_type10.yaml
```diff
@@ -0,0 +1,51 @@
+version: 2
+kind: CapabilityMap
+name: AYANEO Type 10
```
pastaq: If the below comment is true, this will need to be incremented down, along with the ID and device profile ID

pastaq (changes requested):
Thanks for this. It looks good, just a couple things to look into. Note that not all capabilities need to be present on the capability map I mention below, you just need to validate if the ones used here are the same. 

For completeness I'd also like to include the xinput mode if it's something a use can easily switch to our boot into if it was selected prior. Even if it won't be used by most, it's better to do now and avoid potential issues in the future.

Edit: also, yes Xbox elite out similar is usually the best default device.

rootfs/usr/share/inputplumber/capability_maps/ayaneo_type9.yaml
Drakulix: done

rootfs/usr/share/inputplumber/capability_maps/ayaneo_type10.yaml
Drakulix: done

Drakulix:
> Thanks for this. It looks good, just a couple things to look into. Note that not all capabilities need to be present on the capability map I mention below, you just need to validate if the ones used here are the same.

Thanks, that was really helpful, missed that the generic map exist.

> For completeness I'd also like to include the xinput mode if it's something a use can easily switch to our boot into if it was selected prior. Even if it won't be used by most, it's better to do now and avoid potential issues in the future.

Added. I opted to use the `phys_path` for matching as the controller is always on the same pcie port and name,product_id,vendor_id are all equal to a legit xbox 360 gamepad. This was even in Xinput mode only the internal controller should match and additional ones connected via usb shouldn't be affected.

I tested with the new config and both modes and they behave identically.

> Edit: also, yes Xbox elite out similar is usually the best default device.

Fixed.

All of this made the PR much smaller, so thanks a bunch!

pastaq:
Changes look good. I just wanted to poke about the dinput capability map on the xinput device. It seems odd to me they would spoof an Xbox controller but retain the dinput button mapping. You said you tested it so I believe you, I just wanted to confirm that it is in fact needed on that device before I approve it.

Drakulix:
> Changes look good. I just wanted to poke about the dinput capability map on the xinput device. It seems odd to me they would spoof an Xbox controller but retain the dinput button mapping. You said you tested it so I believe you, I just wanted to confirm that it is in fact needed on that device before I approve it.

No you are right, this was a copy&paste error. Testing all inputs again the capability-map messed with the triggers. Removing it did the trick.

pastaq (approved):

PR #631, opened 2026-07-23
fix(Hardware Support): Fix Claw 8 EX Config

pastaq:
The Claw 8 EX has enough variation to warrant its own capability map. This improves detection of the guide button dramatically compared to the previous config.

### DIFF
--- a/rootfs/usr/share/inputplumber/capability_maps/msiclaw_type2.yaml
+++ b/rootfs/usr/share/inputplumber/capability_maps/msiclaw_type2.yaml
@@ -1,0 +1,52 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ShadowBlip/InputPlumber/main/rootfs/usr/share/inputplumber/schema/capability_map_v1.json
+# Schema version number
+version: 1
+
+# The type of configuration schema
+kind: CapabilityMap
+
+# Name for the device event map
+name: MSI Claw Type 2
+
+# Unique identifier of the capability mapping
+id: claw2
+
+# List of mapped events that are activated by a specific set of activation keys.
+mapping:
+  - name: Guide (Short)
+    source_events:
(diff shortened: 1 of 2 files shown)


rootfs/usr/share/inputplumber/capability_maps/msiclaw_type1.yaml
```diff
@@ -6,41 +6,33 @@ version: 1
 kind: CapabilityMap
 
 # Name for the device event map
-name: MSI Claw Type 1
+name: MSI Claw Type 2
```
LukeShortCloud: This `name` and `id` suggest it's type 2 but the file name is type 1.

rootfs/usr/share/inputplumber/capability_maps/msiclaw_type2.yaml
```diff
@@ -0,0 +1,60 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ShadowBlip/InputPlumber/main/rootfs/usr/share/inputplumber/schema/capability_map_v1.json
+# Schema version number
+version: 1
+
+# The type of configuration schema
+kind: CapabilityMap
+
+# Name for the device event map
+name: MSI Claw Type 1
```
LukeShortCloud: And vice-versa: the `name` and `id` suggest it's type 1 but the file name is type 2.

LukeShortCloud (changes requested):
I'm just double-checking to see if the file names are what you intended.

rootfs/usr/share/inputplumber/capability_maps/msiclaw_type1.yaml
pastaq: Good catch, in my haste I accidentally modified the original instead of the copy

PR #643, opened 2026-08-07
Xbox one BT support

strycore:
This adds support for the Xbox One controller connected via Bluetooth. Tested on GameOS, the controller is now working.


pastaq:
Has gameos tried the xpad neo driver? It's exclusively for Xbox One Bluetooth controllers. IMO we should drop this driver entirely and use the evdev device from the kernel driver

strycore:
we're going to update to a fc44 kernel, I'll test that first and see if the patch is still needed

strycore:
On the fc44 build of GameOS, the Xbox One controller works OOTB, no patch required.

PR #659, opened 2026-08-23
feat(Hardware Support): Add AYN Odin 3 and other RSInput devices

jenneron:
Adds support for AYN Odin 3.

This follows https://github.com/ShadowBlip/InputPlumber/pull/627 to make use of this in postmarketOS: https://wiki.postmarketos.org/wiki/AYN_Odin_3_(ayn-odin-3). I haven't tested it in UI yet (since https://gitlab.postmarketos.org/postmarketOS/pmaports/-/merge_requests/8979 is not merged), but `evtest` seems to produce expected results

It combines 2 input devices (gamepad + gpio-keys for paddles). It maps D-Pad, paddles and also maps back button to `QuickAccess` (please let me know whether it's a good choice). 

Note that this device kernel support is not upstreamed yet, so if anything changes during upstreaming there would have to be a follow-up PR. pmOS uses ROCKNIX kernel patches, so we should be on the same page between different distributions.

### DIFF
--- a/rootfs/usr/share/inputplumber/capability_maps/rsinput_type1.yaml
+++ b/rootfs/usr/share/inputplumber/capability_maps/rsinput_type1.yaml
@@ -1,0 +1,102 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ShadowBlip/InputPlumber/main/rootfs/usr/share/inputplumber/schema/capability_map_v2.json
+version: 2
+kind: CapabilityMap
+name: RSInput Type 1
+id: rsinput1
+
+mapping:
+  - name: D-Pad Up
+    source_events:
+      - evdev:
+          event_type: KEY
+          event_code: BTN_DPAD_UP
+          value_type: button
+    target_event:
+      gamepad:
+        button: DPadUp
+
+  - name: D-Pad Down
+    source_events:
+      - evdev:
+          event_type: KEY
+          event_code: BTN_DPAD_DOWN
+          value_type: button
+    target_event:
+      gamepad:
+        button: DPadDown
+
+  - name: D-Pad Left
+    source_events:
+      - evdev:
+          event_type: KEY
+          event_code: BTN_DPAD_LEFT
+          value_type: button
+    target_event:
+      gamepad:
+        button: DPadLeft
+
+  - name: D-Pad Right
+    source_events:
+      - evdev:
+          event_type: KEY
+          event_code: BTN_DPAD_RIGHT
+          value_type: button
+    target_event:
+      gamepad:
+        button: DPadRight
+
+  - name: Back Button
+    source_events:
+      - evdev:
+          event_type: KEY
+          event_code: BTN_BACK
+          value_type: button
+    target_event:
+      gamepad:
+        button: QuickAccess
+
+  - name: Left Back Paddle
+    source_events:
+      - evdev:
+          event_type: KEY
+          event_code: BTN_Z
+          value_type: button
+    target_event:
+      gamepad:
+        button: LeftPaddle1
+
+  - name: Right Back Paddle
+    source_events:
+      - evdev:
+          event_type: KEY
+          event_code: BTN_C
+          value_type: button
+    target_event:
+      gamepad:
+        button: RightPaddle1
+
+  # Kernel reports NORTH/WEST by physical position, but Xbox layout 
+  # relies on these swapped. This results in NORTH/WEST swapped in
+  # evtest, but produces expected behaviour in games.
+
+  - name: West Button
+    source_events:
+      - evdev:
+          event_type: KEY
+          event_code: BTN_NORTH
+          value_type: button
+    target_event:
+      gamepad:
+        button: West
+
+  - name: North Button
+    source_events:
+      - evdev:
+          event_type: KEY
+          event_code: BTN_WEST
+          value_type: button
+    target_event:
+      gamepad:
+        button: North
+
+filtered_events: []
--- a/rootfs/usr/share/inputplumber/devices/50-ayn_odin2.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-ayn_odin2.yaml
@@ -1,0 +1,33 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ShadowBlip/InputPlumber/main/rootfs/usr/share/inputplumber/schema/composite_device_v1.json
+version: 1
+kind: CompositeDevice
+name: AYN Odin 2
+
+options:
+  auto_manage: true
+
+matches:
+  - udev:
+      sys_path: /sys/firmware/devicetree/base
+      attributes:
+        - name: compatible 
+          value: ayn,odin2
+
+source_devices:
+  - group: gamepad
+    capability_map_id: rsinput1
+    evdev:
+      name: AYN Odin2 Gamepad
+      phys_path: rsinput-gamepad/input0
+      vendor_id: "2020"
+      product_id: "3001"
+      handler: event*
+
+  - group: gamepad
+    capability_map_id: rsinput1
+    evdev:
+      name: gpio-keys-paddles
+      handler: event*
+
+target_devices:
+  - xbox-elite
--- a/rootfs/usr/share/inputplumber/devices/50-ayn_odin2.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-ayn_odin2.yaml
@@ -12,6 +12,11 @@ matches:
       attributes:
         - name: compatible 
           value: ayn,odin2
+  - udev:
+      sys_path: /sys/firmware/devicetree/base
+      attributes:
+        - name: compatible
+          value: ayn,odin2portal
 
 source_devices:
   - group: gamepad
--- a/rootfs/usr/share/inputplumber/devices/50-ayn_odin3.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-ayn_odin3.yaml
@@ -1,0 +1,39 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ShadowBlip/InputPlumber/main/rootfs/usr/share/inputplumber/schema/composite_device_v1.json
+version: 1
+kind: CompositeDevice
+name: AYN Odin 3
+
+# TODO: Add support for gyroscope, see
+# https://github.com/ShadowBlip/InputPlumber/pull/590
+# This command proves gyroscope is present and supported with libssc:
+# ssccli --sensor=gyroscope
+
+options:
+  auto_manage: true
+
+matches:
+  - udev:
+      sys_path: /sys/firmware/devicetree/base
+      attributes:
+        - name: compatible
+          value: ayn,odin3
+
+source_devices:
+  - group: gamepad
+    capability_map_id: rsinput1
+    evdev:
+      name: AYN Odin3 Gamepad
+      phys_path: rsinput-gamepad/input0
+      vendor_id: "2020"
+      product_id: "3001"
+      handler: event*
+
+  - group: gamepad
+    capability_map_id: rsinput1
+    evdev:
+      name: gpio-keys-paddles
+      phys_path: gpio-keys/input0
+      handler: event*
+
+target_devices:
+  - xbox-elite
--- a/rootfs/usr/share/inputplumber/devices/50-retroid_pocket6.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-retroid_pocket6.yaml
@@ -1,0 +1,37 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ShadowBlip/InputPlumber/main/rootfs/usr/share/inputplumber/schema/composite_device_v1.json
+version: 1
+kind: CompositeDevice
+name: Retroid Pocket 6
+
+options:
+  auto_manage: true
+
+matches:
+  - udev:
+      sys_path: /sys/firmware/devicetree/base
+      attributes:
+        - name: compatible
+          value: retroidpocket,rp6
+
+source_devices:
+  - group: gamepad
+    evdev:
+      name: AYN Odin2 Gamepad
+      phys_path: rsinput-gamepad/input0
+      vendor_id: "2020"
+      product_id: "3001"
+      handler: event*
+    capability_map_id: rsinput1
+
+  - group: gamepad
+    evdev:
+      name: gpio-keys
+      phys_path: gpio-keys/input0
+      handler: event*
+    capability_map_id: rsinput1
+
+target_devices:
+  - xbox-elite
+  # gpio-keys node is shared between gamepad and volume up key,
+  # we need keyboard target here for volume up to work
+  - keyboard
--- a/rootfs/usr/share/inputplumber/devices/50-retroid_pocket_nova.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-retroid_pocket_nova.yaml
@@ -1,0 +1,37 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ShadowBlip/InputPlumber/main/rootfs/usr/share/inputplumber/schema/composite_device_v1.json
+version: 1
+kind: CompositeDevice
+name: Retroid Pocket Nova
+
+options:
+  auto_manage: true
+
+matches:
+  - udev:
+      sys_path: /sys/firmware/devicetree/base
+      attributes:
+        - name: compatible
+          value: retroidpocket,rpnova
+
+source_devices:
+  - group: gamepad
+    capability_map_id: rsinput1
+    evdev:
+      name: AYN Odin2 Gamepad
+      phys_path: rsinput-gamepad/input0
+      vendor_id: "2020"
+      product_id: "3001"
+      handler: event*
+
+  - group: gamepad
+    capability_map_id: rsinput1
+    evdev:
+      name: gpio-keys
+      phys_path: gpio-keys/input0
+      handler: event*
+
+target_devices:
+  - xbox-elite
+  # gpio-keys node is shared between gamepad and volume up key,
+  # we need keyboard target here for volume up to work
+  - keyboard


justradical:
Hi, this is missing this piece if its the rocknix DTBs, at least this is what we have in Armada. I'm working on upstreaming our devices as well https://github.com/justradical/InputPlumber/tree/armada-devices

```yaml
  - name: West Button
    source_events:
      - evdev:
          event_type: KEY
          event_code: BTN_NORTH
          value_type: button
    target_event:
      gamepad:
        button: West

  - name: North Button
    source_events:
      - evdev:
          event_type: KEY
          event_code: BTN_WEST
          value_type: button
    target_event:
      gamepad:
        button: North
```

jenneron:
> Hi, this is missing this piece if its the rocknix DTBs, at least this is what we have in Armada. I'm working on upstreaming our devices as well https://github.com/justradical/InputPlumber/tree/armada-devices

@justradical As far as I understand this is supposed to make physical Nintendo labels match Xbox ones? Please correct me if I'm wrong

In my opinion, we should match Xbox gamepad not by labels, but by key positions in order to fit Xbox ergonomics and muscle memory. If someone wants to match labels on their device they should swap physical buttons as this is what AYN advertises to do on their product page:

<details>
  <summary>Product page picture</summary>

image
  
</details>

jenneron:
Ah, I see, Xbox layout relies on north and west swapped

justradical:
> Ah, I see, Xbox layout relies on north and west swapped

yes the north and west keys are incorrectly labelled

jenneron:
> yes the north and west keys are incorrectly labelled

i've pushed the change with Co-authored-by, so this is fixed

justradical:
> > yes the north and west keys are incorrectly labelled
> 
> i've pushed the chage with Co-authored-by, so this is fixed

also have you tested that volume keys work without the keyboard target? Volume up doesnt function on the retroid pocket 6 without also targeting keyboard as an output

jenneron:
> also have you tested that volume keys work without the keyboard target? Volume up doesnt function on the retroid pocket 6 without also targeting keyboard as an output

no problems with volume keys when inputplumber is running, at least on phosh

> Volume up doesnt function on the retroid pocket 6 without also targeting keyboard as an output

maybe volume up on this device shares gpio-keys DT node with gamepad keys?

it's not the case on odin 3, it has two separated gpio-keys nodes:

- https://github.com/ROCKNIX/distribution/blob/6166516b9f5e6ae331eb98e9fce2f7349cb1422f/projects/ROCKNIX/devices/SM8750/patches/linux/0046-arm64-dts-qcom-Add-AYN-CQ8725S-Common.patch#L118-L132
- https://github.com/ROCKNIX/distribution/blob/6166516b9f5e6ae331eb98e9fce2f7349cb1422f/projects/ROCKNIX/devices/SM8750/patches/linux/0047-arm64-dts-qcom-Add-AYN-Odin3.patch#L363-L382

which results in:

```
I: Bus=0019 Vendor=0001 Product=0001 Version=0100
N: Name="gpio-keys"
P: Phys=gpio-keys/input0
S: Sysfs=/devices/platform/gpio-keys/input/input2
U: Uniq=
H: Handlers=kbd event2 
B: PROP=0
B: EV=3
B: KEY=8000000000000 0

I: Bus=0019 Vendor=0001 Product=0001 Version=0100
N: Name="gpio-keys-paddles"
P: Phys=gpio-keys/input0
S: Sysfs=/devices/platform/gpio-keys-paddles/input/input3
U: Uniq=
H: Handlers=event3 
B: PROP=0
B: EV=3
B: KEY=24000000000000 0 0 0 0
```

and this config only matches `name: gpio-keys-paddles`

rootfs/usr/share/inputplumber/devices/50-ayn_odin3.yaml
```diff
@@ -0,0 +1,41 @@
...
+      product_id: "3001"
+      handler: event*
+
+  - group: gamepad
+    capability_map_id: ayn2
+    evdev:
+      name: gpio-keys-paddles
+      phys_path: gpio-keys/input0
+      vendor_id: "0001"
+      product_id: "0001"
```
justradical: is there any reason to be specifying vendor_id and product_id here at all? This doesnt do anything in here

rootfs/usr/share/inputplumber/devices/50-ayn_odin3.yaml
```diff
@@ -0,0 +1,41 @@
...
+          value: ayn,odin3
+
+source_devices:
+  - group: gamepad
+    capability_map_id: ayn2
+    evdev:
+      name: AYN Odin3 Gamepad
+      phys_path: rsinput-gamepad/input0
+      vendor_id: "2020"
+      product_id: "3001"
```
justradical, line 28: kind of redundant with the devicetree but @pastaq or @ShadowApex should comment on that

rootfs/usr/share/inputplumber/capability_maps/ayn_type2.yaml
```diff
@@ -0,0 +1,102 @@
...
+          event_type: KEY
+          event_code: BTN_C
+          value_type: button
+    target_event:
+      gamepad:
+        button: RightPaddle1
+
+  # Kernel reports NORTH/WEST by phisical position, but Xbox layout 
+  # relies on these swapped. This results in NORTH/WEST swapped in
+  # evtest, but produces expected behaviour in games.
```
justradical: this note here is incorrect, it doesn't report by physical position, if it was by physical then the config wouldn't be needed, for some reason they arent for NORTH/WEST and have to be swapped in the config, SOUTH/EAST keys are correct, it might be a driver bug. but anyways, apart from this comment id say this lgtm

justradical:
> > also have you tested that volume keys work without the keyboard target? Volume up doesnt function on the retroid pocket 6 without also targeting keyboard as an output
> 
> no problems with volume keys when inputplumber is running, at least on phosh
> 
> > Volume up doesnt function on the retroid pocket 6 without also targeting keyboard as an output
> 
> maybe volume up on this device shares gpio-keys DT node with gamepad keys?
> 
> it's not the case on odin 3, it has two separated gpio-keys nodes:
> 
>     * [ROCKNIX/distribution@`6166516`/projects/ROCKNIX/devices/SM8750/patches/linux/0046-arm64-dts-qcom-Add-AYN-CQ8725S-Common.patch#L118-L132](https://github.com/ROCKNIX/distribution/blob/6166516b9f5e6ae331eb98e9fce2f7349cb1422f/projects/ROCKNIX/devices/SM8750/patches/linux/0046-arm64-dts-qcom-Add-AYN-CQ8725S-Common.patch#L118-L132)
> 
>     * [ROCKNIX/distribution@`6166516`/projects/ROCKNIX/devices/SM8750/patches/linux/0047-arm64-dts-qcom-Add-AYN-Odin3.patch#L363-L382](https://github.com/ROCKNIX/distribution/blob/6166516b9f5e6ae331eb98e9fce2f7349cb1422f/projects/ROCKNIX/devices/SM8750/patches/linux/0047-arm64-dts-qcom-Add-AYN-Odin3.patch#L363-L382)
> 
> 
> which results in:
> 
> ```
> I: Bus=0019 Vendor=0001 Product=0001 Version=0100
> N: Name="gpio-keys"
> P: Phys=gpio-keys/input0
> S: Sysfs=/devices/platform/gpio-keys/input/input2
> U: Uniq=
> H: Handlers=kbd event2 
> B: PROP=0
> B: EV=3
> B: KEY=8000000000000 0
> 
> I: Bus=0019 Vendor=0001 Product=0001 Version=0100
> N: Name="gpio-keys-paddles"
> P: Phys=gpio-keys/input0
> S: Sysfs=/devices/platform/gpio-keys-paddles/input/input3
> U: Uniq=
> H: Handlers=event3 
> B: PROP=0
> B: EV=3
> B: KEY=24000000000000 0 0 0 0
> ```
> 
> and this config only matches `name: gpio-keys-paddles`

seeing varying results with other armada devs on this, odin 2 portal seemingly needs it, thor doesnt, retroid pocket 6 does. Maybe AYN fixed something in their devices between the odin 2 and 3/thor

jenneron:
> seeing varying results with other armada devs on this, odin 2 portal seemingly needs it, thor doesnt, retroid pocket 6 does

are all of these devices upstream? we could patch dts

> Maybe AYN fixed something in their devices between the odin 2 and 3/thor

this is really trivial to change in kernel, i can help with this if needed, but we need to have a good reason for it, "it's better for userspace" doesn't always work in kernel development. i think we should use "consistency" as the reason to do this

justradical:
> > seeing varying results with other armada devs on this, odin 2 portal seemingly needs it, thor doesnt, retroid pocket 6 does
> 
> are all of these devices upstream? we could patch dts
> 
> > Maybe AYN fixed something in their devices between the odin 2 and 3/thor
> 
> this is really trivial to change in kernel, i can help with this if needed, but we need to have a good reason for it, "it's better for userspace" doesn't always work in kernel development. i think we should use "consistency" as the reason to do this

Not upstreamed afaik, we carry dts files and patches for them in our armada-packages repo, `armada-packages/kernel/dts`. On mobile now since its late so i cant send a link easily

rootfs/usr/share/inputplumber/devices/50-ayn_odin3.yaml
jenneron: agree about this one and previous, it should be able to work without it, although keeping these makes it more explicit documenting expected vendor/product ids. i'm happy to remove it if maintainers prefer

rootfs/usr/share/inputplumber/capability_maps/ayn_type2.yaml
jenneron: > it might be a driver bug

if it's a driver bug i would rather fix kernel instead, but i think my comment is correct

it seems that for some historical reason all xbox gamepads have to be "wrong" for userspace apps to understand them. i tested it with my 8bitdo gamepad and north/west are swapped in `evtest`, meanwhile odin 3 reports it correctly. I honestly don't understand why

jenneron: basically: 
- positioned north/west are correct for kernel - https://docs.kernel.org/input/gamepad.html
- swapped north/west are correct for userspace

justradical: Strange stuff..

rootfs/usr/share/inputplumber/capability_maps/rsinput_type1.yaml
```diff
@@ -0,0 +1,102 @@
...
+          event_type: KEY
+          event_code: BTN_C
+          value_type: button
+    target_event:
+      gamepad:
+        button: RightPaddle1
+
+  # Kernel reports NORTH/WEST by physical position, but Xbox layout 
+  # relies on these swapped. This results in NORTH/WEST swapped in
+  # evtest, but produces expected behaviour in games.
```
pastaq, line 80: Can you attach a picture of the ABXY with these mappings labeled?

rootfs/usr/share/inputplumber/devices/50-ayn_odin3.yaml
pastaq: please retain any identifying information. It helps looking back historically.

rootfs/usr/share/inputplumber/devices/50-ayn_odin3.yaml
pastaq: These are generic VID/PID so they can be dropped

jenneron: dropped

rootfs/usr/share/inputplumber/capability_maps/rsinput_type1.yaml
jenneron: <details>
  <summary>this is what this config results into in evtest</summary>

image
  
</details>

which matches my 8bitdo with xbox layout

justradical (approved):

pastaq (approved):

rootfs/usr/share/inputplumber/devices/50-ayn_odin3.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+  - group: gamepad
+    capability_map_id: ayn2
+    evdev:
+      name: gpio-keys-paddles
+      phys_path: gpio-keys/input0
+      handler: event*
+
+target_devices:
+  - xbox-elite
```
jenneron, line 39: @pastaq do you think this target device will be problematic for supporting gyroscope if https://github.com/ShadowBlip/InputPlumber/pull/590 gets merged?

SproedKartoffelChip: On Rocknix the Dualsense Edge was chosen as the target device for this reason (up until recently when the change had to be reverted temporarily for bindings to be dealt with on various emulators)

pastaq: Distros can ship a modified config in /etc/inputplumber/devices.d and there are multiple options for userspace to change the target at runtime (steamos-manager, OpenGamepadUI, deckyplumber, etc.). The preference is that glyphs match expectation for the end user. Horipad would be a preferable target device but it doesn't express e rumble on the rdesc. I'll revisit that soon.

jenneron: yeah, we have both rumble and gyroscope, and xbox glyphs are preferred

SproedKartoffelChip: Well the glyphs still don't exactly match the positions using the xbox-elite preset, given they're doing north/east/south/west -> X/A/B/Y instead of Y/B/A/X, not to mention that the triggers and bumpers on the Odin 3 use the PSX style L1/R1 and L2/R2 labelling.

AYN-ODIN3-BLACK-LISTING-DONE-02 jpg

It's a compromise either way and with Xbox you're sacrificing an additional form of input with the lack of gyro as well.

pastaq: We should probably just finish the switch pro controller target device. That would fix all the issues for this one

jenneron: > Well the glyphs still don't exactly match the positions using the xbox-elite preset, given they're doing north/east/south/west -> X/A/B/Y instead of Y/B/A/X

AYN supports xbox glyph configuration by switching them physically, it's advertised on their product page

SproedKartoffelChip: Not exactly an easy process to switch to the Xbox layout, I went through the process of swapping the buttons myself for third party PSX ones due to preferring the glossy look. It requires a partial teardown that I wouldn't personally call end user friendly, multiple small ribbons and various connectors for daughter boards have to be removed before you can get to the buttons which are *glued* down to the button rubber membrane pad, so you have to do some careful cutting with a knife to both detach the buttons themselves as well as reflatten the surface by trimming the glue residue left from the separation. 

handheld_stickbuttonmod

And on using the Switch Pro as a target, afaict that has no analog trigger capability, so racing titles and any other games that benefit greatly from added trigger input fidelity would suffer.

jenneron: why can't we have nice things?

does it make sense to add `ayn-odin` target? i believe it could be shared between odin 1/2/3, the only difference i can see is that odin 1 is missing back key

i suppose we can keep `xbox-elite` for now since gyroscope is not yet supported

pastaq: Again, this is just the default target. It's fully configurable through config files or GUI apps. Nintendo layout with PlayStation triggers is certainly a design choice. There are some 8bitdo and gulikit controllers we could emulate as well, but that is more long term

jenneron:
Renamed `ayn_type2.yaml` into `rsinput_type1.yaml` since it can be shared with non-AYN devices

jenneron:
Added other RSInput devices by @justradical

rootfs/usr/share/inputplumber/devices/50-retroid_pocket6.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - group: gamepad
+    evdev:
+      name: gpio-keys
+      phys_path: gpio-keys/input0
+      handler: event*
+    capability_map_id: rsinput1
+
+target_devices:
+  - xbox-elite
+  - keyboard
```
justradical, line 37: rp6 and rp nova didnt have a separate gpio-keys-paddles, so the keyboard target_device is needed here for volume up to work properly

rootfs/usr/share/inputplumber/devices/50-retroid_pocket_nova.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - group: gamepad
+    capability_map_id: rsinput1
+    evdev:
+      name: gpio-keys
+      phys_path: gpio-keys/input0
+      handler: event*
+
+target_devices:
+  - xbox-elite
+  - keyboard
```
justradical, line 37: same as above

rootfs/usr/share/inputplumber/devices/50-ayn_odin2.yaml
```diff
@@ -0,0 +1,38 @@
...
+      handler: event*
+
+  - group: gamepad
+    capability_map_id: rsinput1
+    evdev:
+      name: gpio-keys-paddles
+      handler: event*
+
+target_devices:
+  - xbox-elite
```
justradical, line 38: luckily the odin 2 did have this so it works without it

pastaq (changes requested):
Add missing keyboard targets per @justradical 

Other than that, no issues.

jenneron:
> Add missing keyboard targets per @justradical

I think those are just comments on changes, keyboard targets are in place

justradical:
> > Add missing keyboard targets per @justradical
> 
> I think those are just comments on changes, keyboard targets are in place

Yes, the comments just indicated where they are and why

pastaq (approved):

jenneron:
added comments about keyboard target

pastaq:

example:

pastaq (approved):
I just opened up issue https://github.com/ShadowBlip/InputPlumber/issues/690 that makes my request impossible with the v2 maps, so we'll defer for now.

Issue #628, opened 2026-07-14
Xbox ROG Ally X - Xbox and Armoury Create - Duplicate Steam menu button


FrostZG:
Good afternoon! SteamOS on the Asus Xbox Rog Ally X - duplicates the Xbox and Armoury Create keys. I would like to be able to reassign the Armoury Create key in the system to something else, for example, to create screenshots or videos, or anything else. Maybe there is some quick fix for this issue?

pastaq:
The fix is to create a new capability map for the Xbox Ally X that has the desired mapping and then change it in the device config. I'll but that the long press of the AC button is already mapped to QuickAccess2, which in the default profile is mapped to Screenshot already, so you might want to pick something else. Long press of CC should already be keyboard.

FrostZG:
> The fix is to create a new capability map for the Xbox Ally X that has the desired mapping and then change it in the device config. I'll but that the long press of the AC button is already mapped to QuickAccess2, which in the default profile is mapped to Screenshot already, so you might want to pick something else. Long press of CC should already be keyboard.

Thanks for the answer! It turns out that in SteamOS, pressing the Armoury Crate key for a long time does not work at the moment. Because long press CC gives only a small vibration and nothing else, and long press AC gives nothing at all:(

pastaq:
Can you run `inputplumber device 0 test` in an ssh, tty, or terminal, then press and hold the button and see if the corresponding events fire in the UI? It will probably be a quick flash if it works. If they don't, run
`sudo systemctl stop inputplumber && sudo libinput debug-events`, then long press the cc and ac buttons again and post the results?

FrostZG:
> Can you run `inputplumber device 0 test` in an ssh, tty, or terminal, then press and hold the button and see if the corresponding events fire in the UI? It will probably be a quick flash if it works. If they don't, run `sudo systemctl stop inputplumber && sudo libinput debug-events`, then long press the cc and ac buttons again and post the results?

Yes, inputplumber device 0 test It shows that when long press AC - directly KeyF21. CC long press - KeyF22.

But in game mode, long press doesn't work:(

pastaq:
Okay, then we can map those in the config as they're different events from older Ally's. 

Do they stay on when you hold them or flash on and off?

pastaq:
It looks like the non x version has the same issue. 
[...]

PR #662, opened 2026-08-23
withdrawn: remap Xbox Ally special buttons

nileshchakraborty:
## Withdrawn

This proposed mapping change was incorrect and must not be merged.

The existing `ally_type2` map from #629 is intentional:

- `KEY_F16` -> `Keyboard`
- `KEY_F21` -> `LeftTop`
- `KEY_PROG1` -> `QuickAccess`
- `KEY_F22` -> `RightTop`
- `KEY_F14` -> `LeftPaddle2`
- `KEY_F15` -> `RightPaddle2`
- target: `xbox-elite`, plus mouse and keyboard

The top/long/rear capabilities become assignable virtual L4/R4/L5/R5 controls in Steam Input. Changing `KEY_F16` to `QuickAccess`, `KEY_PROG1` to `Guide`, and the target to `deck-uhid` caused multiple special buttons to be interpreted as screenshots during testing.

The runtime patch was removed from the test device and this PR was closed. A tests-only follow-up protects the existing upstream behavior without changing device mappings.

### DIFF
--- a/rootfs/usr/share/inputplumber/capability_maps/ally_type2.yaml
+++ b/rootfs/usr/share/inputplumber/capability_maps/ally_type2.yaml
@@ -13,12 +13,12 @@ id: aly2
 
 # List of mapped events that are activated by a specific set of activation keys.
 mapping:
-  - name: Control Center (Short)
+  - name: Armory Crate (Short)
     source_events:
       - keyboard: KeyF16
     target_event:
       gamepad:
-        button: Keyboard
+        button: QuickAccess
   - name: Control Center (Long)
     source_events:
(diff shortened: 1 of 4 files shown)


nileshchakraborty:
Closing this PR. The Ally X button semantics and Steam Home behavior are not sufficiently resolved, and the current mapping can interfere with other controls. No change should be merged.

nileshchakraborty:
The corrected follow-up is tests-only: #663. It preserves the mapping and xbox-elite target already validated in #628/#629 and adds regression coverage so this incorrect remap is not reintroduced.

PR #663, opened 2026-08-23
test: lock Xbox Ally button mappings

nileshchakraborty:
## Summary

Add regression coverage for the existing ROG Xbox Ally / Xbox Ally X capability map and target selection introduced by #629.

## Why

The special controls intentionally expose distinct actions and virtual Steam Input controls:

| Source | InputPlumber capability |
| --- | --- |
| `KEY_F16` | `Keyboard` |
| `KEY_F21` | `LeftTop` |
| `KEY_PROG1` | `QuickAccess` |
| `KEY_F22` | `RightTop` |
| `KEY_F14` | `LeftPaddle2` |
| `KEY_F15` | `RightPaddle2` |

The `xbox-elite` target is also intentional: Steam presents the long-press controls and rear paddles as virtual L4/R4/L5/R5 inputs that users can assign without changing the raw hardware map.

PR #662 incorrectly tried to remap these actions and switch the target, which caused multiple special buttons to trigger screenshots. It was closed and none of those runtime changes are included here.

## Validation

```text
make in-docker TARGET=test
```

- 12 unit tests passed, including the two new mapping/config tests
- 4 doctests passed
- Clippy passed with `-D warnings`
- `git diff --check` passed

This PR changes tests only; shipped device behavior is unchanged.

### DIFF
--- a/src/config/capability_map.rs
+++ b/src/config/capability_map.rs
@@ -218,6 +218,44 @@ pub struct GamepadCapability {
     pub dial: Option<DialCapability>,
 }
 
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn xbox_ally_special_buttons_keep_their_distinct_capabilities() {
+        let config = CapabilityMapConfig::from_yaml_file(
+            "./rootfs/usr/share/inputplumber/capability_maps/ally_type2.yaml",
+        )
+        .expect("Ally Type 2 capability map should load");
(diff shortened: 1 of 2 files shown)


nileshchakraborty:
Closing this tests-only follow-up after deeper review. The existing mapping was already intentionally introduced and hardware-validated in #628/#629, and no current code regression needs fixing. Hard-coding that policy in unit tests would add maintenance burden without changing or correcting runtime behavior. #662 remains withdrawn.

Issue #655, opened 2026-08-21
Zotac gaming zone QAM and Guide /Steam Button don't work


Tayrint:
Hi this is the first time I write anything on github.
Don't know if I'm doing it wrong 😅.

Since updating to bazzite 44
The QAM (3 dots) button and the menu button one the left don't work anymore.
Short press or holding do nothing. 

The radial dials hasn't worked in bazzite 43 hoped it would would now but no. And back buttons don't work either.

whirlwind80:
I have the same Zotac Gaming Zone (G0A1W) and dug into this — opened a PR with fixes for the Steam/QAM button and back-button-related regression: #664

Root causes found:

1. **`50-zotac-zone.yaml`'s gamepad/keyboard evdev `name` fields don't match what the kernel actually reports** (`ZOTAC Gaming Zone Gamepad/Keyboard` vs the real `Zotac Technology Limited ZOTAC GAMING ZONE[...]`), so those source devices were never being matched at all.
2. **The Steam and QAM button capability bindings in `zone_type1.yaml` are swapped.** F17 (Steam button) was bound to `QuickAccess`, which synthesizes a Guide+South chord that gamescope treats as "open Quick Access Menu" — so the Steam button opened QAM instead of the main Steam menu. F18 (QAM button) was bound to `QuickAccess2`, which has zero evdev output on the xbox-elite target — a complete no-op. That's exactly the "menu button does nothing, QAM does something but wrong" behavior described here.
3. **Duplicate virtual controllers + missing View button.** The gamepad evdev source matches by name only, and that same name is reported on three different USB interfaces, so InputPlumber was creating three duplicate virtual Xbox controllers for one physical device. Separately, the real face buttons/sticks/View button are exposed by the kernel's built-in `xpad` driver on interface 1.0, which the config never referenced — Steam was reading that raw joystick directly instead of InputPlumber's translated device, which is why View didn't work either.

All three are fixed and verified on real hardware in #664 (Steam button now opens the main menu, QAM button opens QAM, View button works, only one virtual controller is created).

[...]

PR #664, opened 2026-08-24
fix(Zotac Zone): fix Steam/QAM/View buttons and duplicate virtual gamepad

whirlwind80:
## Summary

Four related fixes for the Zotac Gaming Zone (board G0A1W/G1A1W), found while
debugging why the physical Steam and QAM buttons stopped working after
switching from an HHD-based image to InputPlumber:

- **Device name mismatch** (`50-zotac-zone.yaml`): the gamepad and keyboard
  evdev source names didn't match what the kernel actually reports
  (`Zotac Technology Limited ZOTAC GAMING ZONE[...]`), so those source
  devices were never matched at all.
- **Steam/QAM capability bindings swapped** (`zone_type1.yaml`): F17 (Steam
  button) was bound to `QuickAccess`, which synthesizes the Guide+South
  chord that opens the Quick Access Menu — so the Steam button opened QAM
  instead of the main Steam menu. F18 (QAM button) was bound to
  `QuickAccess2`, which has no evdev output at all on the xbox-elite target,
  so it did nothing. Swapped so F17 -> `Guide` and F18 -> `QuickAccess`,
  matching how every other handheld's single QAM-style shortcut key is
  mapped elsewhere in this repo.
- **Duplicate composite device + missing View button** (`50-zotac-zone.yaml`):
  the gamepad evdev source matched purely by name, but that name is reported
  identically on USB interfaces 1.1, 1.2, and 1.3, causing InputPlumber to
  create three separate CompositeDevices (and three duplicate virtual Xbox
  controllers) for one physical controller. Added `phys_path` to pin the
  match to interface 1.1 only. Separately, the real face buttons / sticks /
  triggers / View (Select) button are exposed by the kernel's built-in
  `xpad` driver as a standard joystick on interface 1.0, which wasn't
  referenced by this config at all — Steam was reading that raw joystick
  directly instead of the translated virtual controller, so View never
  reached it. Added that node as a gamepad source so InputPlumber grabs and
  hides it from Steam and translates it like the rest of the controller.
- **Dial mappings swallowed the left touchpad** (`zone_type1.yaml`): the
  dials reach no evdev node — their HID usages are declared on the Generic
  Desktop page but hold Consumer page codes, so the kernel translates none
  of them — meaning the `REL_WHEEL`/`REL_HWHEEL` entries written for them
  never fired for the dials. They claimed events all the same: capability
  maps are applied per composite device rather than per source, so the wheel
  events of the left touchpad (a scroll surface, not a pointer one) became
  `LeftStickDial`/`RightStickDial`, which none of this device's targets
  consume — leaving that touchpad dead. This was latent until the first
  commit here corrected the mouse source's evdev name, which is what made
  that source match in the first place. Dropped the two mappings; reading
  the dials needs a hidraw report parser rather than a capability map.

## Test plan

All verified on real Zotac Gaming Zone (G0A1W) hardware running Bazzite:

- [x] `make test` (clippy + full test suite + autostart-rules check) passes
- [x] Steam button opens the main Steam menu
- [x] QAM button opens the Quick Access Menu
- [x] Analog sticks, ABXY, shoulder buttons, triggers, and D-pad work through
      a single virtual gamepad (previously duplicated 3x)
- [x] View (Select) button works
- [x] Only one `CompositeDevice`/virtual gamepad is created per physical
      controller (verified via `busctl` DBus introspection)
- [x] Left touchpad scrolls again, right touchpad still moves the pointer

## Follow-up

- Rear paddle (macro) buttons are fixed separately in #668, which applies to
  `main` independently of this PR — the two touch different files. The
  buttons turned out to ship with an empty mapping in the device's own
  configuration, so the firmware sends nothing at all for them until that
  mapping is written over the vendor config interface.
- Dial wheels remain unsupported. The protocol is known (report ID 3 on USB
  interface 1, one bit per dial and direction), but reading it needs a hidraw
  parser in the `zotac_zone` source driver and is not included here.

## Disclosure

Root-caused, implemented, and verified on hardware in collaboration with
Claude Sonnet 5 and Claude Opus 5 (Claude Code), which diagnosed the
capability-translation behavior in `xpad.rs`/`evdev.rs`, the
duplicate-device/raw-joystick behavior via `evtest`/`journalctl`/DBus
introspection on the running device, and the per-composite-device capability
map translation that was swallowing the touchpad's wheel events, and proposed
these fixes. I reviewed and understand every changed line.

### DIFF
--- a/rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
+++ b/rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
@@ -73,26 +73,17 @@ mapping:
       target_event:
           gamepad:
               button: RightPaddle1
-    - name: Left Dial
-      source_events:
-          - evdev:
-                event_type: REL
-                event_code: REL_HWHEEL
-                value_type: joystick_x
-      target_event:
-          gamepad:
-              dial:
-                  name: LeftStickDial
-    - name: Right Dial
-      source_events:
-          - evdev:
-                event_code: REL_WHEEL
-                event_type: REL
-                value_type: joystick_x
-      target_event:
-          gamepad:
-              dial:
-                  name: RightStickDial
+    # The dials are deliberately not mapped here. They reach no evdev node at
+    # all - their HID usages are declared on the Generic Desktop page but hold
+    # Consumer page codes, so the kernel translates none of them - which means
+    # a REL_WHEEL/REL_HWHEEL mapping never fired for the dials it was written
+    # for.
+    #
+    # It claimed events all the same: capability maps are applied per composite
+    # device rather than per source, so the wheel events of the left touchpad -
+    # which is a scroll surface, not a pointer one - were turned into a dial
+    # capability that none of this device's targets consume, leaving that
+    # touchpad dead.
 
 # List of events to filter from the source devices
 filtered_events: []
--- a/rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
+++ b/rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
@@ -22,9 +22,6 @@ mapping:
       target_event:
           gamepad:
               button: Guide
-    # Plain Guide press opens the main Steam menu. QuickAccess2 has no evdev
-    # output on the xbox-elite target (see event_codes_from_capability in
-    # src/input/event/evdev.rs), so binding this button to it was a no-op.
     - name: MORE
       source_events:
           - evdev:
@@ -34,9 +31,6 @@ mapping:
       target_event:
           gamepad:
               button: Guide
-    # QuickAccess triggers the Guide+South chord that gamescope recognizes
-    # as "open Quick Access Menu" (see write_event's QuickAccess special case
-    # in src/input/target/xpad.rs).
     - name: HOME (shortpress)
       source_events:
           - evdev:
@@ -55,18 +49,6 @@ mapping:
       target_event:
           gamepad:
               button: Keyboard
-    # The physical HOME button sends these chords on its own (not through the
-    # F16-F19 scheme above - see the vendor driver reasoning in
-    # devices/50-zotac-zone.yaml). Screenshot and Guide are the only two
-    # buttons in this file with real evdev output on the xbox-elite target
-    # (see event_codes_from_capability in src/input/event/evdev.rs and the
-    # special-cased write_event handling in src/input/target/xpad.rs) that
-    # also matches what these chords already did by accident: consuming the
-    # keys stops them from leaking through to the desktop session's own
-    # Meta+D ("show desktop") / Ctrl+Alt+KP. bindings, which is what produced
-    # a screenshot/Steam-menu-like effect before this mapping existed.
-    # Note: this trades away that desktop-mode "show desktop" shortcut, since
-    # capability_map translation applies regardless of session type.
     - name: HOME (shortpress, real)
       mapping_type:
           evdev: chord
@@ -82,7 +64,7 @@ mapping:
       target_event:
           gamepad:
               button: Screenshot
-    - name: HOME (longpress, real)
+    - name: HOME (longpress)
       mapping_type:
           evdev: chord
       source_events:
--- a/rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
+++ b/rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
@@ -55,6 +55,52 @@ mapping:
       target_event:
           gamepad:
               button: Keyboard
+    # The physical HOME button sends these chords on its own (not through the
+    # F16-F19 scheme above - see the vendor driver reasoning in
+    # devices/50-zotac-zone.yaml). Screenshot and Guide are the only two
+    # buttons in this file with real evdev output on the xbox-elite target
+    # (see event_codes_from_capability in src/input/event/evdev.rs and the
+    # special-cased write_event handling in src/input/target/xpad.rs) that
+    # also matches what these chords already did by accident: consuming the
+    # keys stops them from leaking through to the desktop session's own
+    # Meta+D ("show desktop") / Ctrl+Alt+KP. bindings, which is what produced
+    # a screenshot/Steam-menu-like effect before this mapping existed.
+    # Note: this trades away that desktop-mode "show desktop" shortcut, since
+    # capability_map translation applies regardless of session type.
+    - name: HOME (shortpress, real)
+      mapping_type:
+          evdev: chord
+      source_events:
+          - evdev:
+                event_type: KEY
+                event_code: KEY_LEFTMETA
+                value_type: button
+          - evdev:
+                event_type: KEY
+                event_code: KEY_D
+                value_type: button
+      target_event:
+          gamepad:
+              button: Screenshot
+    - name: HOME (longpress, real)
+      mapping_type:
+          evdev: chord
+      source_events:
+          - evdev:
+                event_type: KEY
+                event_code: KEY_LEFTCTRL
+                value_type: button
+          - evdev:
+                event_type: KEY
+                event_code: KEY_LEFTALT
+                value_type: button
+          - evdev:
+                event_type: KEY
+                event_code: KEY_KPDOT
+                value_type: button
+      target_event:
+          gamepad:
+              button: Guide
     - name: Left Paddle
       source_events:
           - evdev:
--- a/rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
+++ b/rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
@@ -22,6 +22,9 @@ mapping:
       target_event:
           gamepad:
               button: Guide
+    # Plain Guide press opens the main Steam menu. QuickAccess2 has no evdev
+    # output on the xbox-elite target (see event_codes_from_capability in
+    # src/input/event/evdev.rs), so binding this button to it was a no-op.
     - name: MORE
       source_events:
           - evdev:
@@ -30,7 +33,10 @@ mapping:
                 value_type: button
       target_event:
           gamepad:
-              button: QuickAccess
+              button: Guide
+    # QuickAccess triggers the Guide+South chord that gamescope recognizes
+    # as "open Quick Access Menu" (see write_event's QuickAccess special case
+    # in src/input/target/xpad.rs).
     - name: HOME (shortpress)
       source_events:
           - evdev:
@@ -39,7 +45,7 @@ mapping:
                 value_type: button
       target_event:
           gamepad:
-              button: QuickAccess2
+              button: QuickAccess
     - name: HOME (longpress)
       source_events:
           - evdev:
--- a/rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
+++ b/rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
@@ -73,17 +73,6 @@ mapping:
       target_event:
           gamepad:
               button: RightPaddle1
-    # The dials are deliberately not mapped here. They reach no evdev node at
-    # all - their HID usages are declared on the Generic Desktop page but hold
-    # Consumer page codes, so the kernel translates none of them - which means
-    # a REL_WHEEL/REL_HWHEEL mapping never fired for the dials it was written
-    # for.
-    #
-    # It claimed events all the same: capability maps are applied per composite
-    # device rather than per source, so the wheel events of the left touchpad -
-    # which is a scroll surface, not a pointer one - were turned into a dial
-    # capability that none of this device's targets consume, leaving that
-    # touchpad dead.
 
 # List of events to filter from the source devices
 filtered_events: []
--- a/rootfs/usr/share/inputplumber/devices/50-zotac-zone.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-zotac-zone.yaml
@@ -27,13 +27,31 @@ source_devices:
       vendor_id: 0x1ee9
       product_id: 0x1590
       interface_num: 3
+  # This name matches identically-named evdev nodes on interfaces 1.2 and 1.3
+  # in addition to the real gamepad node on interface 1.1, so phys_path pins
+  # the match to interface 1.1 only. Without it, InputPlumber creates three
+  # duplicate CompositeDevices (and three duplicate virtual gamepads) for one
+  # physical controller.
   - group: gamepad
     evdev:
       name: Zotac Technology Limited ZOTAC GAMING ZONE
       vendor_id: "1ee9"
       product_id: "1590"
+      phys_path: "*/input1"
       handler: event*
     capability_map_id: zone1
+  # Raw XInput-compatible joystick node exposed by the kernel's xpad driver
+  # (interface 1.0). Carries the real face buttons/sticks/triggers/Select
+  # (View)/Start/Guide. Grabbing it here hides it from Steam so those inputs
+  # go through the translated virtual gamepad instead of a second, untranslated
+  # raw controller.
+  - group: gamepad
+    evdev:
+      name: ZOTAC Gaming Zone
+      vendor_id: "1ee9"
+      product_id: "1590"
+      phys_path: "*/input0"
+      handler: event*
   - group: keyboard
     evdev:
       name: Zotac Technology Limited ZOTAC GAMING ZONE Keyboard
@@ -50,7 +68,7 @@ source_devices:
   #    handler: event*
   - group: mouse
     evdev:
-      name: ZOTAC Gaming Zone Dials
+      name: Zotac Technology Limited ZOTAC GAMING ZONE Mouse
       vendor_id: "1ee9"
       product_id: "1590"
       handler: event*
--- a/rootfs/usr/share/inputplumber/devices/50-zotac-zone.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-zotac-zone.yaml
@@ -29,14 +29,14 @@ source_devices:
       interface_num: 3
   - group: gamepad
     evdev:
-      name: ZOTAC Gaming Zone Gamepad
+      name: Zotac Technology Limited ZOTAC GAMING ZONE
       vendor_id: "1ee9"
       product_id: "1590"
       handler: event*
     capability_map_id: zone1
   - group: keyboard
     evdev:
-      name: ZOTAC Gaming Zone Keyboard
+      name: Zotac Technology Limited ZOTAC GAMING ZONE Keyboard
       vendor_id: "1ee9"
       product_id: "1590"
       handler: event*
--- a/rootfs/usr/share/inputplumber/devices/50-zotac-zone.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-zotac-zone.yaml
@@ -27,24 +27,14 @@ source_devices:
       vendor_id: 0x1ee9
       product_id: 0x1590
       interface_num: 3
-  # This name matches identically-named evdev nodes on interfaces 1.2 and 1.3
-  # in addition to the real gamepad node on interface 1.1, so phys_path pins
-  # the match to interface 1.1 only. Without it, InputPlumber creates three
-  # duplicate CompositeDevices (and three duplicate virtual gamepads) for one
-  # physical controller.
   - group: gamepad
     evdev:
-      name: Zotac Technology Limited ZOTAC GAMING ZONE
+      name: "{ZOTAC Gaming Zone Gamepad,Zotac Technology Limited ZOTAC GAMING ZONE}"
       vendor_id: "1ee9"
       product_id: "1590"
       phys_path: "*/input1"
       handler: event*
     capability_map_id: zone1
-  # Raw XInput-compatible joystick node exposed by the kernel's xpad driver
-  # (interface 1.0). Carries the real face buttons/sticks/triggers/Select
-  # (View)/Start/Guide. Grabbing it here hides it from Steam so those inputs
-  # go through the translated virtual gamepad instead of a second, untranslated
-  # raw controller.
   - group: gamepad
     evdev:
       name: ZOTAC Gaming Zone
@@ -54,7 +44,7 @@ source_devices:
       handler: event*
   - group: keyboard
     evdev:
-      name: Zotac Technology Limited ZOTAC GAMING ZONE Keyboard
+      name: "{ZOTAC Gaming Zone Keyboard,Zotac Technology Limited ZOTAC GAMING ZONE Keyboard}"
       vendor_id: "1ee9"
       product_id: "1590"
       handler: event*
@@ -68,11 +58,10 @@ source_devices:
   #    handler: event*
   - group: mouse
     evdev:
-      name: Zotac Technology Limited ZOTAC GAMING ZONE Mouse
+      name: "{ZOTAC Gaming Zone Mouse,ZOTAC Gaming Zone Dials,Zotac Technology Limited ZOTAC GAMING ZONE Mouse}"
       vendor_id: "1ee9"
       product_id: "1590"
       handler: event*
-    capability_map_id: zone1
   - group: imu
     iio:
       name: "{i2c-BMI0160:00,bmi260}"
--- a/rootfs/usr/share/inputplumber/capability_maps/zone_type1_dial.yaml
+++ b/rootfs/usr/share/inputplumber/capability_maps/zone_type1_dial.yaml
@@ -1,0 +1,45 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ShadowBlip/InputPlumber/main/rootfs/usr/share/inputplumber/schema/capability_map_v2.json
+# Schema version number
+version: 2
+
+# The type of configuration schema
+kind: CapabilityMap
+
+# Name for the device event map
+name: Zotac Zone Type 1 Dial
+
+# Unique identifier of the capability mapping
+id: zone1_dial
+
+# Not attached to any source_devices entry in 50-zotac-zone.yaml yet: the
+# dials only reach a hidraw report (interface 1, report ID 3) with no evdev
+# node at all, and no source driver reads it. Kept here, scoped to its own
+# id, for whichever evdev-producing source ends up carrying REL_WHEEL /
+# REL_HWHEEL for the dials (e.g. once a vendor kernel driver or a hidraw
+# parser is in place) so it can be attached without also catching the
+# unrelated left touchpad, which emits the same REL_WHEEL/REL_HWHEEL codes
+# on its own source and does not want them translated.
+mapping:
+    - name: Left Dial
+      source_events:
+          - evdev:
+                event_type: REL
+                event_code: REL_HWHEEL
+                value_type: joystick_x
+      target_event:
+          gamepad:
+              dial:
+                  name: LeftStickDial
+    - name: Right Dial
+      source_events:
+          - evdev:
+                event_code: REL_WHEEL
+                event_type: REL
+                value_type: joystick_x
+      target_event:
+          gamepad:
+              dial:
+                  name: RightStickDial
+
+# List of events to filter from the source devices
+filtered_events: []


rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
```diff
@@ -67,26 +73,17 @@ mapping:
...
-      source_events:
-          - evdev:
-                event_code: REL_WHEEL
-                event_type: REL
-                value_type: joystick_x
-      target_event:
-          gamepad:
-              dial:
-                  name: RightStickDial
+    # The dials are deliberately not mapped here. They reach no evdev node at
```
pastaq: Instead of removing these, make a new capability map v2 and attach it only to the scroll wheels. These are exposed with the kernel driver.

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
```diff
@@ -67,26 +73,17 @@ mapping:
...
+    # all - their HID usages are declared on the Generic Desktop page but hold
+    # Consumer page codes, so the kernel translates none of them - which means
```
pastaq: Also, remove this comment block.

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
```diff
@@ -39,7 +45,7 @@ mapping:
                 value_type: button
       target_event:
           gamepad:
-              button: QuickAccess2
```
pastaq, line 42: Don't change these. The Zotac button is already mapped to guide. ... is mapped to QAM and home is mapped to QAM2. QAM2 is used by OpenGamepadUI to open the quick bar in overlay mode, and in the default profile this is translated to screenshot.

rootfs/usr/share/inputplumber/devices/50-zotac-zone.yaml
```diff
@@ -27,16 +27,34 @@ source_devices:
       vendor_id: 0x1ee9
       product_id: 0x1590
       interface_num: 3
+  # This name matches identically-named evdev nodes on interfaces 1.2 and 1.3
+  # in addition to the real gamepad node on interface 1.1, so phys_path pins
```
pastaq: Remove excessively verbose LLM generated code comment

rootfs/usr/share/inputplumber/devices/50-zotac-zone.yaml
```diff
@@ -27,16 +27,34 @@ source_devices:
...
+  # the match to interface 1.1 only. Without it, InputPlumber creates three
+  # duplicate CompositeDevices (and three duplicate virtual gamepads) for one
+  # physical controller.
   - group: gamepad
     evdev:
-      name: ZOTAC Gaming Zone Gamepad
+      name: Zotac Technology Limited ZOTAC GAMING ZONE
```
pastaq: Add a new entry or glob match instead of overwriting this. Different firmware presents differently.

rootfs/usr/share/inputplumber/devices/50-zotac-zone.yaml
```diff
@@ -27,16 +27,34 @@ source_devices:
...
       vendor_id: "1ee9"
       product_id: "1590"
+      phys_path: "*/input1"
       handler: event*
     capability_map_id: zone1
+  # Raw XInput-compatible joystick node exposed by the kernel's xpad driver
+  # (interface 1.0). Carries the real face buttons/sticks/triggers/Select
```
pastaq: Remove comment

rootfs/usr/share/inputplumber/devices/50-zotac-zone.yaml
```diff
@@ -27,16 +27,34 @@ source_devices:
...
+    evdev:
+      name: ZOTAC Gaming Zone
+      vendor_id: "1ee9"
+      product_id: "1590"
+      phys_path: "*/input0"
+      handler: event*
   - group: keyboard
     evdev:
-      name: ZOTAC Gaming Zone Keyboard
+      name: Zotac Technology Limited ZOTAC GAMING ZONE Keyboard
```
pastaq: Same as above

rootfs/usr/share/inputplumber/devices/50-zotac-zone.yaml
```diff
@@ -50,7 +68,7 @@ source_devices:
   #    handler: event*
   - group: mouse
     evdev:
-      name: ZOTAC Gaming Zone Dials
+      name: Zotac Technology Limited ZOTAC GAMING ZONE Mouse
```
pastaq: Same as above

pastaq (changes requested):

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
whirlwind80: Thanks for the review! I want to push back on this specific point with hardware evidence — I actually started from the same assumption you're describing here, then re-verified on real hardware.
  
  I captured the raw HID reports (hidraw0, interface 1, report ID 2) from each physical button in isolation:

  - ZOTAC (Steam) button: 02 09 00 6c ... → byte[3] = 0x6C
  - MORE (QAM) button: 02 09 00 6d ... → byte[3] = 0x6D
  - HOME button (short press): 02 08 00 07 ... → Meta+D (byte[1]=0x08 = LGUI modifier, byte[3]=0x07 = 'd')
  - ZOTAC button held for 2–8s: no additional or different report at all

  0x6C and 0x6D are literally the standard USB HID usage IDs for F17 and F18. On hid-generic — which is what binds all three of this device's HID interfaces here, and presumably on every stock-kernel install — those
  raw bytes pass straight through as KEY_F17/KEY_F18 untranslated. F16 and F19 never fire from any button on this hardware in that state, including a long ZOTAC press.

  I also traced why the mapping you're describing (ZOTAC→F16, MORE→F17→QAM, HOME→F18/F19→QAM2) exists at all: it's the output of the out-of-tree zotac_zone_hid vendor driver (OpenZotacZone/ZotacZone-Drivers
  (https://github.com/OpenZotacZone/ZotacZone-Drivers)), which isn't loaded here (modinfo zotac_zone_hid finds nothing; all 3 HID interfaces bind hid-generic). Its process_keyboard_report() intercepts the raw report
  and, in its default QAM_MODE_KEYBOARD, rewrites byte[3] before generic HID key translation:

  case 0x6C: data[3] = HID_USAGE_F16; // ZOTAC's raw byte
  case 0x6D: data[3] = HID_USAGE_F17; // MORE's raw byte
  case 0x07: data[3] = HID_USAGE_F18; // HOME short's raw byte
  case 0x63: data[3] = HID_USAGE_F19; // HOME long's raw byte

  That's not a coincidence: zone_type1.yaml's original author is also the author of that vendor driver (same identity on both). The original mapping matches the vendor driver's remapped output exactly because it was
  written against a dev environment that had that driver loaded.
  
  So both mappings are "correct" — just for two different, mutually exclusive states of the same hardware, depending on whether that community out-of-tree module happens to be loaded. Since it isn't part of any
  distro kernel I'm aware of (Bazzite included), I'd like to keep the current mapping (F17→Guide, F18→QuickAccess) as the default, since it matches every stock-kernel install.

  I'd be open to adding driver-based detection as follow-up work so both states are supported at once (Evdev source config is missing the driver match field that Udev sources already have), but I don't have the
  vendor driver built to verify that path on real hardware, so I'd rather not add unverified code in this PR.

  (Not a native English speaker — used AI assistance to translate/polish this comment from my own notes and findings.)

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
whirlwind80: Done in 8d89ead

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
whirlwind80: Done in 8d89ead

rootfs/usr/share/inputplumber/devices/50-zotac-zone.yaml
whirlwind80: Done in 8d89ead

rootfs/usr/share/inputplumber/devices/50-zotac-zone.yaml
whirlwind80: Done in 8d89ead

rootfs/usr/share/inputplumber/devices/50-zotac-zone.yaml
whirlwind80: Done in 8d89ead

rootfs/usr/share/inputplumber/devices/50-zotac-zone.yaml
whirlwind80: Done in 8d89ead

rootfs/usr/share/inputplumber/devices/50-zotac-zone.yaml
whirlwind80: Done in 8d89ead

whirlwind80:
Pushed one more commit (61df761) that changes how the physical HOME button's real chords (Meta+D short press, Ctrl+Alt+KP. long press — separate from the F16-F19 scheme handled elsewhere in this file) get mapped.

The previous version mapped these to GamepadButton::QuickAccess2/Keyboard, assuming OpenGamepadUI would pick them up over DBus. That doesn't apply to this device's config — target_devices here is xbox-elite, mouse, keyboard only, no DBus/unified_gamepad target is attached, and both QuickAccess2 and Keyboard resolve to an empty evdev code list on xbox-elite (event_codes_from_capability in evdev.rs). So that mapping was effectively a no-op.

Switched the targets to GamepadButton::Screenshot and GamepadButton::Guide instead, since those are the only two buttons here with real, special-cased evdev output on xbox-elite (see xpad.rs::write_event). 
Verified on hardware (HID debugfs capture + evtest with the grab released) that the chords themselves are sent reliably.

This device doesn't have any other button better suited to a capture/screenshot function than HOME, so reusing its chords for Screenshot/Guide seemed like the most reasonable choice.

(Translated with AI assistance, as I'm not a native English speaker.)

pastaq:
> Pushed one more commit ([61df761](https://github.com/ShadowBlip/InputPlumber/commit/61df76183c9a1476c71f695e16afe409183c6f8b)) that changes how the physical HOME button's real chords (Meta+D short press, Ctrl+Alt+KP. long press — separate from the F16-F19 scheme handled elsewhere in this file) get mapped.
> 
> The previous version mapped these to GamepadButton::QuickAccess2/Keyboard, assuming OpenGamepadUI would pick them up over DBus. That doesn't apply to this device's config — target_devices here is xbox-elite, mouse, keyboard only, no DBus/unified_gamepad target is attached, and both QuickAccess2 and Keyboard resolve to an empty evdev code list on xbox-elite (event_codes_from_capability in evdev.rs). So that mapping was effectively a no-op.

That isn't how this works at all. Adding the chords produced with the vanilla kernel is fine, but the mappings should mirror what was already there. What your AI failed to account for was how intercept mode and the OpenGamepadUI InputPlumber Profile are applied when OpenGamepadUI is running. It is not a no-op and it will have a use in either case. This matches the behavior of nearly every other device.

The short of it, adding the other chords is fine and wont interfere, but ZOTAC needs to be guide, ... QAM, and Home QAM2. If there is an additional code for long press HOME, Keyboard would be the next logicall and consistent mapping.

> Switched the targets to GamepadButton::Screenshot and GamepadButton::Guide instead, since those are the only two buttons here with real, special-cased evdev output on xbox-elite (see xpad.rs::write_event). Verified on hardware (HID debugfs capture + evtest with the grab released) that the chords themselves are sent reliably.
> 
> This device doesn't have any other button better suited to a capture/screenshot function than HOME, so reusing its chords for Screenshot/Guide seemed like the most reasonable choice.
> 
> (Translated with AI assistance, as I'm not a native English speaker.)

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
pastaq: See my comment below.

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
```diff
@@ -22,6 +22,9 @@ mapping:
       target_event:
           gamepad:
               button: Guide
+    # Plain Guide press opens the main Steam menu. QuickAccess2 has no evdev
```
pastaq: remove

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
```diff
@@ -30,7 +33,10 @@ mapping:
                 value_type: button
       target_event:
           gamepad:
-              button: QuickAccess
+              button: Guide
+    # QuickAccess triggers the Guide+South chord that gamescope recognizes
```
pastaq: remove

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
```diff
@@ -30,7 +33,10 @@ mapping:
                 value_type: button
       target_event:
           gamepad:
-              button: QuickAccess
```
pastaq, line 33: No

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
```diff
@@ -49,44 +55,70 @@ mapping:
...
+                value_type: button
+          - evdev:
+                event_type: KEY
+                event_code: KEY_D
                 value_type: button
       target_event:
           gamepad:
-              button: LeftPaddle1
-    - name: Right Paddle
+              button: Screenshot
```
pastaq, line 66: QuickAccess2

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
```diff
@@ -49,44 +55,70 @@ mapping:
       target_event:
           gamepad:
               button: Keyboard
-    - name: Left Paddle
+    # The physical HOME button sends these chords on its own (not through the
```
pastaq: Remove

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
```diff
@@ -49,44 +55,70 @@ mapping:
...
+          - evdev:
+                event_type: KEY
+                event_code: KEY_D
                 value_type: button
       target_event:
           gamepad:
-              button: LeftPaddle1
-    - name: Right Paddle
+              button: Screenshot
+    - name: HOME (longpress, real)
```
pastaq: Remove `, real`

rootfs/usr/share/inputplumber/capability_maps/zone_type1.yaml
```diff
@@ -49,44 +55,70 @@ mapping:
...
+                value_type: button
+          - evdev:
+                event_type: KEY
+                event_code: KEY_KPDOT
                 value_type: button
       target_event:
           gamepad:
-              button: RightPaddle1
-    - name: Left Dial
+              button: Guide
```
pastaq, line 85: Keyboard

rootfs/usr/share/inputplumber/devices/50-zotac-zone.yaml
```diff
@@ -50,11 +58,10 @@ source_devices:
   #    handler: event*
   - group: mouse
     evdev:
-      name: ZOTAC Gaming Zone Dials
+      name: "{ZOTAC Gaming Zone Mouse,ZOTAC Gaming Zone Dials,Zotac Technology Limited ZOTAC GAMING ZONE Mouse}"
       vendor_id: "1ee9"
       product_id: "1590"
       handler: event*
```
pastaq, line 64:
```
  - group: mouse
    passthrough: true
    evdev:
      name: "{ZOTAC Gaming Zone Mouse,ZOTAC Gaming Zone Dials,Zotac Technology Limited ZOTAC GAMING ZONE Mouse}"
      vendor_id: "1ee9"
      product_id: "1590"
      handler: event*
    events:
      exclude:
        - "*"
      include:
        - "Mouse:Wheel"
    capability_map_id: zone1_dail
        
```

pastaq (changes requested):

whirlwind80:
Thanks for the review. First, I want to explain the motivation behind this PR — I'm an actual daily user of this device (Zotac Gaming Zone), and I fixed a few input issues I ran into myself after switching to Bazzite. Other users have hit the same problems (#655 — QAM/Guide button not working, dials not working, back buttons not working), so I wanted to share the fix; I'm not trying to push a specific implementation for its own sake.

On the HOME/QAM mappings, I reproduced the review feedback (F17→QuickAccess, HOME short→QuickAccess2, HOME long→Keyboard) on real hardware. This device's default Gaming Mode session is already gamescope-session-ogui-steam ("Steam Big Picture Plus", which bundles OpenGamepadUI QAM), so I tested in both Desktop Mode (opengamepadui --overlay-mode) and that Gaming Mode session.

Results:
 - HOME short → QuickAccess2: No reaction in either Desktop Mode or Gaming Mode. This matches event_codes_from_capability (evdev.rs) returning an empty evdev code list for QuickAccess2 on the xbox-elite target,  and there was no corresponding activity in the OpenGamepadUI logs at the time of the button press either.
 - MORE(F17) → QuickAccess: Also no reaction. By contrast, the ZOTAC button (unchanged Guide mapping) reliably opened Steam Gaming Mode's own Quick Access Menu on its own — a plain Guide press alone already seems to be enough to bring that up, which makes me wonder whether the QuickAccess Guide+South synthesized sequence is actually interfering with it.
 - HOME long → Keyboard: Holding for about 5 seconds brought up the system power menu, but that's far longer than this chord's actual long-press threshold, so it looks like an unrelated system-level behavior rather than something caused by this mapping.

I could be missing some configuration around intercept mode / the profile layer — though what I tested was just OpenGamepadUI actually running (opengamepadui --overlay-mode on desktop, and this device's default Gaming Mode session, gamescope-session-ogui-steam). Is simply having OpenGamepadUI running not enough — does intercept mode need to be separately enabled inside OpenGamepadUI itself? If so, could you point me to how, and I'll test again. But based on what I've verified on real hardware so far, the currently merged mapping (Screenshot/Guide) was the only one that actually did something on this device — and it has the added benefit of working at the gamescope/xbox-elite level regardless of whether OpenGamepadUI is running at all.

On the dial capability_map, I saw you'd already worked on this device's touchpad in #410 — it looks like you have real background with this hardware. Looking at that diff, though, it seems like at the time you were treating the group named ZOTAC Gaming Zone Dials as effectively the touchpad, whereas my original REL_WHEEL→LeftStickDial/REL_HWHEEL→RightStickDial mappings in zone_type1.yaml were written assuming that name meant an actual physical dial — there seems to be some history of confusion between what the kernel reports as the device name and what it actually is.

I've already fixed this in a separate commit included in this PR. The actual root cause wasn't "capability_map applies globally regardless of source" — it was that the touchpad (mouse) source_devices entry itself had capability_map_id: zone1 set, and zone1 contained the dial rule, so the touchpad's own dedicated translator mistranslated its own genuine scroll events into the dial capability (source_devices[].capability_map_id builds a translator scoped to just that one source — see input/source/evdev.rs/gamepad.rs). So I moved the dial mapping into its own file (zone_type1_dial.yaml, id zone1_dial) and removed capability_map_id: zone1 from the touchpad entry. zone1_dial isn't attached to any source yet, since dials don't reach evdev at all on this hardware (no vendor driver loaded).
 
Your v2 suggestion (the code on line 64) still attaches a new capability_map to a single group matching {...Mouse,...Dials,...GAMING ZONE Mouse}. On hardware without the vendor driver, that glob only ever resolves to the touchpad, so I'd expect that group's own translator to reintroduce exactly the bug I just described — mistranslating its own genuine wheel events. I read through zotac-zone-hid-core.c in OpenZotacZone/ZotacZone-Drivers: even with the vendor module loaded, the dial's wheel_input and the touchpad's mouse_input are separate evdev devices, but mouse_input's own report handler independently emits REL_WHEEL too (the touchpad's native scroll, unrelated to the dials) — so the same risk seems to remain unless the two are kept as fully separate source_devices entries with non-overlapping name matches. I haven't actually built/loaded the vendor driver to verify this on hardware yet — this is based on reading the source only. Have you tested this exact group configuration with the vendor driver loaded?
 
I've already pushed a commit applying the feedback that isn't in dispute — cleaning up the comments, and (already done in an earlier commit) using a glob/new entry instead of overwriting the evdev name matching.
 
That said, if these mappings stay contested and end up blocking the whole PR, I'm fine dropping these parts or withdrawing the PR entirely — the original goal was just to share a fix I made for myself, not to force through a specific mapping.

pastaq:
> Thanks for the review. First, I want to explain the motivation behind this PR — I'm an actual daily user of this device (Zotac Gaming Zone), and I fixed a few input issues I ran into myself after switching to Bazzite. Other users have hit the same problems (#655 — QAM/Guide button not working, dials not working, back buttons not working), so I wanted to share the fix; I'm not trying to push a specific implementation for its own sake.

I understand, and I appreciate your patience with this. The sudden influx of many AI PR's resulting from the poorly managed Bazzite 44 stable release has me on edge a bit. I apologize for making that your problem. I do have a device in hand but haven't had time to get around to fixing it up yet, so your effort has been quite valuable to me.

> On the HOME/QAM mappings, I reproduced the review feedback (F17→QuickAccess, HOME short→QuickAccess2, HOME long→Keyboard) on real hardware. This device's default Gaming Mode session is already gamescope-session-ogui-steam ("Steam Big Picture Plus", which bundles OpenGamepadUI QAM), so I tested in both Desktop Mode (opengamepadui --overlay-mode) and that Gaming Mode session.
> 
> Results:
> 
>     * HOME short → QuickAccess2: No reaction in either Desktop Mode or Gaming Mode. This matches event_codes_from_capability (evdev.rs) returning an empty evdev code list for QuickAccess2 on the xbox-elite target,  and there was no corresponding activity in the OpenGamepadUI logs at the time of the button press either.
> 
>     * MORE(F17) → QuickAccess: Also no reaction. By contrast, the ZOTAC button (unchanged Guide mapping) reliably opened Steam Gaming Mode's own Quick Access Menu on its own — a plain Guide press alone already seems to be enough to bring that up, which makes me wonder whether the QuickAccess Guide+South synthesized sequence is actually interfering with it.
> 
>     * HOME long → Keyboard: Holding for about 5 seconds brought up the system power menu, but that's far longer than this chord's actual long-press threshold, so it looks like an unrelated system-level behavior rather than something caused by this mapping.

That sounds like he left meta button was produced and powerbuttond saw it. 

I think we should strip back to basics and get the raw truth on some things. Guide should only ever open the left menu in steam, quick access should only ever open the right side, and quick access 2 should either be a screenshot or open the OpenGamepadUI quick bar. That is universal for every device, regardless of config. This all indicates to me that something is misconfigured in the device.

Can you run `inputplumber device 0 test` for testing the config, rather than rely on the fairy broken userspace stack above it. This should work independently of steam failing to see events, or what appears to be a mistake I made in a patch last week for button chords.

For the seeming random power menu thing, `sudo libinput debug-events` and `evtest` together should be abtto figure out the source of that behavior. You'll want to mask inputplumber to do that. That should also help identify if there are additional events populated by other event devices we're not aware of.

> I could be missing some configuration around intercept mode / the profile layer — though what I tested was just OpenGamepadUI actually running (opengamepadui --overlay-mode on desktop, and this device's default Gaming Mode session, gamescope-session-ogui-steam). Is simply having OpenGamepadUI running not enough — does intercept mode need to be separately enabled inside OpenGamepadUI itself? If so, could you point me to how, and I'll test again. But based on what I've verified on real hardware so far, the currently merged mapping (Screenshot/Guide) was the only one that actually did something on this device — and it has the added benefit of working at the gamescope/xbox-elite level regardless of whether OpenGamepadUI is running at all.
>
> On the dial capability_map, I saw you'd already worked on this device's touchpad in #410 — it looks like you have real background with this hardware. Looking at that diff, though, it seems like at the time you were treating the group named ZOTAC Gaming Zone Dials as effectively the touchpad, whereas my original REL_WHEEL→LeftStickDial/REL_HWHEEL→RightStickDial mappings in zone_type1.yaml were written assuming that name meant an actual physical dial — there seems to be some history of confusion between what the kernel reports as the device name and what it actually is.
> 
> I've already fixed this in a separate commit included in this PR. The actual root cause wasn't "capability_map applies globally regardless of source" — it was that the touchpad (mouse) source_devices entry itself had capability_map_id: zone1 set, and zone1 contained the dial rule, so the touchpad's own dedicated translator mistranslated its own genuine scroll events into the dial capability (source_devices[].capability_map_id builds a translator scoped to just that one source — see input/source/evdev.rs/gamepad.rs). So I moved the dial mapping into its own file (zone_type1_dial.yaml, id zone1_dial) and removed capability_map_id: zone1 from the touchpad entry. zone1_dial isn't attached to any source yet, since dials don't reach evdev at all on this hardware (no vendor driver loaded).

Ah, my mistake. I was under the impression that the dials and the touchpad emitted form the same evdev. If you have 0 events form the dials and there are events getting swallowed then your changes make a lot more sense. 

I don't think we need to grab the touchpad at all then, unless there's another button going through that evdev. Remove that device entirely unless you think there is a need for it. I'll revisit the scroll wheels when I get around to fixing up the driver for upstreaming.

> Your v2 suggestion (the code on line 64) still attaches a new capability_map to a single group matching {...Mouse,...Dials,...GAMING ZONE Mouse}. On hardware without the vendor driver, that glob only ever resolves to the touchpad, so I'd expect that group's own translator to reintroduce exactly the bug I just described — mistranslating its own genuine wheel events. I read through zotac-zone-hid-core.c in OpenZotacZone/ZotacZone-Drivers: even with the vendor module loaded, the dial's wheel_input and the touchpad's mouse_input are separate evdev devices, but mouse_input's own report handler independently emits REL_WHEEL too (the touchpad's native scroll, unrelated to the dials) — so the same risk seems to remain unless the two are kept as fully separate source_devices entries with non-overlapping name matches. I haven't actually built/loaded the vendor driver to verify this on hardware yet — this is based on reading the source only. Have you tested this exact group configuration with the vendor driver loaded?
> 
> I've already pushed a commit applying the feedback that isn't in dispute — cleaning up the comments, and (already done in an earlier commit) using a glob/new entry instead of overwriting the evdev name matching.
> 
> That said, if these mappings stay contested and end up blocking the whole PR, I'm fine dropping these parts or withdrawing the PR entirely — the original goal was just to share a fix I made for myself, not to force through a specific mapping.

I'd prefer to keep going, I want you to get the credit for the fix and effort involved. I think we'll find a solution.

whirlwind80:
Thanks for the detailed follow-up — ran all three checks on hardware.

**1) `inputplumber device 0 test`**

I went through the full button set — South, East, North, West (all four face buttons), all four D-pad directions, LB, RB, L3, R3, Select, and Start — and every one reacted correctly. The ZOTAC button (Guide) and HOME long-press (also Guide) reacted correctly too.

`Screenshot` and `QuickAccess` are the only ones that never appear as boxes at all, regardless of whether the mapping actually fires. Looking at the source: the TUI's "Buttons" panel is populated from `composite_device::get_capabilities()`, which is the union of each source's raw, declared evdev capabilities. `KeyboardEventDevice::get_capabilities()` (`source/evdev/keyboard.rs`) builds its list purely by walking `device.supported_events()`/`supported_keys()` through `EvdevEvent::as_capability()` — it never consults the translator (and `SourceDriver::get_capabilities()`'s event filter only subtracts, never adds). Since `Screenshot`/`QuickAccess` exist only as chord-translation output and are never a raw source capability, they can't show up in that panel whether the mapping works or not — so the tool can't confirm those two specifically.

That said, I do have strong indirect confirmation: the physical HOME **long-press** correctly lit up the `Guide` box in the same TUI — that's the *same* chord-translation pipeline the short-press mapping (`Screenshot`) goes through, just landing on a capability that happens to also be a raw one.

**2) The ~5s power-menu**

Confirmed this is unrelated to `KEY_LEFTMETA`/`powerbuttond`. With InputPlumber masked and stopped so the raw session sees everything unmodified, I captured the HOME long-press on the keyboard node with `evtest` and `libinput debug-events` running together.

The `evtest` capture (the one that actually identifies keys) shows only `KEY_LEFTCTRL`+`KEY_LEFTALT`+`KEY_KPDOT` for the entire press, with KPDOT auto-repeating — `KEY_LEFTMETA` never appears once. The `libinput debug-events` side masks keycodes as `***` as you'd expect, so it can't identify keys, but its count and timing agree: exactly three keys pressed and released over 7.905s.

I then isolated it by InputPlumber's presence:

- InputPlumber off (raw chord reaches the session directly) → power menu never appears, even holding well past 5s (checked twice).
- InputPlumber on, current mapping (`HOME long → Guide`) → power menu appears after a few seconds (observed once).

So the trigger looks like the virtual `Guide`/Xbox-Home button being held, not anything about the raw keys. This machine runs `steamos-manager`, which I confirmed is actively integrated with InputPlumber (its logs show it calling `ResetInterceptModes` over DBus during session switches, and it registers its own virtual input device). To be clear though, I could not capture direct log evidence pinning the power menu on `steamos-manager` — `journalctl -u steamos-manager` had no entries for that window. So this part is inference: it looks like the standard "hold Guide N seconds → power menu" gesture common to Deck-style handhelds, reacting to the virtual xbox-elite controller regardless of session type. Either way, it isn't the `HOME long → Guide` mapping misbehaving — it's that mapping working as intended.

**3) Touchpad `source_devices` entry**

Tested removing it entirely, in both desktop mode and Gaming Mode. With it removed (native/ungrabbed):

- Right pad pointer motion — works natively, no regression.
- Left pad scroll (vertical) — works natively, no regression.
- Gaming Mode: the existing buttons (Steam/QAM/ABXY/sticks etc.) — all fine, no regressions.

Regardless of grab state, this touchpad — at least on my unit's firmware (fw 1.3.9) — only ever sends `REL_WHEEL`. I confirmed via `evtest` twice (a deliberately horizontal stroke, and alternating left/right strokes): `REL_HWHEEL` never appeared. Wondering whether that was just a limitation of how `hid-generic` interprets the report, I went a level deeper and parsed the raw HID report descriptor via `HIDIOCGRDESC`. The mouse report (rid=4) declares exactly three Generic Desktop axes at the firmware level — `X`, `Y`, `Wheel` (size=8, count=3, i.e. exactly three bytes, so there's no room for a fourth axis) — and there's no horizontal-scroll usage ("AC Pan", Consumer `0x238`) anywhere in the descriptor. So this isn't a driver/OS-level limitation; the firmware simply never emits horizontal-scroll data.

In practice this means trying to scroll purely horizontally on the left pad still produces vertical scroll ticks, and those ticks even come out in a mixed up/down direction (visible in the raw capture) — my guess is the sensor only really tracks motion along the strip's own axis, so swiping across it just yields noise. Either way it's not something a config change can address, and a different firmware revision might behave differently.

Net result: removing the `source_devices` entry loses nothing real on this hardware, and I don't have a strong reason to keep it. I've verified that locally, but I'm holding off on actually committing the removal until you confirm you're happy with it, since you added it originally (#410) and know this touchpad's history best.

Let me know if you'd like anything re-tested, and thanks again for the thoroughness on this.

pastaq:
> Thanks for the detailed follow-up — ran all three checks on hardware.
> 
> **1) `inputplumber device 0 test`**
> 
> I went through the full button set — South, East, North, West (all four face buttons), all four D-pad directions, LB, RB, L3, R3, Select, and Start — and every one reacted correctly. The ZOTAC button (Guide) and HOME long-press (also Guide) reacted correctly too.
> 
> `Screenshot` and `QuickAccess` are the only ones that never appear as boxes at all, regardless of whether the mapping actually fires. Looking at the source: the TUI's "Buttons" panel is populated from `composite_device::get_capabilities()`, which is the union of each source's raw, declared evdev capabilities. `KeyboardEventDevice::get_capabilities()` (`source/evdev/keyboard.rs`) builds its list purely by walking `device.supported_events()`/`supported_keys()` through `EvdevEvent::as_capability()` — it never consults the translator (and `SourceDriver::get_capabilities()`'s event filter only subtracts, never adds). Since `Screenshot`/`QuickAccess` exist only as chord-translation output and are never a raw source capability, they can't show up in that panel whether the mapping works or not — so the tool can't confirm those two specifically.

All the capabilities should show up and be testable. This is how I validate configs daily, so if that isn't behaving as expected then there is something strange going on. All items in a capability map are added to the composite device's capabilities HashSet:
https://github.com/ShadowBlip/InputPlumber/blob/23f84b78521a4e12f2ff7b16fa9e0dc03ccb1846/src/input/composite_device/mod.rs#L233-L256

I am still adamant that the three button functions follow my recommendation. On every single device config we try to keep the same capability in approximately the same place so so that it can feel intuitive between devices. Guide is always on the left side of the screen when available or the left most button on the right when not, QuickAccess is primarily the left most button on the right side of the screen. If we have an additional unused button we always use QuickAccess2, which becomes either screenshot when OpenGamepadUI is not running, or opens the OGUI Quick Bar when it is. A fourth additional button or function will become Keyboard if available. If any of those buttons have dual modes then it usually makes sense to group QuickAccess and QuickAccess2 together due to the similar function and give keyboard its own action. 

For the gaming zone I'm not particularly partial to which action is which on the right side. Due to the placement and size of each button it would make the most sense for ... to be QuickAcces, home short to be QUickAccess2, and home long to be Keyboard. That would be in line with what all other devices do. The mapping with the driver and the mapping without the driver do need to functionally match 1:1. I expect you will get better results in the input tester if you match the config I'm suggesting before you do the testing.

Given the back and fourth on this I plan to pull your changes this weekend and validate your results on my Zotac Zone. I'll be testing the OGC kernel without the zotac driver and the valve kernel with it.

> That said, I do have strong indirect confirmation: the physical HOME **long-press** correctly lit up the `Guide` box in the same TUI — that's the _same_ chord-translation pipeline the short-press mapping (`Screenshot`) goes through, just landing on a capability that happens to also be a raw one.
> 
> **2) The ~5s power-menu**
> 
> Confirmed this is unrelated to `KEY_LEFTMETA`/`powerbuttond`. With InputPlumber masked and stopped so the raw session sees everything unmodified, I captured the HOME long-press on the keyboard node with `evtest` and `libinput debug-events` running together.
> 
> The `evtest` capture (the one that actually identifies keys) shows only `KEY_LEFTCTRL`+`KEY_LEFTALT`+`KEY_KPDOT` for the entire press, with KPDOT auto-repeating — `KEY_LEFTMETA` never appears once. The `libinput debug-events` side masks keycodes as `***` as you'd expect, so it can't identify keys, but its count and timing agree: exactly three keys pressed and released over 7.905s.
> 
> I then isolated it by InputPlumber's presence:
> 
>     * InputPlumber off (raw chord reaches the session directly) → power menu never appears, even holding well past 5s (checked twice).
> 
>     * InputPlumber on, current mapping (`HOME long → Guide`) → power menu appears after a few seconds (observed once).
> 
> 
> So the trigger looks like the virtual `Guide`/Xbox-Home button being held, not anything about the raw keys. This machine runs `steamos-manager`, which I confirmed is actively integrated with InputPlumber (its logs show it calling `ResetInterceptModes` over DBus during session switches, and it registers its own virtual input device). To be clear though, I could not capture direct log evidence pinning the power menu on `steamos-manager` — `journalctl -u steamos-manager` had no entries for that window. So this part is inference: it looks like the standard "hold Guide N seconds → power menu" gesture common to Deck-style handhelds, reacting to the virtual xbox-elite controller regardless of session type. Either way, it isn't the `HOME long → Guide` mapping misbehaving — it's that mapping working as intended.
> 
> **3) Touchpad `source_devices` entry**
> 
> Tested removing it entirely, in both desktop mode and Gaming Mode. With it removed (native/ungrabbed):
> 
>     * Right pad pointer motion — works natively, no regression.
> 
>     * Left pad scroll (vertical) — works natively, no regression.
> 
>     * Gaming Mode: the existing buttons (Steam/QAM/ABXY/sticks etc.) — all fine, no regressions.
> 
> 
> Regardless of grab state, this touchpad — at least on my unit's firmware (fw 1.3.9) — only ever sends `REL_WHEEL`. I confirmed via `evtest` twice (a deliberately horizontal stroke, and alternating left/right strokes): `REL_HWHEEL` never appeared. Wondering whether that was just a limitation of how `hid-generic` interprets the report, I went a level deeper and parsed the raw HID report descriptor via `HIDIOCGRDESC`. The mouse report (rid=4) declares exactly three Generic Desktop axes at the firmware level — `X`, `Y`, `Wheel` (size=8, count=3, i.e. exactly three bytes, so there's no room for a fourth axis) — and there's no horizontal-scroll usage ("AC Pan", Consumer `0x238`) anywhere in the descriptor. So this isn't a driver/OS-level limitation; the firmware simply never emits horizontal-scroll data.
> 
> In practice this means trying to scroll purely horizontally on the left pad still produces vertical scroll ticks, and those ticks even come out in a mixed up/down direction (visible in the raw capture) — my guess is the sensor only really tracks motion along the strip's own axis, so swiping across it just yields noise. Either way it's not something a config change can address, and a different firmware revision might behave differently.
> 
> Net result: removing the `source_devices` entry loses nothing real on this hardware, and I don't have a strong reason to keep it. I've verified that locally, but I'm holding off on actually committing the removal until you confirm you're happy with it, since you added it originally (#410) and know this touchpad's history best.
> 
> Let me know if you'd like anything re-tested, and thanks again for the thoroughness on this.

rootfs/usr/share/inputplumber/devices/50-zotac-zone.yaml
```diff
@@ -50,11 +58,10 @@ source_devices:
   #    handler: event*
   - group: mouse
```
pastaq, line 59: This matches on multiple devices now, add unique: false

rootfs/usr/share/inputplumber/devices/50-zotac-zone.yaml
```diff
@@ -29,14 +29,22 @@ source_devices:
       interface_num: 3
   - group: gamepad
     evdev:
-      name: ZOTAC Gaming Zone Gamepad
+      name: "{ZOTAC Gaming Zone Gamepad,Zotac Technology Limited ZOTAC GAMING ZONE}"
       vendor_id: "1ee9"
       product_id: "1590"
+      phys_path: "*/input1"
```
pastaq, line 35: This breaks the config when the driver is present

```
I: Bus=0003 Vendor=1ee9 Product=1590 Version=0100
N: Name="ZOTAC Gaming Zone Gamepad"
P: Phys=usb-0000:c4:00.3-4/input3
S: Sysfs=/devices/pci0000:00/0000:00:08.1/0000:c4:00.3/usb1/1-4/1-4:1.3/0003:1EE9:1590.0003/input/input17
U: Uniq=
H: Handlers=kbd event11 js0
B: PROP=0
B: EV=20000b
B: KEY=3f 0 0 0 0 0 0 7cdb000000000000 0 3f00000000000000 0 0
B: ABS=3003f
B: FF=107030000 0
```

use phys_path: "{*/input[1|3]}"

pastaq:
After extensive evaluation with the zotac_zone_hid driver blacklisted I can confidently state that this is not a viable configuration. The gamepad does not enumerate without the driver present at all without a udev rule for xpad.

PR #668, opened 2026-08-25
fix(Zotac Zone): bind rear paddles without the vendor kernel driver

whirlwind80:
## Problem

On the Zotac Gaming Zone the rear macro buttons (M1/M2) do nothing. They produce no input on any evdev node, and no HID report on any of the device's hidraw interfaces — pressing and holding one for five seconds yields nothing at all.

The cause is in the device rather than in the driver stack. Querying the button mapping table over the vendor config interface shows M1 and M2 with an entirely empty mapping, while ordinary buttons have one:

```
button M1 (0x01):  gamepad=00 00 00 00  mod=00  kbd=00 00 00 00 00 00  mouse=00
button M2 (0x02):  gamepad=00 00 00 00  mod=00  kbd=00 00 00 00 00 00  mouse=00
button A  (0x0F):  gamepad=00 10 00 00  ...
```

With nothing bound, the firmware sends nothing, so no amount of work on the host side can recover these buttons — the mapping has to be written to the device.

That is what the existing code intends to do:

```rust
set_attribute(&mut device, "btn_m2/remap/keyboard", "home");
set_attribute(&mut device, "btn_m1/remap/keyboard", "end");
```

but those sysfs attributes only exist when the out-of-tree `zotac_zone_hid` driver is loaded. On a stock kernel (Bazzite 44 here) that module is absent and all three of the device's HID interfaces bind to `hid-generic`, so the paddles stay permanently silent.

## Change

When the vendor driver isn't loaded, ask the device for the same binding directly, speaking its configuration protocol over hidraw: M2 → Home, M1 → End. The `zone1` capability map already translates those into `LeftPaddle1`/`RightPaddle1`, so no config changes are needed.

`CMD_SAVE_CONFIG` is deliberately not sent. The binding stays volatile and is re-applied on each start, leaving the user's stored device configuration untouched.

This also removes the early `Err` return when the vendor driver is absent. That previously made the whole hidraw source fail to be created, which left no way to configure the device at all — the source has to exist for the commands above to be sent.

## Provenance

The frame layout, CRC and command codes were derived from `driver/hid/zotac-zone-hid-config.c` in [OpenZotacZone/ZotacZone-Drivers](https://github.com/OpenZotacZone/ZotacZone-Drivers) (GPL-2.0-or-later, Copyright (c) 2025 Luke D. Jones), used here under that license's "or later" clause, and verified against live captures from the hardware.

## Testing

Hardware: Zotac Gaming Zone (G0A1W), firmware 1.3.9, Bazzite 44.20260820, no vendor kernel module present.

- Unit tests cover the frame construction and the checksum, the latter against a frame captured from the device (`e1 00 c5 3c b2 … fc 8d`).
- With the paddle mapping cleared beforehand, starting InputPlumber binds both buttons and the device acknowledges each command with a zero status:
  ```
  Zotac Zone config interface is bound to 'hid-generic' rather than the vendor driver; configuring it over hidraw instead.
  Mapped Zotac Zone button 0x02 to key 0x4a
  Mapped Zotac Zone button 0x01 to key 0x4d
  ```
- The paddles then register as paddle buttons in Steam Game Mode, with no regression in the existing buttons (Steam, QAM, View, sticks, face buttons, shoulders, triggers, D-pad).
- `make test` passes.

## Note

This is independent of #664 and applies to `main` on its own; the two touch different files.

## AI disclosure

Per CONTRIBUTING.md: the protocol reverse engineering (hidraw capture and report descriptor analysis), the `hid_report.rs` frame/CRC implementation and its unit tests, and the `driver.rs` sysfs/hidraw fallback were developed with Claude Opus 5 in an interactive session. Prompts were investigative rather than code-generating — the work proceeded by capturing HID traffic under controlled inputs, cross-checking the results against the vendor driver source, and validating the CRC against captured frames before any write to the device. Every line has been reviewed and can be explained.

### DIFF
--- a/src/drivers/zotac_zone/driver.rs
+++ b/src/drivers/zotac_zone/driver.rs
@@ -87,9 +87,8 @@ fn configure_via_sysfs(udevice: &UdevDevice) -> bool {
     true
 }
 
-/// Configure the device over hidraw. This is a one-shot, uncommitted
-/// configuration, so it doesn't override the user's saved device settings and
-/// lasts only until the device is power cycled.
+/// Configure the device over hidraw. Oneshot config prevents overriding user
+/// set config.
 fn configure_via_hidraw(udevice: &UdevDevice) -> Result<(), Box<dyn Error + Send + Sync>> {
     let path = CString::new(udevice.devnode())?;
     let api = hidapi::HidApi::new()?;
--- a/src/drivers/zotac_zone/driver.rs
+++ b/src/drivers/zotac_zone/driver.rs
@@ -1,8 +1,16 @@
 use std::error::Error;
+use std::ffi::CString;
+use std::time::Duration;
 
+use hidapi::HidDevice;
 use udev::Device;
 
-use crate::udev::device::{AttributeGetter, AttributeSetter, UdevDevice};
+use crate::udev::device::{AttributeSetter, UdevDevice};
+
+use super::hid_report::{
+    mapping_status, set_keyboard_mapping, BUTTON_M1, BUTTON_M2, HID_KEY_END, HID_KEY_HOME,
+    REPORT_SIZE,
+};
 
 // Hardware ID's
 const ZONE_PID: u16 = 0x1590;
@@ -11,6 +19,9 @@ pub const PIDS: [u16; 1] = [ZONE_PID];
 pub const VID: u16 = 0x1ee9;
 pub const VID_ALT: u16 = 0x1e19;
 pub const VIDS: [u16; 2] = [VID, VID_ALT];
+
+/// How long to wait for the device to acknowledge a configuration command.
+const COMMAND_TIMEOUT: Duration = Duration::from_millis(500);
 
 pub struct Driver {
     _device: UdevDevice,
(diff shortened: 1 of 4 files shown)


src/drivers/zotac_zone/driver.rs
```diff
@@ -24,31 +35,117 @@ impl Driver {
...
         }
 
         Ok(Self { _device: udevice })
     }
 }
 
+/// Bind the rear macro buttons to Home and End, which the `zone1` capability
+/// map translates into the paddle capabilities.
+///
+/// The buttons ship with an empty mapping, so the firmware sends nothing at all
```
pastaq: Remove second paragraph and replace it with:

// Attempts first to bind with kernel driver and falls back to raw HID commands.

src/drivers/zotac_zone/driver.rs
```diff
@@ -24,31 +35,117 @@ impl Driver {
...
+    }
+
+    set_attribute(&mut device, "qam_mode", "0");
+    set_attribute(&mut device, "btn_m2/remap/keyboard", "home");
+    set_attribute(&mut device, "btn_m1/remap/keyboard", "end");
+
+    true
+}
+
+/// Configure the device by speaking the vendor protocol over hidraw.
```
pastaq: // Configure the device over hidraw. Oneshot config prevents overriding user set config.

src/drivers/zotac_zone/hid_report.rs
```diff
@@ -0,0 +1,151 @@
...
+/// an answer to a button mapping command. A status of zero means the device
+/// accepted it.
+pub fn mapping_status(response: &[u8]) -> Option<u8> {
+    if response.get(COMMAND_POS) != Some(&CMD_SET_BUTTON_MAPPING) {
+        return None;
+    }
+    response.get(MAP_STATUS_POS).copied()
+}
+
+#[cfg(test)]
```
pastaq: create an hid_report_test.rs and keep all tests in that file. use super if needed.

src/drivers/zotac_zone/hid_report.rs
```diff
@@ -0,0 +1,151 @@
...
+mod tests {
+    use super::*;
```
pastaq: only import used members

src/drivers/zotac_zone/hid_report.rs
pastaq, line 1: While this filename matches the current convention, it doesn't actually implement a PackedStruct enumeration of an HID report.

src/drivers/zotac_zone/hid_report.rs
```diff
@@ -0,0 +1,151 @@
...
+        let h4 = h3 >> 4;
+
+        crc = (((((h3 << 1) ^ h4) << 4) ^ h2) << 3) as u16 ^ h4 as u16 ^ (crc >> 8);
+    }
+
+    crc
+}
+
+/// Build a command frame with the framing bytes filled in.
+fn build_frame(sequence: u8, command: u8) -> [u8; REPORT_SIZE] {
```
pastaq: PackedStruct is more readable and provides all of three of these functions without the excess constants. Positions are baked in.

You can use an enum for M1/M2, and add a sequence impl that converts from that enum to the appropriate sequence number. Also use an enum for command.

pastaq (changes requested):

src/drivers/zotac_zone/driver.rs
whirlwind80: Done in 2535562

src/drivers/zotac_zone/driver.rs
whirlwind80: Done in 2535562

src/drivers/zotac_zone/hid_report.rs
whirlwind80: Done in 2535562

src/drivers/zotac_zone/hid_report.rs
whirlwind80: Done in 2535562

src/drivers/zotac_zone/hid_report.rs
whirlwind80: Done in 2535562

src/drivers/zotac_zone/hid_report.rs
whirlwind80: Done in 2535562

whirlwind80:
Thanks for the review — pushed 2535562 addressing all of this:
 
 - Reworded the configure_macro_buttons doc comment to just state the kernel-driver-then-hidraw-fallback behavior.
 - Reworded the configure_via_hidraw doc comment down to a single line about the one-shot, non-overriding config.
 - Moved the unit tests into hid_report_test.rs, matching the convention already used by this repo's other drivers/*/hid_report.rs modules.
 - Dropped the use super::*; glob import there in favor of named imports.
 - Rewrote hid_report.rs around PackedStruct: Command and ButtonId are now enums (ButtonId also carries the per-button sequence number via a sequence() method), and ButtonMappingRequest/ButtonMappingResponse 
 replace the old offset-constants-and-array-poking functions.

 calc_crc itself is still a plain function, since PackedStruct doesn't compute checksums — kept it pub(crate) so the existing CRC test against a real captured device frame still works (that capture's command byte
 isn't a Command this module sends, so it can't be built through the type-safe request API).

 Re-verified on hardware: with the mapping cleared, this version still rebinds M2/M1 to Home/End with a zero status from the device, independently confirmed by re-reading the device's mapping table, and confirmed working in Steam Game Mode with no regression on the other buttons.

 (Translated with AI assistance, as I'm not a native English speaker.)

src/drivers/zotac_zone/driver.rs
whirlwind80: Actually tightened further in f231971 — matches your suggestion almost verbatim now.

src/drivers/zotac_zone/hid_report.rs
```diff
@@ -0,0 +1,141 @@
+//! Vendor configuration protocol spoken on the Zotac Zone's command interface.
+//!
+//! The frame layout, CRC and command codes were derived from the vendor kernel
+//! driver at <https://github.com/OpenZotacZone/ZotacZone-Drivers>
+//! (`driver/hid/zotac-zone-hid-config.c`, GPL-2.0-or-later, Copyright (c) 2025
+//! Luke D. Jones), which InputPlumber may use under that license's "or later"
+//! clause. Values were verified against live captures from a Zotac Zone
+//! running firmware 1.3.9.
+
+use packed_struct::prelude::*;
```
pastaq: Seems another one snuck in

whirlwind80: Fixed in 3389ed7

whirlwind80:
All feedback from both review passes should be addressed now (last one: 3389ed7, replacing the wildcard packed_struct import with named imports). CI is green and this should be ready for another look whenever you get a chance — thanks for the reviews.

pastaq (approved):

pastaq:
Reminder for me to squash merge this PR

pastaq:
We're going to ship the driver in OGC as this isn't a viable path forward.

PR #672, opened 2026-08-27
fix(Hardware Support): Add OneXPlayer X2 Mini Pro and 3

pastaq:
Adds OneXPlayer X2 Mini Pro config. This device matches the X1 devices for the most part, with the exception that it has two hidraw endpoints for the controller. THe right controller uses bInterfaceNumber 00, so remove matching on IID in the hid-oxp driver

Pulls in changes from #723 since they use the same driver and capability map.

### DIFF
--- a/rootfs/usr/share/inputplumber/capability_maps/onexplayer_type10.yaml
+++ b/rootfs/usr/share/inputplumber/capability_maps/onexplayer_type10.yaml
@@ -6,33 +6,26 @@ version: 1
 kind: CapabilityMap
 
 # Name for the device event map
-name: OneXPlayer Type 5
+name: OneXPlayer Type 10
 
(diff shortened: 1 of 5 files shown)


pastaq:
@Grippy98 can you test again and LMK if back paddles work with the HID driver in the kernel?

Grippy98:
Hey I've been out for some health issues, I'll try as soon as I get home but it might take me a few days. That's why I haven't been able to update on LKML either. Doing better though

pastaq:
> Hey I've been out for some health issues, I'll try as soon as I get home but it might take me a few days. That's why I haven't been able to update on LKML either. Doing better though

No worries, take care of yourself. There's no rush here. Glad to hear you're doing better.

PR #673, opened 2026-08-28
feat(Hardware Support): Add AYN Thor

justradical:
depends on #659 for the capabilitymap

in theory, a lot of ayn/retroid devices can use a single file, currently i have made separate ones for each but this can be refactored if thats preferred


pastaq:
Additional files are free and ensure we have flexibility if devices decide to do things like mount an IMU upside down. I'd prefer to keep them separate. TBS, since this depends on #659 it would make more sense to cherry-pick this commit into that PR, or target it directly as the merge target.

justradical:
closing this in favor of merging with #659

PR #675, opened 2026-08-28
feat(Hardware Support): Add AYN Thor and Odin 2 Mini

justradical:
Needs separate compatibility maps from #659 because of no back buttons (paddles) on these models.

The differences from the `rsinput1` map is that the Thor has 1 extra button that can be used for QAM2, and both devices have no back buttons:
- `ayn2` has the AYN key and no back buttons
- `ayn3` has no AYN key and no back buttons

### DIFF
--- a/rootfs/usr/share/inputplumber/devices/50-ayn_thor.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-ayn_thor.yaml
@@ -1,0 +1,34 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ShadowBlip/InputPlumber/main/rootfs/usr/share/inputplumber/schema/composite_device_v1.json
+version: 1
+kind: CompositeDevice
+name: AYN Thor
+
+options:
+  auto_manage: true
+
+matches:
+  - udev:
+      sys_path: /sys/firmware/devicetree/base
+      attributes:
+        - name: compatible
+          value: ayn,thor
+
+source_devices:
+  - group: gamepad
+    capability_map_id: ayn2
+    evdev:
+      name: AYN Odin2 Gamepad
+      phys_path: rsinput-gamepad/input0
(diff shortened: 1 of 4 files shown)


rootfs/usr/share/inputplumber/devices/50-ayn_thor.yaml
```diff
@@ -0,0 +1,52 @@
...
+  - group: gamepad
+    evdev:
+      name: AYN Odin2 Gamepad
+      phys_path: rsinput-gamepad/input0
+      handler: event*
+      vendor_id: "2020"
+      product_id: "3001"
+    capability_map_id: ayn2
+
+  - group: gamepad
```
pastaq, line 26: Can we add notes on if the gpio-keys are split or have volume keys on all the configs with gpio-keys?

Also applies to patches in #659 #676 #673

pastaq (changes requested):

rootfs/usr/share/inputplumber/devices/50-ayn_thor.yaml
justradical: This only really applies in #659, none of these devices using the `ayn2`, `ret1`, or `ret2` profiles have this issue

jenneron: added comments in #659, see https://github.com/ShadowBlip/InputPlumber/compare/2322f5fe07b6d9a1a6d352d3c91ed769d0dfc25c..03ee51d4cbb9749135c253da3e324811b2a459c8

justradical:
AYN key is now bound to QAM 2, i checked with a user and it has no extra behavior when held, and the profiles have been separated to `ayn2` and `ayn3`

pastaq (approved):

pastaq:

example:

PR #676, opened 2026-08-28
feat(Hardware Support): Add RP 5, Mini, Mini V2 and Flip 2

justradical:
Adds `ret1` capability map for the RP5, RPMini and RPMiniV2, and `ret2` for the Flip 2, which doesnt have the android back button for QAM


pastaq (approved):

pastaq (changes requested):

example:

pastaq (approved):
I just opened up issue #690 that makes my request impossible with the v2 maps, so we'll defer for now.

PR #677, opened 2026-08-29
fix(Legion Go 1): fix gyro not working on latest controller firmware, enable gyro right handle IMU

tctinh:
## Summary

Newer Legion Go 1 controller firmware exposes the controllers as hardware generation 2 (`17ef:61eb`), routing their input through the Go 2 driver. On the Legion Go 1 (`83E1`), the handle IMU was disabled or filtered make gyro not working at all, this PR re enable gyro on right controller which will work for both handheld and detached controller mode (Note it is possible to implement the gyro to use middle body gyro, I prefer the right controller)

This change:

- Uses the current `hid-lenovo-go` sysfs attributes.
- Enables IMU and IMU bypass for both handles.
- Keeps the right-handle accelerometer and gyroscope on Legion Go 1 while filtering the left and center controller sensors.
- Excludes the center IIO gyroscope to avoid competing motion streams.

## Testing

Tested on a Legion Go 1 (`83E1`) running SteamOS 3.8.26 with controller PID `17ef:61eb`.

Verified that accelerometer and gyroscope input from the right controller work while detached.


pastaq:
Thanks for trying to fix this. I think this approach is too opinionated towards your personal use case. I'm currently working on a full refactor of the way gyro and accelerometer are managed that is a bit more flexible.

PR #694, opened 2026-09-04
fix(config): add Mayflash GameCube Controller Adapter Support

spenoff:
This adapter's HID report descriptor declares its top-level collection as Usage(Joystick) [0x04] rather than Usage(Gamepad) [0x05], with its 16 face/shoulder buttons reported as a plain bitmask under the generic Button usage page. Because of this, the Linux kernel assigns the legacy joystick-era evdev codes BTN_TRIGGER (0x120) through BTN_DEAD (0x12f) to these buttons instead of the modern gamepad codes (BTN_SOUTH/0x130 etc.) that InputPlumber's generic evdev translation recognizes, so none of this adapter's face/shoulder buttons produced any input at all.

Adds:
 - A capability map + device profile for this adapter, with the bit-to- button layout empirically confirmed on real hardware (adapter in PC/ DInput mode) by pressing each button in complete isolation and watching the translated output - not inferred from the original GameCube console protocol, which does not match this adapter's actual bit order.
 - Three new EventCode variants (evdev codes 0x12c-0x12e) which have no symbolic name in the Linux kernel headers but are legitimately reported by this class of hardware.
 - A test validating both new config files parse correctly and that capability_map_id is set on the source_devices entry itself, not just at the CompositeDeviceConfig level - only the former actually wires up per-event translation at runtime.

Co-developed-by: Claude Sonnet 5

AI disclosure: this fix was developed interactively with Claude Code (Sonnet 5) over an extended debugging session. The root-cause diagnosis (HID Joystick vs Gamepad usage causing legacy evdev codes) was Claude's own investigation of the InputPlumber source. The bit-to-button mapping in the capability map was NOT AI-inferred - it was determined empirically by the human author pressing each physical button on real hardware in isolation while Claude captured and reported the resulting translated evdev output; several early AI-generated assumptions about the bit layout (based on the original GameCube console protocol) turned out to be wrong and were corrected only after the human author pushed back and asked for further isolated testing. The three new EventCode enum variants and the accompanying test were written by Claude and reviewed by the human author.

### DIFF
--- a/rootfs/usr/share/inputplumber/capability_maps/mayflash_gamecube_adapter.yaml
+++ b/rootfs/usr/share/inputplumber/capability_maps/mayflash_gamecube_adapter.yaml
@@ -1,0 +1,182 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ShadowBlip/InputPlumber/main/rootfs/usr/share/inputplumber/schema/capability_map_v2.json
+# Schema version number
+version: 2
+
+# The type of configuration schema
+kind: CapabilityMap
+
+# Name for the device event map
+name: Mayflash GameCube Controller Adapter
+
+# Unique identifier of the capability mapping
+id: mayflash_gamecube_adapter
+
+# Root cause: this adapter's HID report descriptor declares its top-level
+# collection as Usage(Joystick) [0x04] rather than Usage(Gamepad) [0x05], with
+# its 16 face/shoulder buttons reported under the generic Button usage page
+# (0x09, usages 1-16) as a single 16-bit bitmask - not under a labeled gamepad
+# usage. Because of this, Linux's kernel HID input core (drivers/hid/hid-input.c)
+# assigns the legacy joystick-era evdev codes BTN_TRIGGER (0x120) through
+# BTN_DEAD (0x12f) sequentially to these 16 buttons, instead of the modern
+# gamepad codes BTN_SOUTH (0x130) etc. that InputPlumber's generic evdev
+# translation (src/input/event/evdev.rs) recognizes. Every one of these codes
+# currently falls through to the catch-all `_ => Capability::NotImplemented`
+# arm, so none of this adapter's face/shoulder buttons produce any usable
+# input via the generic gamepad profile.
+#
+# Bit-to-button mapping below is EMPIRICALLY CONFIRMED against real hardware
+# (adapter set to PC/DInput mode - it also has separate Wii U and Pokken
+# Tournament modes, selected via button combos, which likely use a
+# completely different report layout; this mapping is PC-mode-specific), one
+# physical button pressed in complete isolation at a time while watching the
+# translated xb360 target device's raw evdev output.
+#
+# An earlier version of this file assumed the raw bit order followed the
+# original Nintendo console-to-controller protocol layout naively (A/B/X/Y
+# in that literal order at bits 0-3, Z/R-click/L-click at bits 12-14, and
+# Start unambiguous at bit 4). That assumption was WRONG on multiple counts,
+# each caught by isolated single-button testing after initial batch tests
+# gave misleading results:
+#   - A/B/X/Y are NOT in that order on the raw bitmask (confirmed by
+#     isolated presses of each one individually - this actually matches an
+#     initially-doubted community SDL gamecontrollerdb.txt entry for this
+#     same adapter, which turned out to be correct after all).
+#   - Z/R-click/L-click are NOT at bits 12-14 at all - those bits never
+#     fire no matter what is pressed.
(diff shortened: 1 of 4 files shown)


rootfs/usr/share/inputplumber/capability_maps/mayflash_gamecube_adapter.yaml
```diff
@@ -0,0 +1,182 @@
...
+# The type of configuration schema
+kind: CapabilityMap
+
+# Name for the device event map
+name: Mayflash GameCube Controller Adapter
+
+# Unique identifier of the capability mapping
+id: mayflash_gamecube_adapter
+
+# Root cause: this adapter's HID report descriptor declares its top-level
```
pastaq, line 14: This comment is over 60 lines long. Nobody is going to read that. Based on the first sentence it is probably not even necessary. Delete it.

rootfs/usr/share/inputplumber/capability_maps/mayflash_gamecube_adapter.yaml
```diff
@@ -0,0 +1,182 @@
...
+# GameCube button -> InputPlumber's South/East/North/West follows the usual
+# cross-platform convention of matching by *function* (South = primary
+# confirm, East = secondary/cancel, etc.), not by exact physical diamond
+# position, since GameCube's face buttons aren't laid out symmetrically like
+# an Xbox pad. Z has no clean Xbox-pad equivalent, so it's mapped to Select
+# (unclaimed on this profile's "xb360" target model and present on every
+# xb360 model, unlike RightPaddle1/2 which are xbox-elite-only and were
+# confirmed to silently produce no output when targeted here).
+mapping:
+  # Physical GameCube A (confirmed via isolated press) is raw bit 1.
```
pastaq, line 89: All of these  button comments describe what we can already read in the config. Delete them.

rootfs/usr/share/inputplumber/devices/55-mayflash_gamecube_adapter.yaml
```diff
@@ -0,0 +1,69 @@
...
+# Schema version number
+version: 1
+
+# The type of configuration schema
+kind: CompositeDevice
+
+# Name of the composite device mapping
+name: Mayflash GameCube Controller Adapter
+
+# This adapter's HID report descriptor declares its top-level collection as
```
pastaq, line 11: More AI slop comments. Delete them.

src/config/config_test.rs
```diff
@@ -145,3 +145,50 @@ async fn check_autostart_rules() -> Result<(), Box<dyn Error>> {
 
     Ok(())
 }
+
+/// Quick validation test for the new Mayflash GameCube Controller Adapter
```
pastaq, line 149: This doesn't need its own test

src/config/capability_map/evdev.rs
```diff
@@ -362,6 +362,17 @@ pub enum EventCode {
     BtnBase4,
     BtnBase5,
     BtnBase6,
+    /// evdev code 0x12c. No symbolic name exists in the Linux kernel headers
```
pastaq, line 365: More unnecessary comments.

src/config/capability_map/evdev.rs
```diff
@@ -362,6 +362,17 @@ pub enum EventCode {
...
+    /// (linux/input-event-codes.h jumps from BTN_BASE6/0x12b straight to
+    /// BTN_DEAD/0x12f), but the code itself is a valid, distinct evdev key
+    /// code that some hardware (e.g. HID joysticks whose report descriptor
+    /// declares a generic 16+ button Button usage page under a Joystick
+    /// rather than Gamepad top-level collection) legitimately reports.
+    BtnExtraOne,
```
pastaq, line 371: This has no translation to a NativeEvent. It needs a full path to becoming a Capability or it isn't valid as an event. 

It also starts a new convention instead of continuing an existing one, why note ButtonBase7-9?

src/config/capability_map/evdev.rs
```diff
@@ -362,6 +362,17 @@ pub enum EventCode {
     BtnBase4,
     BtnBase5,
     BtnBase6,
+    /// evdev code 0x12c. No symbolic name exists in the Linux kernel headers
+    /// (linux/input-event-codes.h jumps from BTN_BASE6/0x12b straight to
```
pastaq, line 366: This should be fixed in the kernel.

pastaq (changes requested):
Please remove all the AI slop comments. They don't add anything useful and nobody is going to read them. 

Your specific config didn't need a test, the generic one works fine. 

Complete the Capability for the new events so they have a capability they can translate from/to.

PR #670, opened 2026-08-26
feat(input): add KONKR Pocket FIT controller support

mydanyi:
## Summary

I have tested two KONKR Pocket FIT variants. They share the same enclosure and control layout, but use different Qualcomm SoCs and completely different input pipelines. InputPlumber selects the correct profile through each device tree's `compatible` value.

| Device | Platform | Device-tree compatible | Input pipeline |
| --- | --- | --- | --- |
| KONKR Pocket FIT | Qualcomm SM8650 / Snapdragon 8 Gen 3 | `konkr,pocket-fit` | USB AYANEO DirectInput/XInput controller, AYANEO auxiliary input, and hidraw rumble |
| KONKR Pocket FIT Elite | Qualcomm SM8750 / Snapdragon 8 Elite | `konkr,pocket-fit-elite` | SPI `AYANEO MCU Gamepad` and UART `KONKR System Buttons` evdev devices |

This PR adds:

- separate device profiles for both variants
- `konkr2` for the SM8650 USB controller and auxiliary input
- `konkr3` for the SM8750 system-button input
- direct `LeftPaddle1` / `RightPaddle1` capabilities for LC / RC on both variants
- an Xbox Elite default target for both variants
- a native hidraw rumble driver for the SM8650 AYANEO `4001:0428` DirectInput controller

The SM8750 Elite profile does not use the SM8650 USB or hidraw implementation. Its main controller already reports standard evdev events and native `FF_RUMBLE`, so InputPlumber's built-in evdev translation handles that source directly.

## KONKR Pocket FIT (SM8650) button mapping

| Physical control | Source event | InputPlumber output |
| --- | --- | --- |
| A / B / X / Y | `BTN_SOUTH` / `BTN_EAST` / `BTN_WEST` / `BTN_NORTH` | South / East / West / North |
| Start / Select | `BTN_START` / `BTN_SELECT` | Start / Select |
| L1 / R1 | `BTN_TL` / `BTN_TR` | Left / Right bumper |
| L2 / R2 | `ABS_BRAKE` / `ABS_GAS` | Left / Right trigger |
| Left and right sticks | `ABS_X/Y` and `ABS_Z/RZ` | Left and right sticks |
| L3 / R3 | `BTN_THUMBL` / `BTN_THUMBR` | Left and right stick click |
| D-pad | `ABS_HAT0X/Y` | D-pad |
| Navigation / Steam / Xbox Guide button | `BTN_MODE` | Guide |
[...]

PR #700, opened 2026-09-08
feat(Hardware Support): Add AYANEO Pocket DS

justradical:
Based on #627 and #670 

Adds the AYANEO Pocket DS as a supported device, tested all buttons on the device in both Xbox and AYANEO/dinput mode. Needs 2 capability maps due to the remapping of guide/start so that we could use the rocker for extra paddle keys, also avoids dead keys on the controller.

### DIFF
--- a/rootfs/usr/share/inputplumber/capability_maps/ayaneo_type10.yaml
+++ b/rootfs/usr/share/inputplumber/capability_maps/ayaneo_type10.yaml
@@ -1,0 +1,122 @@
+version: 2
+kind: CapabilityMap
+name: AYANEO Type 10
+id: aya10
+
+mapping:
+  - name: LC
+    source_events:
+      - evdev:
+            event_type: KEY
+            event_code: BTN3
+            value_type: button
+    target_event:
+      gamepad:
+        button: LeftPaddle1
+
+  - name: RC
+    source_events:
+      - evdev:
+            event_type: KEY
+            event_code: BTN4
+            value_type: button
+    target_event:
+      gamepad:
+        button: RightPaddle1
+
+  - name: Start Button to Guide
+    source_events:
+      - evdev:
+          event_type: KEY
+          event_code: BTN_START
+          value_type: button
+    target_event:
+      gamepad:
+        button: Guide
+
+  - name: Share looking icon to LeftPaddle2
+    source_events:
+      - evdev:
+          event_type: KEY
+          event_code: BTN2
+          value_type: button
+    target_event:
+      gamepad:
+        button: LeftPaddle2
+
+  - name: 3 dots to RightPaddle2
+    source_events:
+      - evdev:
+          event_type: KEY
+          event_code: BTN_MODE
+          value_type: button
+    target_event:
+      gamepad:
+        button: RightPaddle2
+
+  - name: Evil secondary start to QAM2
+    source_events:
+      - evdev:
+            event_type: KEY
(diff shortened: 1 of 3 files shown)


rootfs/usr/share/inputplumber/capability_maps/ayaneo_type10.yaml
```diff
@@ -0,0 +1,271 @@
...
+          value_type: button
+    target_event:
+      gamepad:
+        button: Start
+
+  - name: South Button
+    source_events:
+      - evdev:
+          event_type: KEY
+          event_code: BTN_SOUTH
```
pastaq: This will automatically convert, no need to add it to the map. Same for west/east/north/select/thumb_l/thumb_r, etc. If the evdev code matched the default capability you can omit.

pastaq (changes requested):

rootfs/usr/share/inputplumber/capability_maps/ayaneo_type10.yaml
justradical: I've removed everything that i could without the input map breaking

rootfs/usr/share/inputplumber/capability_maps/ayaneo_type10.yaml
```diff
@@ -0,0 +1,122 @@
...
+    source_events:
+      - evdev:
+          event_type: KEY
+          event_code: BTN_MODE
+          value_type: button
+    target_event:
+      gamepad:
+        button: RightPaddle2
+
+  - name: Evil secondary start to QAM2
```
pastaq, line 57: Lol

rootfs/usr/share/inputplumber/capability_maps/ayaneo_type11.yaml
pastaq, line 1: Both of these mappings are on the other map, is this file necessary?

pastaq (commented):
One question before I approve, I think we can drop the second config.

justradical:
> One question before I approve, I think we can drop the second config.

oh why do you think we can drop it?

Edit: nevermind, missed the review comment

rootfs/usr/share/inputplumber/capability_maps/ayaneo_type10.yaml
justradical: oh i forgot about those, i dont even know what the glyphs are supposed to be

rootfs/usr/share/inputplumber/capability_maps/ayaneo_type11.yaml
justradical: Wont dropping this make the joysticks and triggers act up since they are mapped in the dinput mode?

pastaq (approved):

rootfs/usr/share/inputplumber/capability_maps/ayaneo_type11.yaml
pastaq: Valid

rootfs/usr/share/inputplumber/capability_maps/ayaneo_type10.yaml
justradical: Do you want me to fix this? 😂

pastaq: Na

PR #701, opened 2026-09-08
Fix AYANEO NEXT 2 input device grouping and gyro

GhostMadneS:
Tested on the AYANEO NEXT 2 (AB09).

Changes:
- Fix duplicate composite devices by using `unique: false`.
- Match the AYANEO keyboard/special-button interface specifically via `input1`, leaving the native mouse interface untouched.
- Keep DInput/AYANEO desktop mode unmanaged so the firmware's native mouse/keyboard behavior continues to work.
- Fix the BMI323 mount matrix orientation.
- Use the DualSense Edge target so the BMI323 gyro is exposed correctly to Steam.

Tested in XInput mode:
- One composite device
- Gamepad controls working
- Back/special buttons working
- Native mouse/touchpad working
- Gyro working in InputPlumber and Steam

The RGB section from the existing profile was retained but has not yet been tested.

### DIFF
--- a/rootfs/usr/share/inputplumber/devices/50-ayaneo_next2.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-ayaneo_next2.yaml
@@ -26,6 +26,19 @@ source_devices:
       name: AYANEO COMPOSITE DEVICE
       phys_path: usb-0000:c6:00.0-1/input1
       handler: event*
+      events:
+        exclude:
+          - "*"
+        include:
+          - Keyboard:KeyF16
+          - Keyboard:KeyF17
+          - Keyboard:KeyF18
+          - Keyboard:KeyF19
+          - Keyboard:KeyF20
+          - Keyboard:KeyF21
+          - Keyboard:KeyF22
+          - Keyboard:KeyF23
+          - Keyboard:KeyLeftMeta
 
   # IMU
   - group: imu
@@ -46,7 +59,7 @@ options:
   auto_manage: true
 
 target_devices:
-  - ds5-edge
+  - xbox-elite
   - mouse
   - keyboard
 
--- a/rootfs/usr/share/inputplumber/devices/50-ayaneo_next2.yaml
+++ b/rootfs/usr/share/inputplumber/devices/50-ayaneo_next2.yaml
@@ -11,42 +11,29 @@ matches:
       sys_vendor: AYA
 
 source_devices:
-  # Xinput mode
+  # XInput mode
   - group: gamepad
+    unique: false
     evdev:
       name: Microsoft X-Box 360 pad
       phys_path: usb-0000:c6:00.0-4/input0
       handler: event*
-  # Dinput mode
-  - group: gamepad
-    evdev:
-      name: Microsoft X-Box 360 pad
-      phys_path: usb-0000:c6:00.0-4/input0
-      handler: event*
-    capability_map_id: dinput_generic
-  - group: gamepad
+
+  # AYANEO special/back buttons
+  - group: keyboard
+    unique: false
     evdev:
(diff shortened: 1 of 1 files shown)


rootfs/usr/share/inputplumber/devices/50-ayaneo_next2.yaml
```diff
@@ -11,42 +11,29 @@ matches:
...
+    unique: false
     evdev:
       name: AYANEO COMPOSITE DEVICE
+      phys_path: usb-0000:c6:00.0-1/input1
       handler: event*
-    events:
-      exclude:
-        - "*"
-      include:
-        - Keyboard:KeyF16
```
pastaq, line 35: We really want these present or 100's of capabilities that aren't valid will show up. in the input tester

rootfs/usr/share/inputplumber/devices/50-ayaneo_next2.yaml
```diff
@@ -59,7 +46,7 @@ options:
   auto_manage: true
 
 target_devices:
-  - xbox-elite
+  - ds5-edge
```
pastaq: Please leave this as xbox-elite. The preference of most users is that button glyphs match. Plenty of tools can reconfigure the device at runtime and advanced users can make their own override if they desire, but we want the default experience to be what is best for the majority of users.

rootfs/usr/share/inputplumber/devices/50-ayaneo_next2.yaml
```diff
@@ -11,42 +11,29 @@ matches:
...
 source_devices:
-  # Xinput mode
+  # XInput mode
   - group: gamepad
+    unique: false
     evdev:
       name: Microsoft X-Box 360 pad
       phys_path: usb-0000:c6:00.0-4/input0
       handler: event*
-  # Dinput mode
```
pastaq, line 20: No support for dinput mode?

pastaq (changes requested):

rootfs/usr/share/inputplumber/devices/50-ayaneo_next2.yaml
GhostMadneS: This is due to edge the only controller reads the gyro in steam mode so its visible in the controller options and working in steam controller or xbox elite its not there im not sure if there are different way to make steam read them or work amd im not sure gyro is important at all maybe to some...

rootfs/usr/share/inputplumber/devices/50-ayaneo_next2.yaml
GhostMadneS: I don't thino its needed since it breaks everything.. Leaving it like this makes the controller behave as Ayaneo inteded it to be which works similar to windows left stick is mouse movement control and A is left click X is right click, Y is enter,  B is ESC and so on basically mouse and keybored mode and it works interducing aything else will break it and break touchpads... So this is fine and works i  desktop amd in steam mode.

rootfs/usr/share/inputplumber/devices/50-ayaneo_next2.yaml
GhostMadneS: Ill try to add them to my test environment and see how they react and if i lose some fuctions but i understand what you mean i saw it..

rootfs/usr/share/inputplumber/devices/50-ayaneo_next2.yaml
pastaq: I know that gyro won't work in this mode. We do this for every device because we get a lot more complaints otherwise.

GhostMadneS:
RGB is not working wnd li ux doesnt expose anything intractable

pastaq (approved):

PR #710, opened 2026-09-12
fix(udev): restore device node permissions on unhide and re-trigger udev add

### LINKED ISSUE #686: Stopping inputplumber.service leaves grabbed evdev nodes at mode 000 (system left without a pointer)
## Summary

Stopping `inputplumber.service` leaves the evdev nodes it had grabbed at mode `000`, so the machine is left with **no usable pointer at all** until they are corrected by hand. The service does not restore them on shutdown, and `udevadm trigger` does not either.

## Environment

- Lenovo **Legion Go 2** (`83N0001ALM`, Z2 Extreme / 8ASP2)
- CachyOS Handheld, kernel `7.1.8-1-cachyos-deckify`
- InputPlumber **0.78.1**, stock `50-legion_go_2.yaml`, no custom config in `/etc/inputplumber`

## Steps to reproduce

1. Boot with `inputplumber.service` running. It grabs the controller's evdev nodes and logs:

   ```
   Blocking input events from /dev/input/event3
   Blocking input events from /dev/input/event4
   Blocking input events from /dev/input/event5
   ```

   At this point all three are mode `000`, which is expected while the daemon owns them.

2. `sudo systemctl stop inputplumber`

3. Inspect the same nodes.

## Observed

```
event3    Legion Controller  Keyboard    000 -> 660   released
event4    Legion Controller  Mouse       000 -> 000   still blocked
event5    Legion Controller  Touchpad    000 -> 000   still blocked
event10   Generic X-Box pad              000 -> 060   partially restored
```

The virtual target devices disappear with the daemon, and the physical mouse and touchpad are still unreachable, so the desktop session has no pointer device left. Only the built-in touchscreen (a device InputPlumber never touched) keeps working.

`udevadm trigger --action=change` on those nodes does **not** restore the mode or ownership — udev does not reapply permissions for a `change` event on an already-present device.

## Expected

On shutdown, the daemon should release its grabbed source devices and leave their mode and ownership as it found them (normally `660 root:input`), so the system falls back to using them directly.

## Workaround

```bash
sudo chown root:input /dev/input/event4 /dev/input/event5
sudo chmod 660 /dev/input/event4 /dev/input/event5
```

Or reboot.

## Notes

This may share a root cause with #562 ("Composite Device sources may not fully clean up during /dev reset"), but the trigger is different: that one is about `/dev` being reset during boot, this one is a plain `systemctl stop` of the service.

It also makes the "stop `inputplumber.service`" workaround suggested in #538 and #604 harder to use than it looks: on this device stopping the service leaves you with no pointer until the permissions are fixed manually, and once they are, the touchpad's absolute node is misclassified as a touchscreen — described in my comment on #538.

---

*Investigation assisted by Claude (Opus 5) via Claude Code.*

nileshchakraborty:
When `hide_device()` hides an input device, dynamic udev rules modify permissions on the source devnode.

On service stop or `unhide_all()`, permissions on existing device nodes were not restored, leaving them inaccessible (#686).

### Changes
- Dynamically record pre-existing device permissions before hiding and restore them on `unhide_device()` / `unhide_all()`.
- Clean up `/dev/inputplumber/by-hidden` symlinks on restore.
- In `reload_all()`, trigger `remove` and `add` actions scoped to `-s input` and `-s hidraw` to reapply default udev tags and ownership.

Fixes #686

### DIFF
--- a/src/udev/mod.rs
+++ b/src/udev/mod.rs
@@ -165,6 +165,8 @@ pub async fn unhide_device(path: String) -> Result<(), Box<dyn Error>> {
     let Some(parent) = device.get_parent() else {
         return Err("Unable to determine parent for device".into());
     };
+
+    // Remove all created udev rules
     let rule_path = format!(
         "{RULES_PREFIX}/{RULE_HIDE_DEVICE_EARLY_PRIORITY}-inputplumber-hide-{name}-early.rules"
     );
@@ -181,6 +183,7 @@ pub async fn unhide_device(path: String) -> Result<(), Box<dyn Error>> {
         let _ = fs::remove_file(&hidden_symlink);
     }
 
+    // Move the device back
     let src_path = format!("/dev/inputplumber/sources/{name}");
     let dst_path = if name.starts_with("event") || name.starts_with("js") {
         format!("/dev/input/{name}")
@@ -206,6 +209,7 @@ pub async fn unhide_device(path: String) -> Result<(), Box<dyn Error>> {
         }
     }
 
+    // Reload udev
     reload_children(parent).await?;
 
     Ok(())
@@ -213,6 +217,7 @@ pub async fn unhide_device(path: String) -> Result<(), Box<dyn Error>> {
 
 /// Unhide all devices hidden by InputPlumber
 pub async fn unhide_all() -> Result<(), Box<dyn Error>> {
+    // Remove all created udev rules
     if let Ok(entries) = fs::read_dir(RULES_PREFIX) {
         for entry in entries.flatten() {
             let filename = entry.file_name().to_string_lossy().to_string();
@@ -249,6 +254,7 @@ pub async fn unhide_all() -> Result<(), Box<dyn Error>> {
         let _ = fs::remove_dir("/dev/inputplumber/by-hidden");
     }
 
+    // Move all devices back
     if Path::new("/dev/inputplumber/sources").is_dir() {
         if let Ok(entries) = fs::read_dir("/dev/inputplumber/sources") {
             for entry in entries.flatten() {
@@ -286,6 +292,7 @@ pub async fn unhide_all() -> Result<(), Box<dyn Error>> {
 
     let _ = fs::remove_dir("/dev/inputplumber");
 
+    // Reload udev rules
     reload_all().await?;
 
     Ok(())
--- a/src/udev/mod.rs
+++ b/src/udev/mod.rs
@@ -9,6 +9,7 @@ pub mod device;
 use std::{
     error::Error,
     fs,
+    os::unix::fs::PermissionsExt,
     path::{Path, PathBuf},
 };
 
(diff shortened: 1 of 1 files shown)


src/udev/mod.rs
```diff
@@ -154,17 +155,22 @@ pub async fn unhide_device(path: String) -> Result<(), Box<dyn Error>> {
...
-            format!("/dev/{name}")
-        };
         log::debug!("Restoring device node path '{src_path}' to '{dst_path}'");
         if let Err(e) = fs::rename(&src_path, &dst_path) {
             log::warn!("Failed to move device node from {src_path} to {dst_path}: {e}");
         }
     }
+    if let Ok(metadata) = fs::metadata(&dst_path) {
+        let mut permissions = metadata.permissions();
+        permissions.set_mode(0o660);
```
pastaq: This still has a flaw in that is assumes the permissions of a device node. This could be too permissive or too restrictive depending on the original permissions. Perhaps we should store the prexisting permission level in a hash set by dst_path when adding the udev rules so we can restore the correct permissions on restore.

src/udev/mod.rs
```diff
@@ -212,6 +218,11 @@ pub async fn unhide_all() -> Result<(), Box<dyn Error>> {
         if let Err(e) = fs::rename(&path, &dst_path) {
             log::warn!("Failed to move device node from {path:?} to {dst_path}: {e}");
         }
+        if let Ok(metadata) = fs::metadata(&dst_path) {
+            let mut permissions = metadata.permissions();
+            permissions.set_mode(0o660);
```
pastaq: Here too

pastaq (changes requested):

nileshchakraborty:
### Testing & Validation

1. **Unit Tests & Clippy (Fedora 44 container)**:
   - `cargo test --all-targets` -> 10 passed, 0 failed, 4 doc-tests passed.
   - `cargo clippy --all-targets -- -D warnings` -> Clean, 0 warnings.

2. **Live Device Reproduction & Verification (Bazzite Deck 44, ROG Ally X)**:
   - Confirmed hidden nodes under 0.79.0 before service stop:
     ```text
     /dev/inputplumber/by-hidden/event3 -> /dev/input/event3 (mode 0000: c---------)
     /dev/inputplumber/by-hidden/event4 -> /dev/input/event4 (mode 0000: c---------)
     /dev/inputplumber/by-hidden/event6 -> /dev/input/event6 (mode 0000: c---------)
     /dev/inputplumber/by-hidden/hidraw5 -> /dev/hidraw5 (mode 0000: c---------)
     /dev/inputplumber/by-hidden/hidraw6 -> /dev/hidraw6 (mode 0000: c---------)
     /dev/inputplumber/by-hidden/js0 -> /dev/input/js0 (mode 0000: c---------)
     ```
   - Verified that `fs::set_permissions(&target, 0o660)` resets character node permissions to `crw-rw----` and scoped udev triggers (`udevadm trigger -s input -s hidraw`) restore `uaccess` console ACLs upon stopping.

src/udev/mod.rs
nileshchakraborty: Updated. We now record the pre-existing permissions mode in a map keyed by destination path prior to applying hide rules and restore that exact mode on unhide.

src/udev/mod.rs
nileshchakraborty: Fixed here as well in both by-hidden and sources unhide paths.

src/udev/mod.rs
```diff
@@ -145,76 +169,123 @@ pub async fn unhide_device(path: String) -> Result<(), Box<dyn Error>> {
...
         "{RULES_PREFIX}/{RULE_HIDE_DEVICE_LATE_PRIORITY}-inputplumber-hide-{name}-late.rules"
     );
     log::debug!("Removing hide rule: {rule_path}");
-    fs::remove_file(rule_path)?;
+    let _ = fs::remove_file(&rule_path);
+
+    let hidden_symlink = format!("/dev/inputplumber/by-hidden/{name}");
+    if Path::new(&hidden_symlink).exists() {
+        let _ = fs::remove_file(&hidden_symlink);
+    }
```
pastaq, line 185: Retain comments

src/udev/mod.rs
```diff
@@ -145,76 +169,123 @@ pub async fn unhide_device(path: String) -> Result<(), Box<dyn Error>> {
...
+        format!("/dev/{name}")
+    };
+    if Path::new(&src_path).exists() {
         log::debug!("Restoring device node path '{src_path}' to '{dst_path}'");
         if let Err(e) = fs::rename(&src_path, &dst_path) {
             log::warn!("Failed to move device node from {src_path} to {dst_path}: {e}");
         }
     }
 
-    // Reload udev
```
pastaq: Here too

src/udev/mod.rs
```diff
@@ -145,76 +169,123 @@ pub async fn unhide_device(path: String) -> Result<(), Box<dyn Error>> {
...
+    }
+
     reload_children(parent).await?;
 
     Ok(())
 }
 
 /// Unhide all devices hidden by InputPlumber
 pub async fn unhide_all() -> Result<(), Box<dyn Error>> {
-    // Remove all created udev rules
```
pastaq: Here too

src/udev/mod.rs
```diff
@@ -145,76 +169,123 @@ pub async fn unhide_device(path: String) -> Result<(), Box<dyn Error>> {
...
+                        }
+                    }
+                }
+                let _ = fs::remove_file(symlink_path);
+            }
+        }
+        let _ = fs::remove_dir("/dev/inputplumber/by-hidden");
     }
 
-    // Move all devices back
```
pastaq: Here too

src/udev/mod.rs
```diff
@@ -145,76 +169,123 @@ pub async fn unhide_device(path: String) -> Result<(), Box<dyn Error>> {
...
+            }
         }
+        let _ = fs::remove_dir("/dev/inputplumber/sources");
+    }
+
+    if let Ok(mut saved) = SAVED_PERMISSIONS.lock() {
+        saved.clear();
     }
 
-    // Reload udev rules
```
pastaq: Here too

pastaq (changes requested):

src/udev/mod.rs
nileshchakraborty: Restored all original section comments in both `unhide_device` and `unhide_all`.

pastaq (approved):

PR #730, opened 2026-09-20
fix: Persist target device settings

### LINKED ISSUE #699: DualSense emulated controller detection issues in Steam Game Mode
Quite often in game mode, when the gamepad is emulated to DualSense, Steam reports no controllers connected but the inputs do work to navigate the menus and the gamepad works in games. That would only be a cosmetic problem but Steam Input customization is completely turned off if Steam thinks there's no controller detected.

Switching to an always working emulation device (like the Xbox Elite Controller) and then back to DualSense has a chance to fix it and finally make it appear. Sometimes it takes just one try, sometimes 2 or even up to 5.

Running Bazzite on GPD Win 4 2023 7640u

[logs-this-boot.txt](https://github.com/user-attachments/files/31948676/logs-this-boot.txt)

https://github.com/user-attachments/assets/c32ef13c-6a45-4538-bf7c-64fb763f36d7

Image

pastaq:
Fixes various issues related to target device serial numbers being inconsistent.
- Derive a valid & repeatable MAC address for dualsense/edge targets and set its uniq so the same device is always recognized between disconnects and restarts. Also fixes double input from the virtual devices evdev and SDL hidraw capture.
- Derive a valid & repeatable serial and board number for deck targets so the same device is always recognized between disconnects and restarts.
- Derive a valid & repeatable MAC address for 8BitDo Ultimate 2 target and set its uniq so the same device is always recognized between disconnects and restarts. Also fixes double input from the virtual devices evdev and SDL hidraw capture.
- Removes rand dependency.

Closes #699 

AI Disclosure:
I used an LLM to research some of the concepts used here, and to break the code up into individual commits once the concept was flushed out. I also used it to do an initial review.


src/input/target/dualsense.rs
```diff
@@ -61,31 +68,58 @@ pub enum BusType {
...
+
+/// Returns the address to use for a new device of the given model. Prefers
+/// deriving it from `persistent_id`; falls back to the next address in that
+/// model's counter sequence when no persistent id is available.
+pub fn generate_mac(model: ModelType, persistent_id: Option<&str>) -> [u8; 6] {
+    match persistent_id.filter(|id| !id.is_empty()) {
+        Some(id) => mac_from_persistent_id(model, id),
+        None => next_mac(model),
+    }
+}
```
ShadowApex, line 87: Let's make it so target devices are guarenteed to have a persistent id (e.g. it is not an `Option<&str>`). If no source devices have a persistent id, then we can use the composite device's dbus path as a seed for a persistent id.

src/input/target/steam_deck_uhid.rs
```diff
@@ -822,15 +858,12 @@ impl TargetOutputDevice for SteamDeckUhidDevice {
...
+            let device = SteamDeckUhidDevice::create_virtual_device(&config, &serial_number)?;
             self.device = Some(device);
             self.config = config;
             self.config_rx = None;
-            self.serial_number = format!(
-                "{:04x?}-{:04x?}-1ae1c0b",
-                VID,
-                self.config.product_id.to_u32()
-            );
+            self.serial_number = serial_number;
```
ShadowApex, line 866: Should we also generate the board number here?

src/input/target/dualsense.rs
```diff
@@ -61,31 +68,58 @@ pub enum BusType {
...
+
 /// The [DualSenseHardware] defines the kind of DualSense controller to emulate
 #[derive(Debug, Copy, Clone, PartialEq)]
 pub struct DualSenseHardware {
-    model: ModelType,
-    bus_type: BusType,
-    mac_addr: [u8; 6],
+    pub model: ModelType,
+    pub bus_type: BusType,
+    pub mac_addr: [u8; 6],
```
ShadowApex, line 116: No need for these to be public.

src/drivers/steam_deck/mod_test.rs
ShadowApex, line 1: We shouldn't call this `mod_test`

src/drivers/mod.rs
```diff
@@ -16,3 +16,14 @@ pub mod ultimate_2;
...
+/// Hashes `bytes` using FNV-1a.
+pub fn hash_id(bytes: &[u8]) -> u64 {
+    /// Starting value of the hash, before any bytes are mixed in.
+    const OFFSET_BASIS: u64 = 0xcbf29ce484222325;
+    /// Multiplied into the hash after each byte to spread its bits.
+    const PRIME: u64 = 0x100000001b3;
+    bytes.iter().fold(OFFSET_BASIS, |hash, &b| {
+        (hash ^ b as u64).wrapping_mul(PRIME)
+    })
+}
```
ShadowApex, line 29: Since we mainly use this in the target devices, let's put this in the targets module.

ShadowApex (changes requested):

PR #731, opened 2026-09-20
fix(Targets): Stop target device drivers when the composite device is torn down

### LINKED ISSUE #317: [Bug] - Steam Deck target device lost while detaching from a USB dock during resume
When using a device that's configured to emulate a Steam Deck controller like the Legion Go S, it's possible to lose the target device if you follow a certain set of criteria while resuming from suspend. To reproduce this issue, you'll need a dock (I used the Steam Deck dock for this), an external display, and a Bluetooth controller like an Xbox Elite 2.

1. Connect Legion Go S to a dock with at least one display output active
2. Connect a Bluetooth Xbox controller
3. Using the Xbox controller, suspend the device from the left-side menu of Steam
4. Once the device is suspended, resume it with the power button
5. While the device is still waking up, detach the USB dock

At this point, InputPlumber will fail to properly set up the target device again leaving you with basic Xbox controls via InputPlumber without any way to open the side menus. The external controller is not a factor after resume as that will still be turned off from when you suspended the device. 

Restarting the InputPlumer service sometimes works around the issue, although there have been times where I've completely lost all input except for the touchscreen and had to reboot the device to fix it.

Log of a failure with debugging verbosity: [systemreport-20250316064436.txt](https://github.com/user-attachments/files/19307626/systemreport-20250316064436.txt). Line ~12000 is where you see a resume from suspend -> failure.

pastaq:
on_composite_device_stopped() previously just dropped each target's TargetDeviceClient handle instead of calling stop() on it, so the target task exited via a channel-disconnect error path that skipped implementation.stop()/destroy(), that leaked the uhid virtual device on every composite device teardown.

closes #317


src/input/manager.rs
```diff
@@ -922,7 +922,11 @@ impl Manager {
         // Find any target devices that were in use by the composite device
         if let Some(target_device_paths) = self.composite_device_targets.get(&path) {
             for target_device_path in target_device_paths {
-                self.target_devices.remove(target_device_path);
+                if let Some(client) = self.target_devices.remove(target_device_path) {
+                    if let Err(e) = client.stop().await {
+                        log::warn!("Failed to stop target device {target_device_path}: {e:?}");
+                    }
+                }
```
ShadowApex, line 929: We should use `let Some(client) = ... else { ... }` here

ShadowApex (changes requested):

PR #733, opened 2026-09-21
fix(config): add AYANEO Pocket S2 hidraw rumble source

mydanyi:
The AYANEO Pocket S2 profile already matches its evdev controller, but it does not declare the controller's hidraw haptics interface. As a result, the input path is available while the InputPlumber haptics source is not attached for the device.

Add the existing upstream hidraw source format used by `50-ayaneo_pocket_ds.yaml`, selecting the Pocket S2 interface with vendor `0x4001`, product `0x0428`, and interface `0`. This is a six-line configuration-only change; controller targets and input mappings are unchanged.

This follows the device profile introduced by #627. It is separate from #729: that PR covers generic/driver rumble behavior, while this change adds the missing Pocket S2 source declaration so the existing haptics path can be discovered.

Validation:

- `cargo test config::config_test::check_autostart_rules -- --exact --show-output` in the project Docker image: passed (1 test).
- `git diff --check`: passed.
- The profile was compared with the existing AYANEO Pocket DS hidraw declaration.
- The equivalent source was tested on the Pocket S2 hardware; Steam's controller identification rumble worked after the source was attached.

AI disclosure: CodeBuddy DeepSeek V4.1 Flash drafted this YAML-only addition. Codex independently reviewed the upstream profiles, checked for overlapping upstream work, and ran the configuration test.


pastaq:
Please just put `Assisted-by: LLM` in the commit message. We're not interested in advertising specific models.

